PR #3, opened 2025-02-12
[WEB-3574] data transformation for api migration

hhkarimi:


scripts/tgp_to_gp/ai_chat.sql
```diff
@@ -0,0 +1,25 @@
...
+UPDATE ai_chat SET created_at = to_timestamp(createdAt / 1000.0);
+UPDATE ai_chat SET updated_at = to_timestamp(updatedAt / 1000.0);
+
+-- drop the original columns
+ALTER TABLE ai_chat DROP COLUMN createdAt;
+ALTER TABLE ai_chat DROP COLUMN updatedAt;
+
+-- rename the new columns
+ALTER TABLE ai_chat RENAME COLUMN created_at TO createdAt;
+ALTER TABLE ai_chat RENAME COLUMN updated_at TO updatedAt;
```
oakinh, line 21: While the schema casing has these columns as camelCase, in the actual DB they're snake case, so I think we'll need to keep the names as snake case and not rename them here

hhkarimi: In the source db, I see them as camel case:
image

For the destination db, I can keep them in snake case. Does that sound right?

scripts/tgp_to_gp/ai_chat.sql
```diff
@@ -0,0 +1,25 @@
...
+-- rename columns
+ALTER TABLE ai_chat RENAME COLUMN thread TO thread_id;
+ALTER TABLE ai_chat RENAME COLUMN campaign TO campaign_id;
+
+-- case when to convert milliseconds to timestamp
+ALTER TABLE ai_chat ADD COLUMN created_at TIMESTAMP;
+ALTER TABLE ai_chat ADD COLUMN updated_at TIMESTAMP;
+
+UPDATE ai_chat SET created_at = to_timestamp(createdAt / 1000.0);
+UPDATE ai_chat SET updated_at = to_timestamp(updatedAt / 1000.0);
```
oakinh, line 13: This conversion should work well

scripts/tgp_to_gp/campaign.sql
```diff
@@ -0,0 +1,21 @@
...
+ALTER TABLE campaign RENAME COLUMN "isVerified" TO is_verified;
+ALTER TABLE campaign RENAME COLUMN "isPro" TO is_pro;
+ALTER TABLE campaign RENAME COLUMN "didWin" TO did_win;
+ALTER TABLE campaign RENAME COLUMN "dateVerified" TO date_verified;
+ALTER TABLE campaign RENAME COLUMN "dataCopy" TO data_copy;
+ALTER TABLE campaign RENAME COLUMN "pathToVictory" TO path_to_victory;
+ALTER TABLE campaign RENAME COLUMN "aiContent" TO ai_content;
+ALTER TABLE campaign RENAME COLUMN "ballotCandidate" TO ballot_candidate;
+ALTER TABLE campaign RENAME COLUMN "isDemo" TO is_demo;
+ALTER TABLE campaign RENAME COLUMN "vendorTsData" TO vendor_ts_data;
```
oakinh, line 15: You have a duplicate here with dateVerified

You can do it like this instead:

```suggestion
SELECT 
    format(
        'ALTER TABLE campaign RENAME COLUMN "%I" TO %I;', 
        column_name, 
        regexp_replace(column_name, '([a-z0-9])([A-Z])', '\1_\2', 'g')::text
    ) 
FROM information_schema.columns 
WHERE table_name = 'campaign';
```

scripts/tgp_to_gp/ai_chat.sql
oakinh: Yes for the destination DB column names need to be in snake case. Field names on most of the JSON ("data" column) can stay in camelCase, with the only exception being the JSON columns of race and related (county, municipality tables) which stay snake_case

PR #16, opened 2025-03-14
[WEB-3745] dbt: build election mart

hhkarimi:
`int_*` files are data models in the intermediate layer. They rely on an upstream table, typically from the staging/raw/silver layer, which references data after it's initially ingested to data bricks.

`m_election_api_*` are data models that live in the gold/mart layer and will be loaded to the presentation db to serve the election api.


dbt/project/models/mart/election_api/m_election_api__candidacy.sql
hhkarimi, line 1: @oakinh this is the simplest one, `candidacy`, that is nearly loaded to election db as is.

dbt/project/models/intermediate/int__enhanced_place.sql
```diff
@@ -0,0 +1,67 @@
...
+            tbl_fun_facts.unemployment_rate as unemployment_rate,
+            tbl_fun_facts.home_value as home_value
+        /* Need to add the following
+                // Relations
+                Race     Race[]
+                // Recursive relationships
+                children Place[] @relation("PlaceHierarchy")
+                parent   Place?  @relation("PlaceHierarchy", fields: [parentId], references: [id])
+                parentId String? @db.Uuid
+        */
```
oakinh, line 40: We normally mark these with a `TODO:` prefix to easily find them all. Some IDE's have built in support for this

dbt/project/models/intermediate/int__enhanced_position.sql
```diff
@@ -0,0 +1,67 @@
...
+            tbl_fun_facts.unemployment_rate as unemployment_rate,
+            tbl_fun_facts.home_value as home_value
+        /* Need to add the following
+                // Relations
+                Race     Race[]
+                // Recursive relationships
+                children Place[] @relation("PlaceHierarchy")
+                parent   Place?  @relation("PlaceHierarchy", fields: [parentId], references: [id])
+                parentId String? @db.Uuid
+        */
```
oakinh, line 40: `TODO:` prefix

dbt/project/models/intermediate/int__enhanced_race.sql
```diff
@@ -0,0 +1,132 @@
...
+        where tbl_freq.valid_to is null
+        group by position_database_id
+    ),
+    filing_period_ids as (
+        select
+            tbl_race.database_id as race_database_id,
+            /*
+            Take the filing period. Of the 1MM races, only 10k have more than 1 filing period.
+            In these cases, the last value is the latest filing period.
+            */
```
oakinh, line 33: This is super interesting, have you talked about this with Samuel or someone else? Seems like it would have huge impacts if the dates are significantly different for marketing/sales since they prioritize based on filing deadline

dbt/project/models/intermediate/int__enhanced_race.sql
```diff
@@ -0,0 +1,132 @@
...
+            tbl_position.salary,
+            tbl_position.sub_area_name,
+            tbl_position.sub_area_value,
+            tbl_election_frequency.frequency,
+            tbl_filing_period.start_on as filing_date_start,
+            tbl_filing_period.end_on as filing_date_end,
+            -- need to add field `place` (how to represent object?)
+            -- need to add int `place_id` (database_id?, possibly use geo_id to match
+            -- position to a single place)
+            -- try position first
```
oakinh, line 73: Add `TODO:`'s

dbt/project/models/intermediate/int__enhanced_race.sql
```diff
@@ -0,0 +1,132 @@
...
+            {{ generate_salted_uuid(fields=["tbl_race.id"], salt="ballotready") }}
+            as id,
+            tbl_race.id as br_hash_id,
+            tbl_race.database_id as br_database_id,
+            tbl_race.is_primary,
+            tbl_race.is_runoff,
+            tbl_race.created_at,
+            tbl_race.updated_at,
+            to_timestamp(tbl_election.election_day) as election_date,
+            tbl_position.slug as position_slug,
```
oakinh, line 50: Have you verified that BR's position_slug is the same slugified format that we expect?

dbt/project/models/mart/election_api/m_election_api__race.sql
hhkarimi, line 1: @oakinh I'm using position_database_id  here. I haven't filled in place or place_id since I position for location matching to fun facts looks good to me. You can check out the SQL in the comments in databricks to confirm.

Let me know what you think, I can change position_database_id to more cleanly be position_id or position_br_database_id

dbt/project/models/mart/election_api/m_election_api__position.sql
hhkarimi, line 1: @oakinh Since I'm suggesting position instead of place, this is the position table I'd load into the election db. Thoughts?

dbt/project/models/intermediate/int__position_fun_facts.sql
hhkarimi, line 1: @oakinh this is the logic I'm using for matching to fun facts. I'm missing about 2k out of 200k positions; most (if not all) are at the state/federal level, or things like school districts.

I also do the same for place in int__place_fun_facts.

dbt/project/models/mart/election_api/m_election_api__candidacy.sql
hhkarimi, line 1: @oakinh this is the simplest one, candidacy, that is nearly loaded to election db as is.

dbt/project/models/mart/election_api/m_election_api__position.sql
oakinh: I'm not sure what you mean by "suggesting position instead of place"
And we don't have a position table in the election db. Unless you mean the race table?

dbt/project/models/intermediate/int__position_fun_facts.sql
oakinh: 198K/200K is way more than expected, so that's great

dbt/project/models/mart/election_api/m_election_api__candidacy.sql
oakinh: Don't even worry about loading `candidacy` at this point to the election db, it could very likely change significantly. I haven't started on the candidacy endpoints yet, so no clue what data we'll need. It's just leftover from when Taylor put all of the tables as-is in the schema

dbt/project/models/intermediate/int__enhanced_place.sql
hhkarimi: For `Race[]`, do you know how this is represented in the postgres db that backs the Prisma objects? If I understand it correctly, this is going to be an array of ids that make up the primary key of the `race` table. Is that correct?

dbt/project/models/intermediate/int__enhanced_place.sql
```diff
@@ -0,0 +1,67 @@
...
+            tbl_fun_facts.income_household_median as income_household_median,
+            tbl_fun_facts.unemployment_rate as unemployment_rate,
+            tbl_fun_facts.home_value as home_value
+        /* Need to add the following
+                // Relations
+                Race     Race[]
+                // Recursive relationships
+                children Place[] @relation("PlaceHierarchy")
+                parent   Place?  @relation("PlaceHierarchy", fields: [parentId], references: [id])
+                parentId String? @db.Uuid
```
hhkarimi, line 39: If we're going to use `position` instead of place for the election/race location, do we need these parent hierarchy in `place`?

dbt/project/models/intermediate/int__enhanced_race.sql
hhkarimi: Good point. I'll tag them.

dbt/project/models/intermediate/int__enhanced_race.sql
hhkarimi: I didn't do any transforms here, it's the raw `slug` from the `Position` object: https://developers.civicengine.com/docs/api/graphql/reference/objects/position#positionslugstring--

Here's an example output:
```
select slug
from dbt_hugh.stg_airbyte_source__ballotready_api_position
limit 100
;

oklahoma-covington-douglas-school-board-member-marshall-precinct
oklahoma-tishomingo-school-board-member-seat-2
oklahoma-holly-creek-school-board-member-seat-2
oklahoma-choctaw-nicoma-park-school-board-member-seat-1
oklahoma-lukfata-school-board-member-seat-2
oklahoma-edmond-school-board-member-district-5
oklahoma-ada-school-board-member-ward-4
oklahoma-allen-school-board-member-seat-1
oklahoma-tecumseh-school-board-member-seat-2
oklahoma-central-school-board-member-seat-4
```
What do you think, does this line up with our use, or do I need to add in some checks/processing?

dbt/project/models/intermediate/int__enhanced_race.sql
hhkarimi: I'll remove `place` with `position`. In that case, I think it only needs the `position_id` in your schema.

dbt/project/models/mart/election_api/m_election_api__position.sql
hhkarimi: I mean when relating a `race` to a location, we can use the location of the [position](https://developers.civicengine.com/docs/api/graphql/reference/objects/position#positionslugstring--). Yes, this would require having a `position` table in the election db, that would look very much like the current `place` table.

In this PR, you can see the analogous `m_election_api__position.sql` with `m_election_api__place.sql`. It's the same columns.

oakinh: Maybe I'm misunderstanding, but this sounds like doing something different than we discussed. Back on the 19th I sent you these two ideas. What were your findings after investigating each of these methods?

> 1. All places that have a geoId found on any given position
>    a. This works only if the geoId on position is actually the correct one (eg, geoId for a county position leads to a county or county equivalent place) I need you to investigate and verify this at least with some spot checking if you go this route
> 2. For each race->position, choose the place with the highest level of specificity in the array

dbt/project/models/intermediate/int__enhanced_place.sql
oakinh: I thought I was having déjà vu for a second there... We talked about this on the schema PR: https://github.com/thegoodparty/election-api/pull/6#discussion_r2006431698

Does that answer your question?

dbt/project/models/intermediate/int__enhanced_race.sql
oakinh: That is not what we need, `position_slug` should be `normalized_position_name` slugified
https://github.com/thegoodparty/tgp-api/blob/b89791c331a61ee05d77f4599c38e6d582799fc9/api/helpers/ballotready/add-race.js#L111-L112
There's a comment in the prisma schema as well for this

dbt/project/models/intermediate/int__enhanced_place.sql
oakinh: We never discussed that nor agreed on changing that, and I'm not clear on why you see that change as necessary

dbt/project/models/intermediate/int__enhanced_place.sql
hhkarimi: Not quite, when I load the data from the mart to the election db, It's unclear what should actually be in `Race[]` if not a list of the primary key `id` in the `race` table. Unless you mean that I should leave `Race[]` as `null` so that Prisma fills it out automatically, but I don't think that's the case.

dbt/project/models/intermediate/int__enhanced_race.sql
hhkarimi: fixed

dbt/project/models/intermediate/int__enhanced_place.sql
hhkarimi: Perfect, I'll remove `Place.Race` from the mart. Appreciate the detailed explanation.

dbt/project/models/mart/election_api/m_election_api__position.sql
hhkarimi: I think we can resolve things in an simpler fashion with `position`. It doesn't rely on going through a nest of geo-ids. I was looking at this last week: https://github.com/thegoodparty/gp-data-platform/pull/16#discussion_r2012528461

We also avoid the ambiguity of entering a 1:many relation by sticking with a 1:1 relation. I did spot checks, the two biggest issues in the join logic ([dbt/project/models/intermediate/int__position_fun_facts.sql](https://github.com/thegoodparty/gp-data-platform/pull/16/files/92067504018dda42463a7447194b6f61f1168a3c#diff-f3c84cb77913228fc964d2eb01dbf4e7551aa70d67b2ef4ad13789642afd0488)) is that it's missing state-level matches (senator, governor, etc) and some school districts though I did manage to pick up a fair amount with fuzzy string matching.

Here's a more complete example:
```
select
  tbl_race.position_name,
  tbl_race.state,
  tbl_race.position_description,
  tbl_pos.geo_id,
  tbl_pos.city_largest,
  tbl_pos.county_name,
  tbl_pos.population,
  tbl_pos.density,
  tbl_pos.income_household_median,
  tbl_pos.unemployment_rate,
  tbl_pos.home_value,
  tbl_pos.state as tbl_position_state
from dbt_hugh.int__enhanced_race as tbl_race
left join dbt_hugh.m_election_api__position as tbl_pos
  on tbl_race.position_database_id = tbl_pos.br_database_id
limit 100
;
```

oakinh: Hugh, this is not what the API needs. I gave you the two options above because they could potentially serve what the API needs.

Using just `position` and only having one to one relations doesn't serve our purposes. We need the hierarchy of `Places` built of off `geoid`

Please stick to the schema I've outlined.

hhkarimi: Can you please explain why the hierachy is required when we can get the location/geoid directly from the 1-1 with `Position`?

hhkarimi: As discussed, I'll update the schema and naming to match the election db schema.

dbt/project/models/intermediate/int__enhanced_place.sql
hhkarimi: As discussed, I'll add in the hierarchy. Dev is in a separate PR: https://github.com/thegoodparty/gp-data-platform/pull/21/files

dbt/project/models/mart/election_api/m_election_api__place.sql
```diff
@@ -0,0 +1,32 @@
...
+    population,
+    density,
+    income_household_median,
+    unemployment_rate,
+    home_value,
+    parent_id
+from {{ ref("int__enhanced_position_w_parent") }}  -- note that the position table is used for the election place table
+{% if is_incremental() %}
+    where updated_at > (select max(updated_at) from {{ this }})
+{% endif %}
```
hhkarimi, line 32: This will be loaded to the `place` table in https://github.com/thegoodparty/election-api/blob/develop/prisma/schema/place.prisma

The output can be previewed in databricks with
```
select * from dbt_hugh.m_election_api__place limit 100;
```

image

oakinh (approved):

PR #18, opened 2025-03-18
[WEB-3880] dbt: add position_to_place mapping

hhkarimi:


dbt/project/models/intermediate/int__ballotready_position_to_place.sql
```diff
@@ -0,0 +1,33 @@
...
+    )
+}}
+
+with
+    exploded_table as (
+        select
+            database_id as position_database_id,
+            explode(places.nodes) as places,
+            created_at as position_created_at,
+            updated_at as position_updated_at
```
oakinh, line 15: You can take these in and keep them, but just be aware that our createdAt and updatedAt will be independently determined

oakinh (approved):

dbt/project/models/intermediate/int__ballotready_position_to_place.sql
hhkarimi: Thanks for the notice. Let me know how/if you'd like it to appear in the mart (the `int_*` prefix here implies it's the intermediate layer--or "silver" layer--and still one step behind the mart layer which will be loaded to the election db).

PR #21, opened 2025-03-19
WEB-3901 - Add geoIdConverter in pseudo code with an example in SQL from Claude

oakinh:
Making this a draft PR since this isn't "production ready" but just gives you the core logic to implement

### DIFF
--- a/claudeExampleGeoIdConverter.sql
+++ b/claudeExampleGeoIdConverter.sql
@@ -1,0 +1,157 @@
+-- Main stored procedure to process a batch of geoIds
+CREATE OR REPLACE PROCEDURE process_geoids(geoid_array text[])
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    current_geoid text;
+    parent_id uuid;
+    place_id uuid;
+    split_indices int[];
+    geoid_exists boolean;
+BEGIN
+    -- Process each geoId in the array
+    FOREACH current_geoid IN ARRAY geoid_array LOOP
+        -- Start a transaction for each geoId
+        BEGIN
+            -- Check if place already exists
+            SELECT EXISTS(SELECT 1 FROM place WHERE geoid = current_geoid) INTO geoid_exists;
+            
+            IF NOT geoid_exists THEN
+                -- Determine split indices based on geoId length
+                CASE LENGTH(current_geoid)
+                    WHEN 2 THEN
+                        split_indices := ARRAY[]::int[]; -- State level, no parents
+                    WHEN 4 THEN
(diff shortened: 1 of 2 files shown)


dbt/project/models/intermediate/int__geo_id_attributes.sql
```diff
@@ -0,0 +1,142 @@
...
+                when size(split_geo_ids) >= 1
+                then split_geo_ids[size(split_geo_ids) - 1]
+                else null
+            end as parent_id
+        from splitted_geo_ids
+    ),
+    geo_id_attributes as (
+        select distinct geo_id, mtfcc, slug, parent_id, left(geo_id, 2) as state
+        from with_parent_id
+    )
```
hhkarimi, line 139: The final output looks like: 
image

oakinh: I'm not clear on if there will be extra steps of transformation done with this data before loading it into the election-api DB so this could be totally irrelevant:
1) We just covered this in DM, but these slugs will need to be of the names of the locations that each part of the geoId points to, ex: `ny-queens-new-york` 
2) If parentId here is just for tracking geoIds, totally fine, but if this is for the parentId of `place` in the election-api DB, this should be the actual id of the `place`, not the parent's geoid

dbt/project/models/intermediate/int__geo_id_attributes.sql
```diff
@@ -0,0 +1,142 @@
...
+                when length(geo_id) = 7
+                then array(2)  -- State + Place/City/School District
+                when length(geo_id) = 10
+                then array(2, 5)  -- State + County + Subdivision
+                when length(geo_id) = 11
+                then array(2, 5)  -- State + County + Census tract
+                when length(geo_id) = 12 and mtfcc like 'X%'
+                then array(2, 5, 11)  -- State + County + Tract + Block Group
+                when length(geo_id) = 12 and mtfcc like 'X%'
+                then array(2, 5, 7)  -- BallotReady City Council, Justice District, School District
```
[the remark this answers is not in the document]
hhkarimi: Good catch, I fixed the typo.

I also added a comment with the doc I'm using to follow the BallotReady MTFCC logic
```sql
/*
implement case by case for geo_id length and mtcc according to
https://docs.google.com/spreadsheets/d/1t1U2oECqFRKPUVO7HYI2me-YzDgtvjbPzp5h9BqKHcU/edit?gid=0#gid=0
*/
```

oakinh (commented):
Looks good!

hhkarimi (approved):

PR #23, opened 2025-03-19
[WEB-3895] dbt: ingest filing_period

hhkarimi:


dbt/project/models/intermediate/int__ballotready_filing_period.py
```diff
@@ -0,0 +1,259 @@
...
+            filing_period_schema.withColumnRenamed("createdAt", "created_at")
+            .withColumnRenamed("databaseId", "database_id")
+            .withColumnRenamed("endOn", "end_on")
+            .withColumnRenamed("startOn", "start_on")
+            .withColumnRenamed("updatedAt", "updated_at"),
+        )
+
+    # For development/testing purposes (commented out by default)
+    # filing_periods = filing_periods.sample(False, 0.1).limit(100000)
+    # logging.info(f"filing_periods.count: {filing_periods.count()}")
```
oakinh, line 233: If this is possible with the environment you're working in, I would have things like this turned on by a flag set by your env or something that can be done from the command line. Would be useful to have these, and is more scalable as you start to have more cases like this

oakinh (changes requested):

dbt/project/models/intermediate/int__ballotready_filing_period.py
hhkarimi: Agreed and implemented.

oakinh (approved):

PR #45, opened 2025-05-05
[WEB-4158] ingest(L2): download from sftp, extract, and upload to s3

hhkarimi:


dbt/project/models/load/load__l2_sftp_to_s3.py
```diff
@@ -0,0 +1,418 @@
...
+                    f"File with base name {source_zip_file_base_name} already exists in S3"
+                )
+                return {
+                    "state_id": None,
+                    "source_file_names": None,
+                    "source_zip_file": None,
+                    "loaded_at": None,
+                    "s3_state_prefix": None,
+                }
+
```
tomer-tgp, line 165: We can extract the s3 functions to a helper file. 
This can be used in other places.

dbt/project/models/load/load__l2_sftp_to_s3.py
```diff
@@ -0,0 +1,418 @@
...
+                        # delete locally extracted files
+                        os.remove(local_file_path)
+
+                # Delete files from s3 prefix that are not in the zip file
+                valid_file_names = [
+                    f for f in file_names_to_upload if re.match(pattern, f)
+                ]
+                for s3_file_name in s3_file_list:
+                    if os.path.basename(s3_file_name) not in valid_file_names:
+                        s3_client.delete_object(Bucket=s3_bucket, Key=s3_file_name)
```
tomer-tgp, line 252: We can extract the s3 functions to a helper file. 
This can be used in other places.

tomer-tgp (changes requested):

dbt/project/models/load/load__l2_sftp_to_s3.py
hhkarimi: I agree that that's a much better design pattern, however `dbt` doesn't support code reuse from a dbt project. It's a requested feature yet to be implemented: https://docs.getdbt.com/docs/build/python-models#code-reuse 

A work around is to create a private python library, have our databricks cluster install that private library, then run our cleaner code. The cost is maintenance of that library.

IMO as our data workflows scale in number and complexity, bringing in Airflow for this type of task is best suited. This PR is for part of a custom ingestion job that Airbyte can't handle on it's own, and it's a job that  databricks/spark/pyspark isn't optimized for (I tried delegating mini batches of States to different spark workers but the long SFTP downloads were unstable on those workers).

dbt/project/models/load/load__l2_sftp_to_s3.py
hhkarimi: I agree with the principal, but it's implementation isn't directly supported. See https://github.com/thegoodparty/gp-data-platform/pull/45#discussion_r2083648951

PR #54, opened 2025-05-28
DT-54 - Enrich with geoids

oakinh:


scripts/win_number/enrich_with_geoids.py
```diff
@@ -0,0 +1,366 @@
...
+            geoids.push(geoid)
+    
+    return geoids
+
+def determine_shapefile_path(L2_district_type, state):
+    # state being a two letter abbreviation
+
+    # shapefiles should follow a convention: year_type_state
+    base_path = "./shapefiles/tl_2024_06"
+    types = L2_TO_TIGER_CODES[L2_district_type]
```
hhkarimi, line 337: Did we ever figure out the number-to-state matching? I recall Nigel was surprised the numbers here go from 1 to 78 and some numbers are skipped: https://www2.census.gov/geo/tiger/GENZ2023/shp/

scripts/win_number/enrich_with_geoids.py
```diff
@@ -0,0 +1,366 @@
...
+
+    paths = []
+
+    for type in types:
+        shapefile_path = base_path + type.lower() + state.lower()
```
hhkarimi, line 342: Maybe I'm looking at the wrong index since the naming convention on the files here don't seem to match: https://www2.census.gov/geo/tiger/GENZ2023/shp/

oakinh: I was using this: https://www.census.gov/cgi-bin/geo/shapefiles/index.php

PR #61, opened 2025-06-11
Ballotready to Hubspot model

datawithtristan:
Created model for obtaining new BallotReady records containing contact information that can be directly uploaded to HubSpot. Replaces the manual process previously handled by research and growth


dbt/project/models/intermediate/ballotready-to-hubspot/int__hubspot_ytd_candidacies.sql
```diff
@@ -0,0 +1,8 @@
+-- pulling in current year results 
+SELECT *
+FROM dbt.stg_airbyte_source__hubspot_api_contacts t1
```
hhkarimi, line 3:
```suggestion
FROM {{ ref('stg_airbyte_source__hubspot_api_contacts') }}
```

hhkarimi (commented):
Overall looks good!

You can clear the pre-commit checks by installing the hooks which will run formatting on your next commit: https://github.com/thegoodparty/gp-data-platform?tab=readme-ov-file#pre-commit-hooks

dbt/project/models/intermediate/ballotready-to-hubspot/int__ballotready_final_candidacies.sql
```diff
@@ -0,0 +1,57 @@
...
+        ELSE district
+    END AS district,
+    office_type,
+    type,
+    contact_owner,
+    owner_name,
+    candidate_id_source,
+    uncontested, 
+    number_of_candidates,
+    current_timestamp() AS created_at
```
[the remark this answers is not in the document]
hhkarimi: Makes sense.

I would keep the ballotready value for the sake of tracking, so 
```suggestion
    created_at as br_created_at,
    current_timestamp() AS created_at
```
it may also be worth picking up `updated_at`

dbt/project/models/intermediate/ballotready-to-hubspot/int__candidacies_contest_final_counts.sql
```diff
@@ -0,0 +1,93 @@
+{{ config(materialized="view", tags=["intermediate", "hubspot", "ballotready"]) }}
```
hhkarimi, line 1: Since this model has multiple stages of transformations, we can speed it up and use less compute by using an incremental materalized model.

It's fine for now, but we can revisit that if this is too slow (possible) for it's use case or eats up compute (less likely).

See an example https://github.com/thegoodparty/gp-data-platform/blob/main/dbt/project/models/intermediate/int__race_to_positions.sql#L26

dbt/project/models/intermediate/ballotready-to-hubspot/int__ballotready_final_candidacies.sql
```diff
@@ -0,0 +1,82 @@
+{{
+    config(
+        materialized="incremental",
+        incremental_strategy="merge",
```
hhkarimi, line 4: You'll need to add an `is_incremental` clause. For example: https://github.com/thegoodparty/gp-data-platform/blob/main/dbt/project/models/intermediate/int__race_to_positions.sql#L26

dbt/project/models/intermediate/ballotready-to-hubspot/int__ballotready_cleancandidacies.sql
```diff
@@ -0,0 +1,468 @@
+{{ config(auto_liquid_cluster=true, tags=["intermediate", "ballotready"]) }}
```
hhkarimi, line 1: The default materialization is a view: https://docs.getdbt.com/docs/build/materializations#configuring-materializations
So the auto clustering won't take effect. Since there are multiple stages of transformations, it's best to write it with an incremental materialization--but depending on the use case (how often it's being queried), may not be so much of an efficiency gain or need.

hhkarimi (commented):
Looks good!

I left some comments on incremental loading. It depends more on the use case then anything else, but generally more data and more transformations suggest that incremental loading is better suited for the data model.

dbt/project/models/intermediate/ballotready-to-hubspot/ballotready_to_hubspot_schema.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - name: br_candidate_code
+        description: Unique identifier for each candidate generated from BallotReady data.
+      - name: br_contest_id
+        description: Unique identifier for each contest.
+      - name: uncontested
+        description: Whether the contest is uncontested or contested.
+      - name: number_of_candidates
+        description: Total number of candidates for the contest (from both BR and HS data).
+      - name: number_of_seats_available
+        description: Number of seats available in the contest (prioritizing HS data).
```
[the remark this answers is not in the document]
datawithtristan: Tests added

dbt/project/models/intermediate/ballotready-to-hubspot/int__candidacies_contest_final_counts.sql
datawithtristan: Same comment as before on incremental loads. No unique keys to use yet.

PR #67, opened 2025-07-07
[DT-77] gp internal candidacy id

hhkarimi:


hhkarimi:
bugbot run

dbt/project/models/intermediate/int__hubspot_contacts_w_campanies.sql
```diff
@@ -0,0 +1,230 @@
...
+                tbl_companies.properties_number_of_opponents
+            ) as number_of_opponents,
+            tbl_contacts.created_at,
+            tbl_contacts.updated_at,
+            tbl_contacts.birth_date as birth_date,
+            tbl_contacts.instagram_handle as instagram_handle,
+            tbl_contacts.population as population,
+            tbl_contacts.email_contacts as email_contacts,
+            tbl_companies.properties_open_seat_ as is_open_seat,
+            tbl_companies.properties_general_election_result as candidacy_result,
```
datawithtristan, line 136: It seems the implication here is that we only keep track of results of GP candidates.

datawithtristan (approved):
Overall looks good! I think the improvements with coalescing should hopefully clean up data gaps which is a nice addition. It seems that there will be a few fields that will only be populated for GP candidates (viability score, candidacy results) which is fine for now, but I wonder if that should be discussed in a later iteration.

dbt/project/models/intermediate/int__hubspot_contacts_w_campanies.sql
```diff
@@ -0,0 +1,230 @@
...
+                tbl_contacts.office_type, tbl_companies.properties_office_type
+            ) as office_type,
+            coalesce(
+                tbl_contacts.party_affiliation, tbl_companies.properties_candidate_party
+            ) as party_affiliation,
+            coalesce(
+                tbl_contacts.is_partisan,
+                try_cast(tbl_companies.properties_partisan_np as string)
+            ) as is_partisan,
+            coalesce(tbl_contacts.state, tbl_companies.properties_state) as state,
```
datawithtristan, line 94: Have you confirmed that the data for the fields 'state' has been standardized? I think presently there are no conventions on how state is recorded in HS (New York, NY, N.Y. new york) so it might be good to handle it here, and also bucket invalid values as 'other' -> this would definitely be a good use case for a macro.

For other string fields, it might also be worth applying trim+lower transformations to minimize the permutations of different fields.

dbt/project/models/intermediate/int__hubspot_contacts_w_campanies.sql
datawithtristan: Actually, we should be able to get candidacy results going forward from the DDHQ dataset so it won't be limited to just GP candidates. Have opened a ticket for this separately

PR #68, opened 2025-07-10
[DT-151] clean up states in hubspot data

hhkarimi:


datawithtristan (approved):
All pretty straightforward, so no comments for changes

PR #70, opened 2025-07-24
[DT-178] use data from gp app when available for candidacy mart

hhkarimi:


datawithtristan (approved):
Looks good, thank you for the quick turnaround

PR #299, opened 2026-04-06
[DATA-1844] Add School Board to Win ICP flag definition

hhkarimi:
## Summary
- Adds 4 school board BallotReady normalized position types (`Local School Board`, `Local School Board Chair`, `County Board of Education`, `Local School Board of Trustees (ME only)`) to the `icp_normalized_position_names` seed
- Adds `serve_eligible` column — school board positions are **not** eligible for Serve ICP
- Adds `win_effective_date` column — school board Win/Win-Supersize ICP only active for elections on or after **2026-06-01**. Positions without a `win_effective_date` (NULL) are always eligible. School board candidacies with a NULL election date are treated as not meeting the date threshold.
- Refactors `int__icp_offices.sql`: LEFT JOIN replaces 3 correlated `NOT IN` subqueries (null-safety per CLAUDE.md), outputs `icp_win_effective_date`
- All 7 downstream models that join `int__icp_offices` gate `is_win_icp` and `is_win_supersize_icp` using `icp_win_effective_date`
- Adds seed-level tests (`unique`, `not_null`, `accepted_values`) for the seed

### Techspeed date type fix
`int__techspeed_candidates` was selecting `general_election_date` and `primary_election_date` as raw strings from the staging model. The staging model already had properly parsed date versions (`*_parsed` columns using `try_cast`/`try_to_date`), but they weren't being used. This PR swaps to the parsed versions, which:
1. Fixes the `CAST_INVALID_INPUT` error when comparing `general_election_date` (string with empty values) against `icp_win_effective_date` (date) in the techspeed mart
2. Aligns with the CLAUDE.md convention that all casting should happen in the staging layer

This also exposed a false-positive `not_null` test on `election_date` (`coalesce(general_election_date, primary_election_date)`) in `int__techspeed_candidates_clean` — 3,554 rows legitimately have neither date. The test only passed before because `coalesce('', '')` returns a non-null empty string. Removed the test.

### Office-level (`int__icp_offices`) — position eligibility before date gate

| Metric | Count |
|--------|-------|
| Total school board offices | 50,932 |
| Win ICP eligible | 44,938 |
| Win-Supersize eligible | 913 |
| Serve ICP eligible | 0 |

### Candidacy-level — after `win_effective_date` gate

| Election date bucket | Candidacies | Win ICP | Win-Supersize |
|---------------------|-------------|---------|---------------|
| `>= 2026-06-01` | 3 | 3 | 0 |
[...]

PR #306, opened 2026-04-10
[DATA-1714] Merge BallotReady + TechSpeed data via ER crosswalk

danpelota:
## Summary

Civics marts now merge 2026+ BallotReady + TechSpeed records via an entity-resolution pipeline. The architecture keeps the mart layer simple by pushing all ER-awareness into a single crosswalk consumed upstream:

- **New `int__civics_er_canonical_ids` crosswalk** — keyed on `(ts_source_candidate_id, ts_stage_election_date)`, maps each clustered TS candidacy_stage to BR's canonical `gp_*` IDs at every level (candidacy_stage, candidacy, candidate, election_stage, election).
- **TS intermediate models adopt BR's canonical IDs** via `coalesce(canonical_*, own_hash)` — with window-based propagation so every row sharing a person / election / race emits a consistent ID (matching how int_candidate / int_election / int_election_stage aggregate). Each TS int joins the crosswalk once and emits merged IDs to downstream consumers.
- **All 5 civics marts collapse to a uniform pattern**: `full outer join on shared gp_*_id` + `coalesce` for column survivorship + `array_compact(array('ballotready', 'techspeed'))` for `source_systems`. No per-mart `*_remap` CTEs, no anti-join filters for net-new TS rows — duplicates from both sides are dropped by a `qualify row_number()` dedup with BR-wins tiebreaker.
- **Field survivorship**: BR wins by default; TS wins for `is_incumbent`, `birth_date`, `phone_number`, `facebook_url`, `filing_deadline`.
- **`source_systems` array column** on every mart for lineage.
- **New hash helper macros** (`generate_ts_gp_candidate_id`, `generate_ts_gp_candidacy_id`, `generate_ts_gp_election_stage_id`) eliminate three repeated `generate_salted_uuid` field lists across the TS int models.
- **Custom assertion tests** guarantee the spine is intact: every BR election / election_stage lands in the corresponding mart, and TS net-new rows are preserved (not silently dropped by the merge).
- **Staging cleanups**: `regexp_replace`-based state normalization and `nullif(trim(...), '')` string cleaning moved into the TechSpeed staging layer; cast fix on `br_candidacy_id` in the ER source; `VIRGINA` / `LOS ANGELES` added to `clean_states`.

## Review notes

- The crosswalk strips the stage-type suffix from the ER `source_id` (`regexp_replace(source_id, '__(primary|general|runoff)$', '')`) to match TS-side `source_candidate_id`.
- Canonical ID cascades are deliberately asymmetric by level:
  - `gp_candidacy_stage_id` and `gp_election_stage_id` — stage-grain (stage-specific match only).
  - `gp_candidacy_id` — candidacy-grain (cascades to sibling stages).
  - `gp_candidate_id` and `gp_election_id` — propagated via `max() over (partition by raw hash)` so unmatched rows in a partially-matched partition still adopt BR's canonical (required for FK relationships to hold).
- `office_type` and `br_position_database_id` remain BR-only in the candidacy mart (TS candidacy intermediate does not expose these).
- `icp_win_effective_date` on `int__icp_offices` — touched with a defensive `cast(... as date)` to force this model into `state:modified+` for CI, because dbt Cloud's deferred prod artifact lagged the column that was added in #299.

## Test plan

- [x] `dbt build` green across `int__civics_er_canonical_ids` + 5 TS intermediates + 5 civics marts (150/150 tests pass locally).
- [x] FK relationship tests pass for all TS int model pairs (`candidacy_stage → candidacy / election_stage`, `candidacy → candidate / election`, `election_stage → election`, mart-level equivalents).
- [x] Custom assertions pass: `assert_all_br_elections_in_mart`, `assert_all_br_election_stages_in_mart`, `assert_ts_net_new_elections_exist`, `assert_ts_net_new_election_stages_exist`.
- [x] `source_systems` populated on every mart row with the expected combinations.
- [ ] Confirm no row-count regression vs production once CI finishes a full run.


dbt/project/macros/variable_standardization/ddhq_standardizations.sql
```diff
@@ -143,7 +143,6 @@
             or lower({{ race_name_col }}) like '%conservation%'
             or lower({{ race_name_col }}) like '%utility district%'
         then 'Other'
```
danpelota: This last block seems unnecessary if we're going to resolve to `'Other'` at the end anyway. If so, remove it an let the final `else` catch this.

dbt/project/models/intermediate/civics/int__civics_cluster_members.sql
```diff
@@ -0,0 +1,127 @@
...
+            'ballotready' as source_name,
+            br_cs.gp_candidacy_stage_id,
+            br_cs.gp_candidacy_id,
+            br_c.gp_candidate_id,
+            br_cs.gp_election_stage_id,
+            br_es.gp_election_id
+        from {{ ref("stg_er_source__clustered_candidacy_stages") }} as cw
+        inner join
+            {{ ref("int__civics_candidacy_stage_ballotready") }} as br_cs
+            on cast(cw.br_candidacy_id as string) = br_cs.br_candidacy_id
```
danpelota: We shouldn't be casting inside join conditions in intermediate models. Instead, cast `br_candidacy_id` as an integer everywhere in upstream staging models (assuming they are in fact all integers).

dbt/project/models/intermediate/civics/int__civics_cluster_members.sql
```diff
@@ -0,0 +1,127 @@
...
+            'techspeed' as source_name,
+            ts_cs.gp_candidacy_stage_id,
+            ts_cs.gp_candidacy_id,
+            ts_c.gp_candidate_id,
+            ts_cs.gp_election_stage_id,
+            ts_es.gp_election_id
+        from {{ ref("stg_er_source__clustered_candidacy_stages") }} as cw
+        inner join
+            {{ ref("int__civics_candidacy_stage_techspeed") }} as ts_cs
+            on regexp_replace(cw.source_id, '__(primary|general|runoff)$', '')
```
danpelota: I don't love the idea that we should have to manipulate a `source_id` from the clustered candidacy_stages ER match model in order to join back onto the source. Also, a candidate_code is generated at the person+position level, so we get the same candidate code if one person runs for the same office over time. 

Instead, we should update the techspeed records in the `stg_er_source__clustered_candidacy_stages` to use a direct, source-native unique identifier from the techspeed file. Since techspeed doesn't provide a unique identifier, we may need to construct one from the _airbyte_raw_id in `_airbyte_raw_id` and use that as the techspeed candidacy_stage primary key.

This should be the `source_id` we use in the ER prematch code.

dbt/project/models/marts/civics/candidacy.sql
```diff
@@ -26,47 +52,75 @@ with
...
+    candidacy_pairs as (
+        select distinct remap_candidacy_id as br_id, gp_candidacy_id as ts_id
+        from {{ ref("int__civics_cluster_members") }}
+        where source_name = 'techspeed' and remap_candidacy_id is not null
+    ),
+
+    candidate_remap as (
+        select distinct gp_candidate_id as ts_id, remap_candidate_id as br_id
+        from {{ ref("int__civics_cluster_members") }}
+        where source_name = 'techspeed' and remap_candidate_id is not null
```
danpelota: I'm unclear why we're pulling a column from `int__civics_cluster_members` model and calling it `br_id`, but filtering on `source_name` = 'techspeed'. Why explicitly filter on `techspeed` in these 3 CTEs? (candidacy_pairs, candidate_remap, election_remap).

dbt/project/models/marts/civics/candidacy_stage.sql
```diff
@@ -1,50 +1,120 @@
...
+            on ts.gp_candidacy_stage_id = cm.gp_candidacy_stage_id
+            and cm.source_name = 'techspeed'
+    ),
+
+    -- FK remaps applied to ALL TS records (not just those in cluster_members).
+    -- Derived from pre-computed remaps — no self-join needed.
+    candidacy_remap as (
+        select distinct gp_candidacy_id as ts_id, remap_candidacy_id as br_id
+        from {{ ref("int__civics_cluster_members") }}
+        where source_name = 'techspeed' and remap_candidacy_id is not null
```
danpelota: Again, I'm unclear why we're filtering `source_name = 'techspeed'` for this model.

danpelota:
@bugbot review

dbt/project/macros/parse_party_affiliation.sql
```diff
@@ -1,5 +1,7 @@
 {% macro parse_party_affiliation(column) %}
     case
+        when nullif(trim({{ column }}), '') is null
+        then null
```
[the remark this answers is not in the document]
danpelota: This macro is identical on `main` and this branch — the `then null` for empty party values was not introduced by this PR. The `generate_salted_uuid` macro already coalesces NULLs to `''`, and this behavior existed before any changes here. No action needed in this PR.

dbt/project/models/intermediate/civics/int__civics_cluster_members.sql
danpelota: Agreed — the `regexp_replace(cw.source_id, '__(primary|general|runoff)$', '')` join is fragile and couples this model to the ER prematch's ID format. Using `_airbyte_raw_id` as techspeed's candidacy_stage primary key (and the corresponding `source_id` in the ER prematch) is the right long-term fix. This will require changes to the ER prematch pipeline upstream of this model — I'll handle that in a follow-up.

dbt/project/models/marts/civics/candidacy.sql
danpelota: The remap columns on `int__civics_cluster_members` (`remap_candidacy_id`, `remap_candidate_id`, etc.) are pre-computed cross-source FK maps. They are only populated for techspeed members because BallotReady is the priority source and doesn't need remapping. So filtering `source_name = 'techspeed'` gives us the TS→BR mapping pairs where:

- `gp_candidacy_id` = the TechSpeed ID (aliased as `ts_id`)
- `remap_candidacy_id` = the corresponding BallotReady ID (aliased as `br_id`)

These pairs are then used in the `merged_2026` full outer join to merge matching TS records onto their BR counterparts, and in the FK remap coalesces so unmatched TS records still point to the correct BR foreign keys.

Happy to rename the aliases if the naming is confusing (e.g. `ts_candidacy_id` / `br_candidacy_id` instead of `ts_id` / `br_id`).

dbt/project/models/marts/civics/candidacy_stage.sql
danpelota: Same pattern as the candidacy mart — see my reply above. The remap columns are only populated for techspeed members, so the filter extracts TS→BR FK pairs for the merge coalesces.

dbt/project/models/staging/airbyte_source/techspeed_grive/stg_airbyte_source__techspeed_gdrive_candidates.sql
```diff
@@ -129,7 +130,9 @@ with
 
         from source as src
         left join
-            clean_states as cs on upper(trim(src.state)) = upper(trim(cs.state_raw))
+            clean_states as cs
```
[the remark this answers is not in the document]
danpelota: Fixed — applied the same `regexp_replace` to `cs.state_raw` in the join predicate so both sides strip special characters before comparing.

dbt/project/models/marts/civics/election_stage.sql
```diff
@@ -25,36 +28,87 @@ with
...
+    -- =========================================================================
+    combined as (
+        select *
+        from archive_2025
+        union all
+        select *
+        from merged_2026
     ),
 
     deduplicated as (
```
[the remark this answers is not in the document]
danpelota: Fixed — changed to `order by updated_at desc nulls last` to match the pattern in candidacy_stage.sql.

dbt/project/macros/variable_standardization/ddhq_standardizations.sql
danpelota: Good catch — removed the redundant `when` block and let the final `else` handle it.

dbt/project/models/intermediate/civics/int__civics_cluster_members.sql
danpelota: Fixed — changed `stg_er_source__clustered_candidacy_stages` to cast `br_candidacy_id` as string (matching the type in the intermediate layer), and removed the cast from the join condition in `int__civics_cluster_members`.

dbt/project/models/marts/civics/candidacy.sql
```diff
@@ -105,27 +159,10 @@ select
...
-        when
-            icp.icp_win_effective_date is not null
-            and (
-                deduplicated.general_election_date is null
-                or deduplicated.general_election_date < icp.icp_win_effective_date
-            )
-        then false
-        else icp.icp_win_supersize
-    end as is_win_supersize_icp,
+    icp.icp_win_supersize as is_win_supersize_icp,
```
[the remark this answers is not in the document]
danpelota: Fixed — restored the `icp_win_effective_date` temporal guard on `is_win_icp` and `is_win_supersize_icp` in all three civics mart models (candidacy, election, election_stage).

dbt/project/macros/variable_standardization/ddhq_standardizations.sql
```diff
@@ -138,12 +138,6 @@
             or lower({{ race_name_col }}) like '%sewer board%'
             or lower({{ race_name_col }}) like '%water supply%'
         then 'Water Supply Board'
-        when
-            lower({{ race_name_col }}) like '%drainage%'
-            or lower({{ race_name_col }}) like '%conservation%'
-            or lower({{ race_name_col }}) like '%utility district%'
-        then 'Other'
-
```
danpelota, line 146: I just realized this was superfluous

dbt/project/models/marts/civics/election_stage.sql
```diff
@@ -25,11 +32,15 @@ with
             cast(null as string) as filing_address,
             cast(null as string) as filing_phone,
             created_at,
-            cast(null as timestamp) as updated_at
+            cast(null as timestamp) as updated_at,
+            array('hubspot') as source_systems
```
hhkarimi, line 52: Great idea! There used to be confusion since HubSpot was an aggregator of TechSpeed and BallotReady; since it's only archives now this is a clean distinction going forward.

dbt/project/models/marts/civics/election_stage.sql
```diff
@@ -25,11 +32,15 @@ with
...
         from {{ ref("int__civics_election_stage_2025") }}
+    ),
 
-        union all
-
+    -- =========================================================================
+    -- BallotReady: full universe of election_stages (authoritative spine)
+    -- =========================================================================
+    br_2026 as (
```
hhkarimi: some indication that this is 2026+ inclusive?
```suggestion
    br_since_2026 as (
```

dbt/project/models/marts/civics/election_stage.sql
```diff
@@ -53,16 +64,72 @@ with
...
+            array('ballotready') as source_systems
         from {{ ref("int__civics_election_stage_ballotready") }}
     ),
 
+    -- =========================================================================
+    -- TechSpeed: full universe. TS int model remaps clustered stages to BR's
+    -- gp_election_stage_id AND gp_election_id (via int__civics_er_canonical_ids),
+    -- so matched rows get deduped away by the qualify below.
+    -- =========================================================================
+    ts_2026 as (
```
hhkarimi: some indication that this is 2026+ inclusive?
```suggestion
    ts_since_2026 as (
```

dbt/project/models/marts/civics/election_stage.sql
```diff
@@ -53,16 +64,72 @@ with
...
+            filing_address,
+            filing_phone,
+            created_at,
+            updated_at,
+            array('techspeed') as source_systems
+        from {{ ref("int__civics_election_stage_techspeed") }}
+    ),
+
+    -- =========================================================================
+    -- Combine all sources
```
hhkarimi: What are your thoughts on column-level coalescing like you apply in other mart tables?

dbt/project/models/marts/civics/election.sql
hhkarimi, line 1: Same comments on `election_stage` apply to `election`.

dbt/project/models/intermediate/civics/int__civics_er_canonical_ids.sql
```diff
@@ -0,0 +1,58 @@
...
+-- Dedupe: keep the most recent BR match per TS stage key
+select
+    ts_source_candidate_id,
+    ts_stage_election_date,
+    canonical_gp_candidacy_stage_id,
+    canonical_gp_election_stage_id,
+    canonical_gp_candidacy_id,
+    canonical_gp_candidate_id,
+    canonical_gp_election_id
+from stage_matches
```
hhkarimi, line 52: 😍

dbt/project/models/intermediate/civics/int__civics_candidate_techspeed.sql
```diff
@@ -7,20 +7,44 @@
...
+
+    candidates_with_id as (
+        -- If ANY candidacy of a person was clustered to BR, all raw rows for
+        -- that person adopt BR's canonical_gp_candidate_id (propagated via a
+        -- window partitioned by the raw TS hash). Otherwise keep the TS hash.
+        select
+            coalesce(
+                max(xw.canonical_gp_candidate_id) over (
+                    partition by {{ generate_ts_gp_candidate_id() }}
+                ),
```
hhkarimi, line 45: Any idea how often candidacy_stage matches between TS and BR end up pointing to mixed Persons? I like this approach, because any changes in matcha are automatically accounted for downstream.

dbt/project/models/marts/civics/election.sql
```diff
@@ -1,19 +1,143 @@
...
+            is_appointed,
+            br_normalized_position_type,
+            created_at,
+            updated_at,
+            array('techspeed') as source_systems
+        from {{ ref("int__civics_election_techspeed") }}
+    ),
+
+    -- =========================================================================
+    -- Combine all sources
```
sanjayr1: I checked the PR schema and the current `election` merge is still effectively BR-only for overlapping IDs.

For 3,693 shared `gp_election_id`s where BR has `filing_deadline` null and TS has it populated, the final mart keeps `filing_deadline` null in all 3,693 rows. The same pattern drops TS-only `population` on 2,524 rows and `is_open_seat` on 3,694 rows, and overlapping rows never retain `techspeed` in `source_systems`.

This looks like `election` may need the same full-outer-join + column-level coalescing pattern the other merged civics marts use, rather than union + BR-wins dedupe?

dbt/project/models/marts/civics/election_stage.sql
```diff
@@ -53,16 +64,72 @@ with
...
+    -- =========================================================================
+    combined as (
```
sanjayr1, line 88: Same as election, validated this in the PR schema and it is already affecting lineage, not just survivorship style.

There are 3,208 `gp_election_stage_id`s present in both `int__civics_election_stage_ballotready` and `int__civics_election_stage_techspeed`, but all 3,208 final mart rows have `source_systems = ['ballotready']`.

So the current union + BR-wins dedupe is under-reporting TechSpeed participation on every matched stage. If we want `source_systems` to reflect both contributors, does this need column-level coalescing instead of choosing the BR row wholesale?

danpelota: Ahh, great catch @sanjayr1 ! Fixing!

dbt/project/models/marts/civics/election.sql
danpelota: Fixed in 7c10c39 — refactored `election.sql` to the same full-outer-join + column-level coalesce pattern used by `candidacy.sql`/`candidacy_stage.sql`/`candidate.sql`. TS wins for `filing_deadline`, `population`, `is_open_seat`, `is_uncontested` (all four are BR-null, TS-populated).

Post-refactor: 3,697 `gp_election_id`s now carry `['ballotready','techspeed']` in `source_systems`. On those overlapping rows, `filing_deadline` / `population` / `is_open_seat` / `is_uncontested` populated on 3,694 / 2,525 / 3,695 / 3,697 rows respectively. Row-count math holds (BR + TS + archive − 3,697 overlap = mart). 43/43 tests green.

dbt/project/models/marts/civics/election_stage.sql
danpelota: Fixed in 7c10c39 — same refactor applied to `election_stage.sql`. No TS-wins columns here (TS doesn't supply data superior to BR on any stage field), but the full outer join now correctly attributes both sources in `source_systems`.

Post-refactor: 3,209 `gp_election_stage_id`s carry `['ballotready','techspeed']` (previously 0). 43/43 tests green.

dbt/project/models/marts/civics/election_stage.sql
danpelota: Applied in 7c10c39 — renamed to `br_since_2026` in both `election.sql` and `election_stage.sql`.

dbt/project/models/marts/civics/election_stage.sql
danpelota: Applied in 7c10c39 — renamed to `ts_since_2026` in both `election.sql` and `election_stage.sql`.

dbt/project/models/marts/civics/election_stage.sql
danpelota: Agreed — refactored in 7c10c39. `election_stage` now uses the same full-outer-join + column-level coalesce pattern as `candidacy.sql`. No TS-wins columns needed (BR is authoritative on every stage field), but the merge correctly reports both contributors in `source_systems` now: 3,209 overlapping stages previously flagged `['ballotready']` only.

dbt/project/models/marts/civics/election.sql
danpelota: Applied the same refactor + CTE renames to `election.sql` in 7c10c39. TS-wins columns here: `filing_deadline` / `population` / `is_open_seat` / `is_uncontested` (all four were BR-null, TS-populated, and were being silently dropped by the prior union+dedup).

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_techspeed.sql
```diff
@@ -54,71 +54,62 @@ with
...
+            canonical_gp_election_stage_id
+        from {{ ref("int__civics_er_canonical_ids") }}
+    ),
+
+    canonical_candidacy as (
+        select ts_source_candidate_id, canonical_gp_candidacy_id
+        from {{ ref("int__civics_er_canonical_ids") }}
+        qualify
+            row_number() over (
+                partition by ts_source_candidate_id order by canonical_gp_candidacy_id
```
hhkarimi, line 77: Thoughts on breaking the tie-breaker? Maybe with desc updated_at or airbyte_extracted_at?

dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
```diff
@@ -26,39 +26,50 @@ with
...
+    canonical_candidacy as (
+        select
+            ts_source_candidate_id,
+            canonical_gp_candidacy_id,
+            canonical_gp_candidate_id,
+            canonical_gp_election_id
+        from {{ ref("int__civics_er_canonical_ids") }}
+        qualify
+            row_number() over (
+                partition by ts_source_candidate_id order by canonical_gp_candidacy_id
```
hhkarimi, line 41: Thoughts on breaking the tie-breaker? Maybe with desc updated_at or airbyte_extracted_at?

dbt/project/models/intermediate/civics/int__civics_candidate_techspeed.sql
```diff
@@ -7,20 +7,44 @@
...
 with
-    candidates_with_id as (
+    -- ER crosswalk: for clustered TS candidacies, adopt BR's canonical candidate_id.
+    -- Deduped per source_candidate_id (one person may have multiple candidacies).
+    canonical_candidate as (
+        select ts_source_candidate_id, canonical_gp_candidate_id
+        from {{ ref("int__civics_er_canonical_ids") }}
+        qualify
+            row_number() over (
+                partition by ts_source_candidate_id order by canonical_gp_candidate_id
```
hhkarimi, line 17: Thoughts on breaking the tie-breaker? Maybe with desc updated_at or airbyte_extracted_at?

dbt/project/seeds/clean_states.csv
```diff
@@ -134,3 +134,5 @@ ILL,IL
 DELEWARE,DE
 NEW JERCEY,NJ
 NEW HAMPSHIRE03104,NH
+VIRGINA,VA
+LOS ANGELES,CA
```
hhkarimi, line 138: "Los Angeles" -> "CA"

Gotta love dirty data

hhkarimi (dismissed):

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_techspeed.sql
danpelota: I'm not sure it's needed – practically, this one (and the tie-breaker for other canonical IDs) only results in a tie if a single `ts_source_candidate_id` clusters to multiple canonical BR IDs. Which is _possible_, but means we have a bug in the ER flow. 

I double checked, and across all 6,025 source candidate IDs in `int_civics_er_canonical_ids`, every one maps to exactly 1 canonical candidate. So technically, this tie never fires today, but we can switch to a timestamp-based break if/when the ER every produce multi-mappings.

hhkarimi (approved):

PR #309, opened 2026-04-13
[DATA-1629] Reconcile is_win_user with organizations model

hhkarimi:
## Summary

Switches `is_win_user` from `has_campaign` (legacy `campaign.user_id` join) to `win_organization_count > 0` (org-centric, post-organization-model). Adds an `expression_is_true` test pinning the two definitions together so any drift fails CI.

Zero metric impact on current data — the two definitions return identical populations.

## Stats

| | users |
|---|---|
| Total | 68,099 |
| `has_campaign` | 61,315 |
| `win_organization_count > 0` | 61,315 |
| Mismatched | **0** |

### SQL

```sql
with
  campaigns as (select * from goodparty_data_catalog.dbt.stg_airbyte_source__gp_api_db_campaign),
  users as (select * from goodparty_data_catalog.dbt.stg_airbyte_source__gp_api_db_user),
  organizations as (select * from goodparty_data_catalog.dbt.stg_airbyte_source__gp_api_db_organization),
  user_stats as (
    select u.id as user_id,
      count(c.id) as campaign_count,
      count(case when o.owner_id = u.id then 1 end) as win_org_campaign_count
    from users u
    left join campaigns c on c.user_id = u.id
    left join organizations o on o.slug = c.organization_slug
    group by u.id
  )
select
  count(*) as total_users,
  sum(case when campaign_count > 0 then 1 else 0 end) as has_campaign,
  sum(case when win_org_campaign_count > 0 then 1 else 0 end) as has_win_org,
  sum(case when campaign_count > 0 and win_org_campaign_count = 0 then 1 else 0 end) as mismatched
from user_stats;
```

## Test plan

- [x] `dbt build --select "users+ users_win_base+ users_win_activity+"` — 132 PASS, 1 pre-existing unrelated WARN, 0 ERROR
- [x] New invariant test `dbt_utils_expression_is_true_users_has_campaign_win_organization_count_0_` passes
- [ ] Dashboard change for DATA-1629 (add `WHERE u.is_win_user` to Activated Rate widget) applied separately in Databricks

---

> [!NOTE]
> **Medium Risk**
> Changes the canonical Win-user flag feeding Win analytics marts, which could impact downstream OKR metrics if organization↔campaign linkage drifts. Adds an invariant dbt test to catch mismatches, but any latent data issues will now fail CI.
> 
> **Overview**
> Updates Win-product user identification to be *organization-centric*: `mart_civics.users.is_win_user` is now computed as `win_organization_count > 0` (instead of inferring via campaigns), and `mart_analytics.users_win_base`/`users_win_activity` switch to consuming `u.is_win_user`.
> 
> Adds documentation updates and a `dbt_utils.expression_is_true` invariant on `mart_civics.users` asserting `has_campaign = (win_organization_count > 0)` so definition drift is surfaced during builds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 027a0b157491177d73f0246e6de912c9b833e815. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/marts/civics/m_civics.yaml
+++ b/dbt/project/models/marts/civics/m_civics.yaml
@@ -91,6 +91,16 @@ models:
         data_tests:
           - not_null
 
+      - name: is_win_user
+        description: >
+          Whether the user is a Win user. True if the user owns at least one
+          organization linked to a campaign (organization_type = 'win'). This
+          is the canonical Win-user flag and the preferred source for Win
+          product metrics. Invariant with has_campaign on healthy data
+          (see the expression_is_true test on this model).
+        data_tests:
+          - not_null
+
       - name: is_serve_user
         description: >
           Whether the user is a Serve user. True if the user has a Serve
@@ -119,6 +129,11 @@ models:
           - not_null:
               config:
                 where: "is_serve_user = true"
+
+    data_tests:
+      - dbt_utils.expression_is_true:
+          arguments:
+            expression: "has_campaign = (win_organization_count > 0)"
 
   - name: campaigns
     description: >
--- a/dbt/project/models/marts/civics/m_civics.yaml
+++ b/dbt/project/models/marts/civics/m_civics.yaml
@@ -94,10 +94,7 @@ models:
       - name: is_win_user
         description: >
           Whether the user is a Win user. True if the user owns at least one
-          organization linked to a campaign (organization_type = 'win'). This
-          is the canonical Win-user flag and the preferred source for Win
-          product metrics. Invariant with has_campaign on healthy data
-          (see the expression_is_true test on this model).
+          organization linked to a campaign (organization_type = 'win').
         data_tests:
           - not_null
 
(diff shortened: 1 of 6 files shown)


hhkarimi:
@claude review

hhkarimi:
@cursor review

dbt/project/models/marts/civics/m_civics.yaml
```diff
@@ -120,6 +127,11 @@ models:
               config:
                 where: "is_serve_user = true"
 
+    data_tests:
+      - dbt_utils.expression_is_true:
+          arguments:
+            expression: "has_campaign = is_win_user"
```
danpelota: Were you thinking we keep `has_campaign` temporarily until we update any dashboards that use it? It might be a little confusing to leave two columns that mean the same thing long-term.

danpelota (dismissed):
Looks good! :shipit:

dbt/project/models/marts/civics/m_civics.yaml
hhkarimi: Yeah good, point, I kept it for backwords compatibility but actually it's not used anywhere I can't update now. And going forward, a campaign link won't be necessary. Updating.

danpelota (approved):
LGTM!

PR #310, opened 2026-04-14
[DATA-1865] Add OfficePicker zip-to-position lookup model

hhkarimi:
## Summary

- Adds `m_election_api__zip_to_position` mart model that maps zip codes to future election positions for the OfficePicker product
- Joins the civics `election` mart with `int__zip_code_to_br_office` and `m_election_api__position`, filtered to elections within a rolling 2-year window
- Uses `position_id` (election API UUID) as the primary identifier, with column names aligned to `m_election_api__position`
- Adds a 0.5% voter percentage threshold to `int__zip_code_to_l2_district` to filter out noisy zip-to-district mappings from tiny geographic overlaps
- Keyword validation join deferred to a future iteration (will use `model_predictions` schema)

## Deployment notes

- `int__zip_code_to_l2_district` needs a `--full-refresh` after merge to apply the voter threshold to historical data. Incremental runs will only apply the threshold to new records.

## Test plan

- [x] `dbt build --select m_election_api__zip_to_position` passes (1 table + 8 tests)
- [x] Inspected data: ~1.1M rows, 34.6K zip codes
- [x] `position_id` is 100% populated with valid FK to `m_election_api__position`
- [x] Grain verified: unique on `(zip_code, position_id, election_date)`
- [ ] Verify Serve team can query the table for OfficePicker integration

### DIFF
--- a/dbt/project/models/intermediate/l2/int__zip_code_to_l2_district.py
+++ b/dbt/project/models/intermediate/l2/int__zip_code_to_l2_district.py
@@ -121,24 +121,19 @@ def model(dbt, session: SparkSession) -> DataFrame:
             ).otherwise(col("zip_code")),
         )
 
-        # On full refresh, count voters per (zip, district_name) and filter out
-        # districts where fewer than MIN_VOTER_PCT of the zip's voters are
-        # assigned there. On incremental runs, skip the threshold since the
-        # percentage is unreliable on partial batches.
-        if not dbt.is_incremental:
-            district_df = district_df.groupBy(
-                "state_postal_code", "zip_code", "district_name"
-            ).agg(count("*").alias("voter_count"))
+        # Count voters per (zip, district_name) and filter out districts where
+        # fewer than MIN_VOTER_PCT of the zip's voters are assigned there.
+        district_df = district_df.groupBy(
+            "state_postal_code", "zip_code", "district_name"
+        ).agg(count("*").alias("voter_count"))
 
-            district_df = (
-                district_df.withColumn(
-                    "total_voters", spark_sum("voter_count").over(zip_window)
-                )
-                .filter(col("voter_count") / col("total_voters") >= MIN_VOTER_PCT)
-                .drop("voter_count", "total_voters")
+        district_df = (
+            district_df.withColumn(
+                "total_voters", spark_sum("voter_count").over(zip_window)
             )
-        else:
-            district_df = district_df.distinct()
+            .filter(col("voter_count") / col("total_voters") >= MIN_VOTER_PCT)
+            .drop("voter_count", "total_voters")
+        )
 
         district_df = district_df.withColumn("district_type", lit(district_type))
 
--- a/dbt/project/models/intermediate/l2/int__zip_code_to_l2_district.py
+++ b/dbt/project/models/intermediate/l2/int__zip_code_to_l2_district.py
@@ -121,19 +121,24 @@ def model(dbt, session: SparkSession) -> DataFrame:
             ).otherwise(col("zip_code")),
         )
 
-        # Count voters per (zip, district_name) and filter out districts where
-        # fewer than MIN_VOTER_PCT of the zip's voters are assigned there.
-        district_df = district_df.groupBy(
-            "state_postal_code", "zip_code", "district_name"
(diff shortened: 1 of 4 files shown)


dbt/project/models/intermediate/l2/int__zip_code_to_l2_district.py
```diff
@@ -111,6 +121,22 @@ def model(dbt, session: SparkSession) -> DataFrame:
             ).otherwise(col("zip_code")),
         )
 
+        # Count voters per (zip, district_name) and filter out districts where
```
sanjayr1, line 135: Because this model only reads `loaded_at > max_loaded_at` on incremental runs, this `voter_count / total_voters` threshold is being computed on the latest batch for the zip, not the full historical zip population. That means the 0.5% cutoff can differ from full-refresh behavior, and the later union with existing `district_names` can make false positives stick around permanently. Do we want to recompute affected zips against the full source population before applying this threshold, or otherwise avoid applying this cutoff in the incremental slice?

dbt/project/models/marts/election_api/m_election_api__zip_to_position.sql
```diff
@@ -0,0 +1,59 @@
...
+
+    positions as (
+        select id as position_id, br_database_id
+        from {{ ref("m_election_api__position") }}
+    ),
+
+    officepicker as (
+        select
+            pos.position_id,
+            elec.official_office_name as name,
```
sanjayr1: This `name` column is documented as aligned to `m_election_api__position.name`, but here we’re sourcing it from `election.official_office_name` instead. In the built PR table there are already real mismatches (for example `Fruitport Village Council` vs `Fruitport Village Board`, and `Covington City Council` vs `Covington City Board of Commissioners`). If the goal is schema consistency with `m_election_api__position`, should this come from `pos.name` instead? If `official_office_name` is the intended display label, then the YAML/PR description should be updated so consumers don’t assume it matches the position mart.

dbt/project/models/marts/election_api/m_election_api__zip_to_position.sql
```diff
@@ -0,0 +1,59 @@
...
+            br_position_database_id
+        from {{ ref("election") }}
+        where
+            election_date > current_date()
+            and election_date <= current_date() + interval 2 years
+    ),
+
+    zip_to_position as (
+        select zip_code, br_database_id from {{ ref("int__zip_code_to_br_office") }}
+    ),
```
sanjayr1, line 28: This looks like it can surface zip mappings that election_api still considers unmatched. Here we reduce `int__zip_code_to_br_office` down to only `zip_code` + `br_database_id`, so we lose the confidence / match-state information, and the later join to `m_election_api__position` only checks `br_database_id` rather than requiring a matched district. In the PR build I found 18,045 rows across 1,231 positions where every backing mapping is below 90 confidence, and all affected positions have `district_id IS NULL` in `m_election_api__position`. Should this table reuse the same acceptance gate as `m_election_api__position` instead of exposing those rows?

dbt/project/models/intermediate/l2/int__zip_code_to_l2_district.py
hhkarimi: On incremental runs, the threshold was being computed against only the new batch, so the percentages would be unreliable — and the `array_union` merge meant old false positives would stick around permanently.

Updated the incremental logic to recompute affected zips against the full source population: we identify which zips have new voter records, then read all voters for those zips from `int__l2_nationwide_uniform` to compute accurate percentages. Also removed the `array_union` merge since the full recomputation is authoritative — the merge upsert replaces old rows directly.

dbt/project/models/marts/election_api/m_election_api__zip_to_position.sql
hhkarimi: Positions without a matched district are now excluded.

dbt/project/models/marts/election_api/m_election_api__zip_to_position.sql
hhkarimi: Good point! Yes, better to use that for consistency.

sanjayr1 (approved):
LGTM

PR #332, opened 2026-04-21
[DATA-1872] Flow HubSpot-only 2025 candidacies through civics mart

hhkarimi:
## Summary

Grain shift on the 2025 archive int models so HubSpot-only candidacies (no DDHQ match) now produce `candidacy_stage` and `election_stage` rows. This unlocks ICP flagging for pre-2026 races via the existing product-DB `br_position_database_id` path that was already wired into `int__hubspot_companies_w_contacts_2025` but never reached the mart.

## Motivation

Ticket: [DATA-1872](https://goodparty.clickup.com/t/90132012119/DATA-1872).

Pre-fix, `int__civics_candidacy_stage_2025.sql` seeded from DDHQ matches and required a non-null `election_stage_date` sourced from DDHQ. HubSpot-only candidacies (the majority of 2025 records) were silently filtered out. `int__civics_election_stage_2025.sql` had the same DDHQ-gated grain, and `marts/civics/election_stage.sql` hardcoded `br_position_id = NULL` for the 2025 branch — so the ICP join could never resolve for pre-2026 stages.

Also fixes a narrower orphan gap where `int__civics_election_2025` dropped elections whose `tbl_contest.election_date` was null/out-of-range even though a matched candidacy carried a valid 2025 `general_election_date` — coalesce contest and candidacy dates to keep those elections represented.

Also plugs an upstream ID-collision: `generate_gp_election_id` hashed all-null inputs when `int__civics_candidacy_2025`'s LEFT JOIN to `int__hubspot_contest_2025` missed, bucketing ~66k unrelated candidacies onto a single shared "fake" UUID. Guard the call so unmatched contest rows yield a null `gp_election_id` instead.

## Impact

Measured against the civics mart before/after:

| metric | before | after |
|---|---|---|
| candidacy-stages with 2025 election_date | 1,867 | 122,875 |
| HubSpot-sourced results in mart | 7 | 13,940 |
| ICP-win-flaggable 2025 stages | ~0 | 8,256 |
| ICP-serve-flaggable 2025 stages | ~0 | 7,052 |

### Pledged 2025 candidacies — results breakdown

| stage_type | election_result | n |
|---|---|---|
| general | Won | 8,792 |
| general | Lost | 4,528 |
| general | Not on Ballot | 375 |
| general | No result recorded | 304 |
| general | Runoff | 108 |
| general | Withdrew | 9 |
| general runoff | No result recorded | 20 |
| primary | No result recorded | 3,475 |
| primary | Won | 1 |

<details>
<summary>SQL</summary>

```sql
select
  es.stage_type,
  coalesce(cs.election_result, 'No result recorded') as election_result,
  count(*) as n
from civics.candidacy_stage cs
join civics.candidacy c on cs.gp_candidacy_id = c.gp_candidacy_id
join civics.election_stage es on cs.gp_election_stage_id = es.gp_election_stage_id
where c.is_pledged
  and year(es.election_date) = 2025
group by 1, 2
order by 1, 3 desc;
```
</details>

### Where are all 9,680 pledged Wins?

The `candidacy` mart reports **9,680** pledged 2025 Wins (`candidacy.candidacy_result = 'Won'`). All of them are represented in both `candidacy` and `candidacy_stage` marts:

| layer | pledged 2025 Won | notes |
|---|---|---|
| `candidacy.candidacy_result = 'Won'` | 9,680 | source of truth |
| `candidacy_stage.election_result = 'Won'` (1:1 with above) | 9,680 | all accounted for |
| joined to `election_stage` (general, 2025) | 8,792 | 888 have `gp_election_id = NULL` and can't resolve |

The 888 gap is from 2025-archive candidacies whose HubSpot contacts lack the office metadata required to derive an election identity. They remain fully visible in `candidacy` and `candidacy_stage`; they just don't resolve through `election_stage` and therefore don't receive ICP flags. This is a known limitation of the 2025 archive's HubSpot data — the 2026+ BallotReady pipeline doesn't share the same shape.

## `gp_election_stage_id` grain — two parallel paths

Following a PR review, `int__civics_election_stage_2025` now unions two grains:

1. **DDHQ-matched path** — grain: `one row per ddhq_race_id`. Salting: `salted_uuid(ddhq_race_id)` (identical to pre-PR behavior for these rows). Distinguishes genuine collisions where two DDHQ races share a `gp_election_id` (e.g. Liberty CUSD At-large vs At-large Special; Pierce County Charter Review position numbers).
2. **HubSpot-only path** — grain: `(gp_election_id, stage_type)` where no DDHQ race exists at that combo. Salting: `salted_uuid(gp_election_id, stage_type)`. Irreducible floor for HubSpot-only data with no finer discriminator.

`int__civics_candidacy_stage_2025` uses the same conditional salting so the FK lines up in both paths.

## Breaking changes

- **`gp_election_stage_id` values change for HubSpot-only 2025 stages** (new synthetic salting `(gp_election_id, stage_type)`). DDHQ-matched 2025 IDs are unchanged from pre-PR. 2026+ BallotReady IDs are unchanged.
- **`gp_election_id` values for ~66k 2025 candidacies change.** Previously these shared a single "all-empty-fields" UUID (a collision bucket, not a real election); they now have `gp_election_id = NULL`. Any query that grouped on the ghost UUID will see NULLs instead.

## Known issues surfaced

- 272 HubSpot-only `gp_election_stage_id`s remain backed by multiple candidacy-stage dates (down from 283 pre-split-grain). These are HubSpot contests where two candidacies in the same gp_election_id disagree on the date. Reducing further requires expanding `generate_gp_election_id`'s hash fields — a separate ticket due to blast radius across all 11 callers.
- 144 general-stage rows have a `ddhq_election_stage_date` that diverges from `int__civics_election_2025.election_date` (up to 364 days). This reflects a genuine HubSpot-vs-DDHQ date disagreement upstream — same pattern exists pre-PR with different thresholds, not introduced by this change.
- 3 `gp_election_id`s have 11 / 757 / 9,767 distinct `br_position_database_id` values among their candidacies — likely UUID collisions from partially-populated contest rows. `max()` picks arbitrarily; affects <0.1% of stage rows.
- 6 small residual ghost `gp_election_id`s (covering a handful of candidacies each) still exist from contest rows that are present but have all-null identifying fields. The guard here only targets the LEFT JOIN miss.
- DDHQ-matched candidacies show ~0% coverage on `br_position_database_id` (0 of 2,274). Confirms the DDHQ path doesn't link to product-DB campaigns via email — ICP flagging for pure-DDHQ races still requires the full Phase B entity-resolution work from the original spike. Not in scope here.

## Test plan

- [x] `dbt build` across the full `+candidacy_stage +election_stage` chain (post-merge with main, including DATA-1714's BR+TS ER crosswalk) — all tests pass
- [x] Verified 0 `gp_election_stage_id`s with multiple distinct `ddhq_race_id`s (was 9 pre-split)
- [x] Verified all 9,680 pledged 2025 Wins survive into `candidacy` and `candidacy_stage` marts 1:1
- [ ] CI checks


dbt/project/models/intermediate/civics/int__civics_election_stage_2025.sql
```diff
@@ -1,95 +1,115 @@
...
-    valid_elections as (
-        select gp_election_id from {{ ref("int__civics_election_2025") }}
+    -- br_position_id is expected to agree within a (gp_election_id, stage_type);
+    -- 99.7% of stages show exactly one distinct value. max() is arbitrary on
+    -- disagreement.
+    stages_agg as (
+        select
+            gp_election_id,
+            stage_type,
+            max(stage_date) as election_stage_date,
```
sanjayr1: Grouping only by `(gp_election_id, stage_type)` means we now collapse multiple distinct dated stages into one `gp_election_stage_id` and keep the latest date via `max(stage_date)`. In the PR schema I found 283 stage IDs backed by multiple candidacy-stage dates (840 rows), while production has 0. There are also 144 general-stage rows where this chosen stage date no longer matches `int__civics_election_2025.election_date`, with deltas up to 364 days. Should this grain retain an additional discriminator so we don’t merge separate elections/stages?

dbt/project/models/intermediate/civics/int__civics_election_stage_2025.sql
```diff
@@ -1,95 +1,115 @@
...
-                else stage.election_stage
-            end as election_stage,
-            stage.ddhq_election_stage_date,
-            stage.ddhq_race_name,
-            stage.total_votes_cast,
-            stage._airbyte_extracted_at as created_at
-        from archived_election_stages as stage
+            c.gp_election_id,
+            {{ normalize_ddhq_stage_type("m.ddhq_election_type") }} as stage_type,
+            any_value(m.ddhq_race_id) as ddhq_race_id,
```
sanjayr1: This assumes there is only one DDHQ race per `(gp_election_id, stage_type)`, but that no longer holds after the grain shift above. In the PR schema I found 9 groups with 2 distinct DDHQ race IDs, so `any_value(ddhq_race_id/race_name/total_votes_cast)` makes the surviving DDHQ metadata arbitrary and drops the other race. Examples include Liberty CUSD 2 School Board (At-large vs At-large Special) and Pierce County Charter Review primaries (different position numbers). Do we need to preserve multiple rows here or add an explicit rule for choosing a canonical DDHQ race?

dbt/project/models/intermediate/civics/int__civics_election_stage_2025.sql
hhkarimi: Restructured in 40a3f14 — int__civics_election_stage_2025 now unions two parallel grains:
 - DDHQ-matched path: grain is ddhq_race_id, with salted_uuid(ddhq_race_id) (same as pre-PR). DDHQ-sourced date comes from stg_airbyte_source__ddhq_gdrive_election_results.election_date instead of candidacy-unpivoted dates.
  - HubSpot-only path: grain is (gp_election_id, stage_type) only when no DDHQ race exists at that combo. If DDHQ already covers it, we don't materialize a HubSpot-only row — those candidacies' gp_election_stage_id nulls via the valid_stage_ids gate in candidacy_stage.
  
Verified on the PR schema after rebuild:
 - Multi-date gp_election_stage_ids: 283 → 272. The residual 272 are pure HubSpot-only rows where two candidacies in the same contest disagree on the stage date — no DDHQ signal to split them further. Reducing this requires expanding generate_gp_election_id's hash fields (separate ticket; 11 callers + full-mart ID churn).
 - General-stage date-delta vs int__civics_election_2025.election_date: still 144 (max 364 days). These turned out to reflect a real HubSpot-vs-DDHQ date disagreement upstream — election_2025's date comes from contest/candidacy, stage's date now comes from DDHQ's actual race date. Pre-PR they happened to agree because the old stage model pulled from tbl_ddhq_matches.ddhq_date; swapping to the DDHQ source table surfaces the same underlying divergence. Logged in the PR's Known Issues.

dbt/project/models/intermediate/civics/int__civics_election_stage_2025.sql
hhkarimi: fully fixed in 40a3f14. DDHQ-matched stages now have grain ddhq_race_id (one row per race, no any_value coalescing across races). Liberty CUSD 2 At-large vs At-large Special, and both Pierce County Charter Review primaries, now materialize as separate gp_election_stage_ids. Their gp_election_stage_id values match pre-PR production since we fell back to salted_uuid(ddhq_race_id).

Verified: 0 gp_election_stage_ids with multiple distinct ddhq_race_ids (was 9). The only remaining any_value in the DDHQ path is within a single race — picking across candidates of that race for race-level attributes (ddhq_race_name, total_votes_cast), which is safe since those are race-level invariants.

sanjayr1 (approved):
LGTM

PR #333, opened 2026-04-22
feat(civics): elected officials mart with TS contact enrichment (DATA-1732)

sanjayr1:
## Summary

This PR ships the EO mart **after** restructuring per Dan's review feedback (comment thread on PR #333). The original architecture (wide BR+TS intermediate with mart-side coalescing) has been replaced with a candidacy-pattern merge: each source has its own person-grain intermediate, the person mart does the full-outer-join + COALESCE.

### Architecture

- **NEW** `int__civics_elected_official_canonical_ids` — deterministic `ts_officeholder_id ↔ BR canonical IDs` crosswalk with reuse flag.
- **NEW** `int__civics_elected_official_ballotready_person` — BR person-grain rollup (latest term per `br_candidate_id`).
- **NEW** `int__civics_elected_official_techspeed_person` — TS person-grain rollup (canonical adoption + reused-ID suppression + per-field FIRST_VALUE rollup + canonical-grain `is_incumbent` conflict detection).
- **CHANGED** `int__civics_elected_official_ballotready` — emits both `gp_elected_official_term_id` and `gp_elected_official_id` (person grain).
- **REFACTORED** `elected_official_terms` — BR-spine + TS provenance markers via crosswalk LEFT JOIN. Drops `is_incumbent`, `has_ts_person_enrichment`. Adds `ts_officeholder_id_is_reused`. Join-based `source_systems`.
- **REFACTORED** `elected_officials` — full-outer-join BR_person ⟕ TS_person with COALESCE per column. Drops `has_ts_person_enrichment`. Adds `is_incumbent`, `ts_incumbent_conflict`, audit columns. Join-based `source_systems`. ICP flags removed at person grain (Hugh's commit `8c22079`).
- **DROPPED** `int__civics_elected_official_ballotready_techspeed` — wide intermediate replaced by the new shape.
- **UNCHANGED** `int__civics_elected_official_techspeed` (preserves prematch contract).

### Key empirical values (PR #333 data)

| Metric | Value |
|---|---|
| BR term intermediate / term mart rows | 529,937 |
| Crosswalk rows | 30,375 |
| Reused TS IDs flagged | 86 |
| BR person intermediate rows | 375,321 |
| TS person intermediate rows | 27,827 |
| Reused-only canonicals dropped | 53 |
| TS conflict canonicals (is_incumbent NULL'd) | 350 |
| Phones preserved by per-field rollup vs single-row dedup | +107 |
| Emails preserved by per-field rollup vs single-row dedup | +1,097 |
| Person mart rows | 375,321 |

### Out of scope (deferred)

- **Prematch evolution** 
- **gp-api EO source integration** — adding gp-api as a third merge branch in `elected_officials.sql` is straightforward; deferred until PR #335 lands.
- **ICP at person grain** — RESOLVED. Hugh shipped commit `8c22079` removing ICP flags from `elected_officials`. Term-grain ICP remains.
- **Reused-ID compatibility scoring** — data quality follow-up.

### Test plan

- [x] Full chain `dbt build` — all 7 models PASS, all column/model/singular tests PASS
- [x] Crosswalk: `unique` on `ts_officeholder_id` and `br_office_holder_id`; deterministic-match `expression_is_true`; 2 singular tests
- [x] BR person: `unique` PK + audit columns + row-count singular test
- [x] TS person: `relationships` to BR person + conflict-implies-null model test + 2 singular tests
- [x] Term mart: `equal_rowcount` vs BR intermediate (529,937)
- [x] Person mart: `equal_rowcount` vs BR person intermediate (375,321)
- [x] FK integrity: every non-vacancy term has a person-mart match (`orphaned = 0`)
- [x] Source attribution: `['ballotready']` + `['ballotready','techspeed']` partition expected
- [x] Empirical counts match spec validation table


sanjayr1:
@claude review

sanjayr1:
bugbot run

sanjayr1:
bugbot run

sanjayr1:
Both bot review comments addressed in commit 5a17c8c:

1. **`merged.tier` → `merged.br_position_tier as tier`** — Fixed. The BR intermediate source SQL outputs `br_position_tier`; the mart now aliases it as `tier` in the output.

2. **`City` added to `office_level` accepted_values** — Fixed. Added to both `elected_official_terms` and `elected_officials` to match the upstream BR intermediate contract.

dbt/project/models/intermediate/civics/int__civics.yaml
```diff
@@ -1214,6 +1214,26 @@ models:
...
+        description: >
+          Whether this position meets Win ICP criteria (non-judicial,
+          non-appointed, 500-100K voters, recognized position type).
+          NULL when voter count data is unavailable. Forced false when
+          term_start_date predates icp_win_effective_date (or is null).
+
+      - name: is_serve_icp
+        description: >
+          Whether this position meets Serve ICP criteria (Win criteria
+          plus serve-eligible position type, 1K-100K voters). No
```
danpelota: What does "Win criteria plus serve-eligible position type" mean here?

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready.sql
```diff
@@ -84,6 +84,41 @@ with
...
+    -- has no effective-date gate today.
+    with_icp_flags as (
+        select
+            eo.*,
+            case
+                when
+                    icp.icp_win_effective_date is not null
+                    and (
+                        eo.term_start_date is null
+                        or eo.term_start_date < icp.icp_win_effective_date
```
danpelota: If serve has no effective-date gate, why do we filter on the elected official term start date and compare against the win effective date?

As I understand it, the Win ICP definition includes "School board races, but only those with election dates >= 2026-06-01". What does an EO term start date have to do with an election date? Which begs the question – should we be including Win ICP flags in this table at all? They're not necessarily candidates in upcoming races.

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready_techspeed.sql
```diff
@@ -0,0 +1,67 @@
+{{ config(materialized="table", tags=["civics"]) }}
```
danpelota: Does this need to be a table?

dbt/project/models/marts/civics/elected_official_terms.sql
```diff
@@ -0,0 +1,175 @@
...
+    merged.mailing_city,
+    merged.mailing_state,
+    merged.mailing_zip,
+
+    -- Metadata
+    merged.br_position_tier as tier,
+    merged.candidate_id_source,
+
+    -- source_systems: output-based — TS included only when a TS value survived
+    -- into a final output column (contact or direct-match fields)
```
danpelota: I think this is different from the pattern we use in the candidacy-related models, where we use this to mean "we have at least some record from this source." I think it makes sense to maintain that logic, since it would help us do things like select records that have not been returned in a match from Techspeed. If BallotReady provides an attribute, and then TechSpeed verifies that same attribute value, I think it's valuable to denote that records as both BallotReady and TechSpeed in the source_systems column.

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready_techspeed.sql
danpelota, line 1: Great insight to just directly join these and skip ER! However...I wanted to think through the pattern here of this join table and adding columns for each BR and TS attribute.

It seems like all we're doing here is a deduplication *within* Techspeed and then a join on the foreign key. I think we could do that pretty minimally feeding these records into the prematch table, no?

More importantly, having this "wide" table with hardcoded `ts_` and `br_` columns means we break the "merge"/survivorship pattern in at the mart layer. Everywhere else, we can have a crosswalk table with an arbitrary number of sources, and we can outer join + coalesce in the same way.

What would you think about a flow where we:

1) Add a model that just does the TS deduplication
2) When we add these records to the ER prematch table, we do the join on the BR id, and construct the array of values for phone/email fields with an array concatenation. So we still wind up with an array of email addresses and phone numbers, but every row has just a BR ID as its key.
3) The ER job runs and resolves the BR records to whatever else we put in there (product records for now). 
4) After the match, our thin crosswalk model on top of the match table just manually unions in the TS IDs to the existing BR clusters resulting from the match. The end result is a schema in looks exactly like the result of the Splink match everywhere else (one row per source+entity with a shared key)
5) We then do the merge/survivorship process at the mart layer like everywhere else

That means the survivorship lives in exactly one place: in the mart, on top of the crosswalk. And it looks the same regardless of how the cluster was formed. It also means the provenance of each row is auditable end-to-end, since every values origin is attached to a source row in the prematch.

dbt/project/models/marts/civics/elected_officials.sql
danpelota, line 1: I like this pattern of starting with our clean "fact" table (which his the gain we're working with from the underlying records from BR and TS) and then pulling the dimensions out of it 🙌

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready.sql
danpelota: At minimum, if we **do** include it (which is kinda dubious), maybe we should use the election date from the actual candidacy, since BR provides the candidacy ID for each EO?

dbt/project/models/intermediate/civics/int__civics.yaml
hhkarimi: fixed in e1348da. It never should have included a reference to "win" since "serve" criteria is treated independently

dbt/project/models/marts/civics/elected_officials.sql
```diff
@@ -1,46 +1,100 @@
...
+    person_rollup.district,
+
+    -- Latest term dates
+    person_rollup.term_start_date,
+    person_rollup.term_end_date,
+
+    -- Flags
+    person_rollup.is_judicial,
+    person_rollup.is_incumbent,
+    person_rollup.is_win_icp,
```
sanjayr1: Now i'm wondering if the `is_win_icp` should exist on the person-grain mart as a scalar copied from the latest term? ICP is office/race scoped, and PR data has 14,322 people with mixed Win ICP values across terms; do we want “latest term wins,” “any term qualifies,” “current term qualifies,” or should Win ICP stay only on `elected_official_terms`?  Or a column rename like `latest_term_is_win_icp`?
Or do we want to remove the win icp stuff from EO tables entirely? (i think it was me who added those columns in the first place 🤦 )

dbt/project/models/marts/civics/elected_officials.sql
```diff
@@ -1,46 +1,100 @@
...
+    person_rollup.is_serve_icp,
+    person_rollup.is_win_supersize_icp,
```
sanjayr1: ^^ Same as is_win_icp

dbt/project/models/marts/civics/elected_officials.sql
```diff
@@ -1,46 +1,100 @@
...
+
+    -- Latest term dates
+    person_rollup.term_start_date,
+    person_rollup.term_end_date,
+
+    -- Flags
+    person_rollup.is_judicial,
+    person_rollup.is_incumbent,
+    person_rollup.is_win_icp,
+    person_rollup.is_serve_icp,
```
sanjayr1: Should `is_serve_icp` on the person mart represent the latest selected term, any current/unexpired Serve ICP term, or should it stay term-scoped? 258 people have at least one current Serve ICP term while the selected latest-term person row is false, and 7 have selected true with no current Serve ICP term, so the current scalar is not clearly “currently serve-eligible.”
(Once again, i'm sure this is my doing. Claude 4.7 pointing fingers at Claude 4.6...)

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready.sql
```diff
@@ -84,6 +84,60 @@ with
...
+    --
+    -- is_serve_icp has no effective-date gate (Serve targets currently-seated
+    -- officials regardless of when they were elected).
+    with_icp_flags as (
+        select
+            eo.*,
+            case
+                when icp.icp_win_effective_date is null
+                then icp.icp_office_win
+                when ced.election_day is null
```
sanjayr1, line 136: Wondering if returning NULL for Win ICP when `election_day` is missing is the intended downstream contract? The comment says only ~24% of EO rows match a BR candidacy, so most EO terms will have unknown Win ICP rather than false.

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready.sql
hhkarimi: I can see this flag being useful for future planning of outreach. We could answer questions like "of elected officials, how many of them would be potential users to Win next year?"

Good idea to use the br candidacy directly to get the election_date. I implemented that. I'm open to removing the win ICP entirely if we think it will lead to confusion; or maybe it's something we want to add "after" removing the "effective_date" filter

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready.sql
hhkarimi: I got it the other way around: ~24% are null so majority are true/false due to missing match, based on joining to the staging table candidacies:
```
  with eo as (
      select eo.gp_elected_official_id, eo.is_win_icp, ced.election_day, icp.icp_win_effective_date
      from dbt_hugh.int__civics_elected_official_ballotready as eo
      left join dbt_hugh.stg_airbyte_source__ballotready_s3_candidacies_v3 as ced
          on eo.br_candidacy_id = cast(ced.br_candidacy_id as int)
      left join dbt_hugh.int__icp_offices as icp
          on eo.br_position_id = icp.br_database_position_id
  )
  select
      count(*) as total_rows,
      count_if(is_win_icp is null and election_day is null and icp_win_effective_date is not null) as null_from_unknown_election_day,
      count_if(is_win_icp is null and (election_day is not null or icp_win_effective_date is null)) as null_from_upstream_voter_count,
      count_if(is_win_icp is not null) as filled_count
  from eo
  ;
```

total_rows	null_from_unknown_election_day	null_from_upstream_voter_count	filled_count
529937	125075	3681	401181

dbt/project/models/marts/civics/elected_officials.sql
hhkarimi: Good point! I removed it entirely since I agree it adds confusion. It sounds like `election_official_terms` is THE table that will hold our results at the end.

dbt/project/models/marts/civics/elected_officials.sql
hhkarimi: Good point! I removed it entirely since I agree it adds confusion. It sounds like election_official_terms is THE table that will hold our results at the end.

dbt/project/models/marts/civics/elected_officials.sql
hhkarimi: Good point! I removed it entirely since I agree it adds confusion. It sounds like election_official_terms is THE table that will hold our results at the end.

dbt/project/models/intermediate/civics/int__civics_elected_official_ballotready.sql
danpelota: Good point – although that'll be relevant if we know they're interested and eligible to run for reelection the following year, since it's office-dependent. Not to be a broken record, but it also bolsters the case for removing election date as part of ICP definition since they *can't* be ICP if they're school board and we don't have a future election date for that position yet.

dbt/project/tests/assert_eo_ts_person_row_count.sql
danpelota, line 1: Love this 😍

danpelota (approved):
Looks great! 🚀

PR #335, opened 2026-04-22
[DATA-1719]  standardize gp-api intermediates for civics mart reconciliation

hhkarimi:
## Summary

Product Database (`gp_api_db`) becomes a first-class provider in the Civics Mart alongside BallotReady, TechSpeed, and DDHQ. This PR adds four intermediate models, extends the ER crosswalk, and adds two new columns surfacing PD-vs-BR date divergence.

## Changes

**New intermediate models** (all `materialized = "table"`, tagged `["civics", "gp_api"]`):
- `int__civics_candidate_gp_api` — one row per user with `campaign_count > 0`
- `int__civics_candidacy_gp_api` — one row per latest-version campaign with an election_date
- `int__civics_candidacy_stage_gp_api` — one row per (campaign × BR election_stage) for matching position + year
- `int__civics_elected_official_gp_api` — one row per `gp_api_db_elected_office` row (orphans filtered)

**ER crosswalk extended:** `int__civics_er_canonical_ids` gains nullable `gp_api_campaign_id` and `gp_api_stage_election_date` columns. The existing TS section is preserved (internal CTE renamed `stage_matches` → `ts_stage_matches`). A union of TS + gp_api sections means both providers now surface BR canonical IDs to their downstream intermediates.

**Campaigns mart:** adds `partisan_type` (from `details:partisanType`) alongside the existing `details:*` extractions, used for race-level `is_partisan`.

**New schema columns** (split-pair pattern: PD value source-faithful, BR canonical alongside):
- `int__civics_candidacy_gp_api.br_general_election_date` — BR's general-stage date for the canonical election; null when no ER match
- `int__civics_candidacy_stage_gp_api.br_election_stage_date` — BR's date for the canonical stage; null when no ER match

**New macros:**
- `generate_gp_api_gp_candidate_id` / `generate_gp_api_gp_candidacy_id` — centralize the salt-field lists used across the four gp_api intermediates (mirrors `generate_ts_gp_ids.sql`)
- `extract_district_geographic` (from main) applied to all 4 civics intermediates that share the district-parse contract

**Out of scope (followups):**
- Mart-layer union updates in `marts/civics/{candidacy,candidate,candidacy_stage}.sql`
- Removal of `int__civics_{candidacy,candidate,candidacy_stage}_2025` archive models
- Standalone `election` / `election_stage` models for Product DB (BR/TS own those grains)

## Notable implementation details

- **Person hash consistency across models.** `candidacy_gp_api` and `candidacy_stage_gp_api` source user fields (`first_name, last_name, email, phone`) from the `users` mart, not from the `campaigns` mart's denormalized `user_*` fields. The two marts are built independently, so their user snapshots can drift.
- **`gp_candidate_id` uses user-level state, not per-campaign state.** A user running in multiple states has one `user_state` (most recent non-demo campaign's state) so the referential-integrity inner join to `candidate_gp_api` survives multi-state users and demo-only users.
- **ER coalesce in `candidacy_gp_api.gp_candidate_id`.** Mirrors the window-propagation pattern in `candidate_gp_api`, so campaigns under a BR-clustered user adopt the same `canonical_gp_candidate_id` the candidate row resolved to.
- **BR stage fanout dedup.** A BR position can carry multiple races of the same `stage_type` in the same year (regular primary + special primary, etc.). Without a dedup, a single Product DB campaign fans across all of them. Mirrors the pattern in `int__er_prematch_candidacy_stages`, including the `br_race_id desc nulls last` tiebreaker for deterministic selection when two BR stages are equidistant.
- **`party_affiliation` normalized via `parse_party_affiliation`.** Raw Product DB values (`democrat`, `Independent Party`, `libertarian`, etc.) collapse to the canonical enum `{Independent, Nonpartisan, Democrat, Republican, Libertarian, Green, Other}`, matching BR/TS.
- **`is_partisan` is race-level**, derived from Product DB's `details:partisanType` (exposed as `campaigns.partisan_type`). Matches BR (`br_position.is_partisan`) and TS (`ts.partisan`) — an Independent candidate in a nonpartisan race correctly yields `is_partisan = false` here, same as BR/TS.
- **`office_level` is source-faithful.** Raw Product DB values `{city, county, federal, state}` pass through; downstream mart-union normalizes to BR's enum.
- **`candidate_id_tier = null`.** Product DB's `campaign.tier` holds WIN/LOSE/TOSSUP, not a viability tier — null until a real tier source lands, matching BR.
- **PD/BR date divergence surfaced as `br_*` columns.** When ER coalesces to a BR canonical ID, the PD source-faithful date stays in `general_election_date` / `election_stage_date`, and BR's canonical date is exposed alongside as `br_general_election_date` / `br_election_stage_date`. Resolves the previous (id, date) contradiction without losing PD's view.
- **Deterministic aggregates.** ER collapse CTEs use `max()` not `any_value()` so the two sibling models deterministically agree on canonical IDs even if ER clustering ever produces split canonicals.

### Row counts

| Model | Rows |
|---|---|
| `int__civics_candidate_gp_api` | 61,232 |
| `int__civics_candidacy_gp_api` | 42,666 |
| `int__civics_candidacy_stage_gp_api` | 20,568 |
| `int__civics_elected_official_gp_api` | 1,811 |
| `int__civics_er_canonical_ids` — TS rows | 9,439 |
| `int__civics_er_canonical_ids` — gp_api rows | 2,375 |
| `int__civics_er_canonical_ids` — total | 11,814 |

### Cross-source ID convergence (ER)

| Metric | Count |
|---|---|
| Distinct gp_api campaigns in `er_canonical_ids` | 1,687 |
| …flowing through to `candidacy_gp_api` | 1,584 (94%) |
| …whose `gp_candidacy_id` matches a row in `int__civics_candidacy_ballotready` | 1,584 (100%) |
| …whose `gp_candidate_id` matches a row in `int__civics_candidate_ballotready` | 1,584 (100%) |

100% of flowing-through campaigns match BR canonical on both candidate and candidacy grain.

**Rebuild note:** full chain rebuild required on this branch (`candidacy_stage_ballotready → er_canonical_ids → all four gp_api models`). Rebuilding only the gp_api models leaves the crosswalk holding stale BR canonical IDs.

### PD vs BR date divergence (split-column visibility)

| Metric | Candidacy (`general_election_date`) | Stage (`election_stage_date`) |
|---|---|---|
| Rows with BR canonical date populated | 1,578 | 1,601 |
| …of which differ from PD value | **56** | **352** |

Both views are now retained — consumers can compare PD's source-of-truth date against BR's canonical without losing either.

### Candidacy shape

| Metric | Count |
|---|---|
| Total candidacies | 42,666 |
| Demo | 2,201 |
| Verified | 23,080 |
| Pledged | 37,050 |
| Partisan (true) | 11,920 |
| Partisan (false) | 26,574 |
| Partisan (null) | 4,172 |
| With `candidacy_result = 'Won'` | 163 |
| With `br_position_database_id` | 38,316 |

`is_partisan` is race-level (true / false / null reflect the race's partisan flag, not the candidate's party).

### Party affiliation distribution (normalized)

| Value | Count |
|---|---|
| null | 16,321 |
| Independent | 13,560 |
| Nonpartisan | 10,344 |
| Other | 1,237 |
| Libertarian | 731 |
| Green | 471 |
| Democrat | 2 |

### Office level distribution (source-faithful Product DB values)

| Value | Count |
|---|---|
| city | 19,434 |
| null | 17,777 |
| county | 1,932 |
| federal | 1,852 |
| state | 1,671 |

Mart-layer union normalizes to the cross-source enum (Local/County/Township/State/Regional/Federal) at join time.

### Candidacy-stage cardinality

| Metric | Value |
|---|---|
| Distinct campaigns with at least one stage | 10,507 |
| Total stage rows | 20,568 |
| Avg stages per campaign | 1.96 |

### Elected official completeness

| Metric | Count |
|---|---|
| Total rows (after orphan filter) | 1,811 |
| With `term_start_date` (sworn_in_date) | 998 (55%) |
| With `party_affiliation` | 948 (52%) |
| With `email` | 1,811 (100%) |
| With `phone` | 1,810 (100%) |
| With `position_name` | 1,529 (84%) |

### Candidate identifying fields

| Field | Populated | % |
|---|---|---|
| `state` | 41,004 / 61,232 | 67% |
| `hubspot_contact_id` | 60,176 / 61,232 | 98% |
| `email` | 61,232 / 61,232 | 100% |
| `phone_number` | 61,100 / 61,232 | 100% |

The state-null rows are users whose only campaigns are demo.

### Test results

- **gp_api models + ER crosswalk:** 62 PASS / 0 WARN / 0 ERROR
- **TS regression:** 38 PASS / 0 FAIL — ER crosswalk extension didn't break downstream consumers

## Ticket
https://goodparty.clickup.com/t/90132012119/DATA-1719


hhkarimi:
@cursor review

hhkarimi:
@claude review

dbt/project/models/intermediate/civics/int__civics_elected_official_gp_api.sql
```diff
@@ -0,0 +1,133 @@
...
+            {{ clean_phone_number("phone") }} as phone,
+            cast(null as string) as office_phone,
+            cast(null as string) as central_phone,
+            position_name,
+            candidate_office,
+            nullif(election_level, '') as office_level,
+            {{ map_office_type("candidate_office") }} as office_type,
+            state,
+            cast(null as string) as city,
+            cast(null as string) as district,
```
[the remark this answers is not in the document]
hhkarimi: Wontfix — verified against the data:

- `extract_district_geographic(normalized_position_name)` yields **0 meaningful values** on the 1,803 elected_office rows. BR uses the raw `position_name` (e.g. "San Jose City Council - District 3") which carries the district suffix; Product DB exposes the *normalized* name (e.g. "City Council") which has the suffix stripped upstream. So `cast(null as string) as district` is byte-equivalent to extracting.
- `extract_city_from_office_name(normalized_position_name)` yields 267 values (15%), and it is substring-based — unreliable on normalized names. Thin, mixed-quality signal.
- The comparison to `int__civics_candidacy_gp_api` is slightly off: that model extracts city/district as hash inputs for `gp_election_id` / `gp_candidacy_id` internally (never output columns), so there is no output-column precedent to match.
- Matches the TS `elected_official` precedent (also null for city/district).

If downstream consumers want weak city-level attribution, the mart layer can apply the extraction uniformly across providers, with knowledge that the Product DB subset is low-coverage. Doing it here would imply parity with BR that doesn't actually exist.

hhkarimi:
@claude review once

dbt/project/models/intermediate/civics/int__civics_elected_official_gp_api.sql
```diff
@@ -0,0 +1,133 @@
...
+            sworn_in_date as term_start_date,
+            cast(null as date) as term_end_date,
+            cast(null as boolean) as is_appointed,
+            cast(null as boolean) as is_judicial,
```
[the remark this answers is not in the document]
hhkarimi: Applied in 6d02271, extending the fix to `candidacy_gp_api` and `candidacy_stage_gp_api` since the same raw-`campaign_party` issue was present there (and the `candidacy_stage` hash depends on it for consistency with `candidacy_gp_api.gp_candidacy_id`).

Summary of changes:
- `int__civics_candidacy_gp_api`: `{{ parse_party_affiliation("c.campaign_party") }} as party_affiliation`
- `int__civics_candidacy_stage_gp_api`: same (keeps hash alignment)
- `int__civics_elected_official_gp_api`: same
- Added `accepted_values` tests on `candidacy_gp_api.party_affiliation` and `elected_official_gp_api.party_affiliation` matching the BR/TS enum.

Post-normalization distribution on `candidacy_gp_api`: Independent: 13,564 · Nonpartisan: 10,347 · Other: 1,236 · Libertarian: 731 · Green: 471 · Democrat: 2 · null: 16,327.

Row counts shifted slightly (42,700 → 42,678 on candidacy; 21,067 → 21,047 on stage) from hash-collision dedup — e.g. a user whose raw party was `independent` previously hashed distinctly from one with `Independent`, but both now collapse to `Independent` and the second row is dropped. Semantically correct.

Different from the `office_level` source-faithful decision: `parse_party_affiliation` is a cleaning operation (same semantic enum, messy casing collapsed), not a taxonomy remap, so normalizing at the intermediate layer doesn't lose signal.

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
```diff
@@ -0,0 +1,208 @@
...
+            {{ extract_district_geographic("c.normalized_position_name") }} as district,
+            br.gp_election_stage_id as br_gp_election_stage_id,
+            br.stage_election_date
+        from latest_campaigns as c
+        inner join users as u on c.user_id = u.user_id
+        inner join
+            br_stages as br
+            on c.ballotready_position_id = br.br_position_id
+            and year(c.election_date) = br.stage_election_year
+    ),
```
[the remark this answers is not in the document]
hhkarimi: Applied in 003f775. Confirmed 509 campaign × stage_type pairs had 2 BR races in the join; the qualify dedup dropped ~486 phantom rows (21,047 → 20,561). Mirrored the prematch pattern (partition by campaign_id + stage_type, order by abs(datediff)).

Skipped the `is_disabled` filter — `int__civics_election_stage_ballotready` doesn't expose `is_disabled` in its output, so filtering here would require either bypassing the BR stage intermediate (inconsistent) or expanding its schema (out of scope for this PR). Worth a separate follow-up at the BR layer.

dbt/project/models/intermediate/civics/int__civics_candidacy_gp_api.sql
```diff
@@ -0,0 +1,249 @@
...
+            is_pledged,
+            is_verified,
+            cast(null as string) as verification_status_reason,
+            case
+                when party_affiliation is null
+                then null
+                when lower(party_affiliation) = 'nonpartisan'
+                then false
+                else true
+            end as is_partisan,
```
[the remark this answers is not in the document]
hhkarimi: Fair catch on the semantic divergence — updated the yaml description in 003f775 to document the candidate-level approximation and the BR/TS race-level divergence explicitly, and flagged the follow-up (exposing `details:partisanType` from the campaigns mart to switch to race-level semantics).

Keeping the SQL as-is for this PR — properly fixing needs a campaigns mart change, which is out of scope. The raw Product DB `details:partisanType` is populated (53k nonpartisan / 23k partisan / 27k null), so a follow-up PR can expose it and swap `is_partisan` to `case when partisan_type = 'partisan' then true when partisan_type = 'nonpartisan' then false else null end`.

dbt/project/models/intermediate/civics/int__civics_candidacy_gp_api.sql
```diff
@@ -0,0 +1,249 @@
...
+    er_canonical as (
+        select
+            gp_api_campaign_id,
+            any_value(canonical_gp_candidacy_id) as canonical_gp_candidacy_id,
+            any_value(canonical_gp_election_id) as canonical_gp_election_id,
+            any_value(canonical_gp_candidate_id) as canonical_gp_candidate_id
+        from {{ ref("int__civics_er_canonical_ids") }}
+        where gp_api_campaign_id is not null
+        group by gp_api_campaign_id
+    ),
```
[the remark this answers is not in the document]
hhkarimi: Applied in 003f775 — `any_value()` → `max()` in both `er_canonical` (candidacy_gp_api) and `er_canonical_candidacy` (candidacy_stage_gp_api).

Verified empirically: zero `gp_api_campaign_id`s in the crosswalk currently carry >1 `canonical_gp_candidacy_id`, so row counts are unchanged. But the defensive swap aligns the two siblings' collapses deterministically if future ER clustering ever produces a split canonical (primary and general stages of one campaign landing in different Splink clusters). Also consistent with `user_er_canonical` in the same file, which already uses `max()`.

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
```diff
@@ -0,0 +1,224 @@
...
+                    )
+                }}
+            ) as gp_candidacy_id,
+
+            -- gp_election_stage_id: adopt BR canonical from ER or fall back to
+            -- the BR stage we inner-joined on. Either way every row has a
+            -- valid gp_election_stage_id in int__civics_election_stage_ballotready.
+            coalesce(
+                canonical_gp_election_stage_id, br_gp_election_stage_id
+            ) as gp_election_stage_id,
```
sanjayr1, line 131: When ER supplies `canonical_gp_election_stage_id`, this row adopts the canonical BR stage ID but later still emits the pre-ER `stage_election_date` as `election_stage_date`. In the PR schema I found 351 rows where `gp_election_stage_id` points to a BallotReady election stage with a different `election_date`, including year-level mismatches. Should we either project the canonical stage date after choosing the canonical stage ID, or split this into canonical vs source stage date columns so downstream consumers do not see an ID/date contradiction?

dbt/project/models/intermediate/civics/int__civics_candidacy_gp_api.sql
```diff
@@ -0,0 +1,258 @@
...
+                    )
+                }}
+            ) as gp_candidate_id,
+
+            -- gp_election_id: salted UUID over (official_office_name, ..., state,
+            -- ..., election_date, seats_available).
+            -- Prefer ER canonical when available.
+            coalesce(
+                xw.canonical_gp_election_id, {{ generate_gp_election_id() }}
+            ) as gp_election_id,
```
sanjayr1, line 139: When ER supplies `canonical_gp_election_id`, this row adopts the canonical BR election ID but still keeps the Product DB campaign's `general_election_date`. In the PR schema, 67 ER-matched candidacies join to a BallotReady election whose `election_date` differs from this row's `general_election_date`, including year-level mismatches. Should we project the canonical election date with the canonical ID, or preserve the Product DB date in a clearly source-specific column?

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
```diff
@@ -0,0 +1,224 @@
...
+        -- DB campaign fans out across all of them (observed ~509 phantom rows
+        -- before fix). Pick the race whose date is closest to the campaign's
+        -- election_date — mirrors the pattern in
+        -- int__er_prematch_candidacy_stages.
+        qualify
+            row_number() over (
+                partition by c.campaign_id, br.stage_type
+                order by abs(datediff(br.stage_election_date, c.election_date))
+            )
+            = 1
```
sanjayr1, line 96: This `row_number()` is not deterministic when two BR stages have the same distance from the campaign election date. I found 389 `(campaign_id, stage_type)` tie groups before filtering, with 366 surviving into the output, so Spark can pick different stage IDs across rebuilds while tests still pass. Can we carry a stable BR identifier through `br_stages` and add a secondary order, similar to `int__er_prematch_candidacy_stages` using `br_race_id desc nulls last`?

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
hhkarimi: I'm now projecting the joined BR election date separately so both the gp-api value and the BR value are available in this intermediate model. I believe when the prod db value differs, it's due to an admin over write.

dbt/project/models/intermediate/civics/int__civics_candidacy_gp_api.sql
hhkarimi: Applied in 8874b78 with the same `br_*` split-column pattern as above. Kept `general_election_date` as Product DB's value and added `br_general_election_date` for BR's canonical view.

Note on derivation: BR doesn't have a single "general election date" at election grain — at the election grain you have stages with their own dates. So BR's canonical general date is derived by looking up the *general stage* of the canonical election, using the explicit `is_primary` / `is_runoff` flags:

```sql
br_general_election_date_lookup as (
    select gp_election_id, max(election_date) as br_general_election_date
    from {{ ref("int__civics_election_stage_ballotready") }}
    where not is_primary and not is_runoff
    group by gp_election_id
)
```

This mirrors the same derivation `int__civics_candidacy_ballotready` uses for its own `general_election_date`. `max()` guards against the rare multi-general-stage fanout per election.

Verified: 1,578 ER-matched candidacies now expose `br_general_election_date`; 56 differ from `general_election_date`. (Slightly fewer than your 67 — likely because some ER-matched elections have no general stage row in BR yet, leaving `br_general_election_date` null where you may have surfaced a different lookup.)

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
hhkarimi: Applied in 8874b78. Added `br_race_id` to the `br_stages` CTE projection and `br_race_id desc nulls last` as the secondary order on the `qualify row_number()`, mirroring the `int__er_prematch_candidacy_stages` pattern exactly.

```sql
qualify
    row_number() over (
        partition by c.campaign_id, br.stage_type
        order by
            abs(datediff(br.stage_election_date, c.election_date)) asc,
            br.br_race_id desc nulls last
    )
    = 1
```

No data correctness change today (no observed corruption — just non-determinism risk on rebuild), but the tiebreaker eliminates the rebuild-flip risk you flagged.

sanjayr1 (approved):
LGTM

PR #348, opened 2026-04-29
[DATA-1882] fix(civics): treat any future election date as WIN-ICP eligible

hhkarimi:
## Summary
Multi-part fix for the June School Board ICP miss reported in #data-civics-mart. Original bug: `is_win_icp` forced `false` whenever `general_election_date` was null or before `icp_win_effective_date`, mis-flagging candidacies whose primary or runoff was the qualifying future date. Investigation surfaced an entity-resolution gap and inconsistencies across downstream consumers; this PR addresses all of them in a single cohesive change.

### 1. WIN-ICP date gate (`win_icp_date_gate` macro)
- New project-level macro replaces six copies of the gate logic across `candidacy.sql`, `election.sql`, `leads_win_candidacy.sql`, and `m_techspeed__candidate_to_hubspot.sql`.
- New behavior: `false` only when **all four** election dates (primary, primary runoff, general, general runoff) are null or before the effective date. Otherwise fall through to the underlying ICP attribute.
- `leads_win_candidacy` now consumes `candidacy.is_win_icp` / `is_win_supersize_icp` directly (mart is the source of truth).

### 2. TS-only candidacy enrichment via `br_race_id`
TS staging captures `br_race_id` (100% populated) but the TS intermediates were discarding it. With the enrichment now in place:
- New `br_race_to_position_lookup()` macro centralises the `br_race_id → br_position_database_id + br_gp_election_id` derivation, shared by `int__civics_candidacy_techspeed`, `int__civics_election_techspeed`, and `int__civics_election_stage_techspeed`.
- `int__civics_candidacy_techspeed` outputs `br_position_database_id`. Primary/general dates use coalesce(TS form, BR election_stage); runoffs come from BR (TS doesn't capture them).
- `candidacy.sql` merges TS's enriched values via `coalesce(br.br_position_database_id, ts.br_position_database_id)`.

### 3. Stable IDs + 436-row admission via conditional substitution
TS-form parsed dates are substituted with BR-fallback dates ONLY when *both* TS dates are NULL. Rows that already had at least one TS date keep their original `gp_candidacy_id` (verified end-to-end). Rows with neither TS date (~436 across the cohort) now flow through with deterministic BR-derived hashes. Same pattern in the TS election + election_stage intermediates.

### 4. Candidacy ↔ election alignment
- All three TS intermediates adopt `brp.br_gp_election_id` as a fallback in the `gp_election_id` coalesce chain — TS-only rows whose `br_race_id` matches a BR election now share BR's `gp_election_id` end-to-end.
- `int__civics_election_techspeed` surfaces `br_position_database_id`; `election.sql` merges it via `coalesce(br, ts)`.

### 5. Tests + docs
- New singular test `assert_candidacy_win_icp_gate_no_future_false` catches WIN-ICP gate regressions, isolating "gate forced false" from "underlying ICP attribute is itself false".
- `int__civics.yaml` documents `br_position_database_id`, `primary_runoff_election_date`, `general_runoff_election_date` on the TS candidacy intermediate, with a relationships test on `br_position_database_id`.

## Impact

**Verification target** (representative June school-board candidacy): `gp_candidacy_id` unchanged (UUID-stable); `gp_election_id` now resolves to BR's id; `br_position_database_id`, full date set, and `is_win_icp = true` all populated.

**TS-only school-board cohort**:
| Metric | Before | After |
|---|---|---|
| Has `br_position_database_id` | 0 | 3,541 |
| Has `general_election_date` | ~0 | 6,952 |
| `is_win_icp = true` | 0 | 646 |

**Candidacy ↔ election alignment**:
| Metric | Pre-PR | After this PR |
|---|---|---|
| Candidacy has position but joined election doesn't | 10,189 | 367 |
| `candidacy.is_win_icp = true` but `election.is_win_icp != true` | 5,403 | 2 |
| `leads_win_candidacy.icp_office_win` differs from `candidacy.is_win_icp` | 230 | 0 |

**Whole candidacy mart**: 85,316 candidacies have `br_position_database_id`; `is_win_icp = true` count: 51,568. Residual 367/2 alignment gaps trace to pre-existing BR data quality issues (some BR election_stages don't roll up into BR election — same root cause as the `relationships_election_stage_gp_election_id` test, counts unchanged from baseline).

<details>
<summary>SQL used to generate the cohort numbers</summary>

```sql
select
    count(*) as ts_only_school_board,
    sum(case when br_position_database_id is not null then 1 else 0 end)
        as has_position,
    sum(case when general_election_date is not null then 1 else 0 end)
        as has_general,
    sum(case when is_win_icp then 1 else 0 end) as is_win_icp_true
from mart_civics.candidacy
where candidate_office ilike '%school board%'
  and array_contains(source_systems, 'techspeed')
  and not array_contains(source_systems, 'ballotready');
```

</details>

## Test plan
- [x] `dbt build` on the affected chain — all touched models build cleanly.
- [x] New `relationships_candidacy_br_position_database_id` and `relationships_election_br_position_database_id` tests PASS.
- [x] New `assert_candidacy_win_icp_gate_no_future_false` test PASSES end-to-end.
- [x] Spot-check on the verification target: stable UUID, position resolves, dates filled, ICP flag true.
- [x] Cohort metrics + alignment metrics measured before/after.
- [x] `leads_win_candidacy.icp_office_win` confirmed 0 disagreements with `candidacy.is_win_icp` after rebuild.
- [ ] Pre-existing failing tests (`gp_candidate_id`/`gp_candidacy_id`/`election_stage→election` relationships) verified to exist before this PR — not regressions.
- [ ] Confirm affected June school-board contacts flip to ICP in HubSpot after next sync.


hhkarimi:
@clause review once

hhkarimi:
@cursor review

dbt/project/models/marts/civics/candidacy.sql
```diff
@@ -139,9 +141,19 @@ select
     case
```
sanjayr1: `leads_win_candidacy` still duplicates the previous general-election-only ICP gate instead of using `candidacy.is_win_icp` / `is_win_supersize_icp` or the updated any-date gate.

In the PR schema I found 230 rows where `candidacy.is_win_icp=true` but `leads_win_candidacy.icp_office_win=false`; 209 of those have no general date but do have another qualifying future election date.

Can we either reuse the candidacy mart flags downstream or update `leads_win_candidacy` to match this new gate?

dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
```diff
@@ -109,6 +152,9 @@ with
         left join
             canonical_candidacy as xw
             on source.techspeed_candidate_code = xw.ts_source_candidate_id
+        left join br_race_to_position as brp on source.br_race_id = brp.br_race_id
+        left join
+            br_election_dates as bed on brp.br_gp_election_id = bed.br_gp_election_id
         where
```
sanjayr1, line 192: This predicate still requires a parsed TS primary/general date before the BR fallback dates can be used. That means TS rows with blank/unparseable TS dates but a valid `br_race_id` and BR election date are filtered out before `bed.br_*_election_date` can fill them.

I found 436 TechSpeed source rows with no parsed TS date but a matching BR race date available; 155 map to underlying WIN-ICP offices. Should this predicate allow `bed.br_primary_election_date` / `bed.br_general_election_date` too? If so, we should also check the candidacy/election ID hashing because it currently hashes the original TS parsed date.

dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
sanjayr1, line 138: The candidacy row is now enriched from BR race data, but `gp_election_id` still comes from the old TS/canonical election ID path, not the BR election ID resolved through `br_race_id`.

In the PR schema, 11,965 TS-only candidacies now have `br_position_database_id`, but 10,189 of their joined `election` rows still have null `br_position_database_id`, and 5,403 have `candidacy.is_win_icp=true` while `election.is_win_icp` is not true.

Should the TS election/election_stage models get the same BR race enrichment, or should this model adopt `brp.br_gp_election_id` when available so candidacy and election stay aligned?

sanjayr1:
I noticed `m_techspeed__candidate_to_hubspot` still has the old general-election-only ICP gate. If that model is still used for any HubSpot syncs before deprecation, it will not pick up this PR’s any-election-date behavior.

If we’re fully moving away from that model and the next HubSpot sync reads from the civics/analytics path instead, then I don’t think this needs to block the PR.

hhkarimi:
all four points in commit a26b326:

`m_techspeed__candidate_to_hubspot` old gate** — converted to use the macro (covered by #1).

### Impact (dbt_hugh schema, after the new commit)
| Metric | Pre-PR | Pre-commit a26b326 | Post-commit a26b326 |
|---|---|---|---|
| candidacy has `br_position_database_id` but joined election doesn't | 10,189 | 971 | **367** |
| `candidacy.is_win_icp=true` but `election.is_win_icp != true` | 5,403 | 23 | **2** |
| `leads_win_candidacy.icp_office_win` differs from `candidacy.is_win_icp` | 230 | (consumes mart) | **0** |

The remaining 367/2 residuals trace to pre-existing BR data quality gaps (some BR election_stages don't roll up into BR election — same root cause as the long-standing `relationships_election_stage_gp_election_id` failure, which counts unchanged or marginally improved between baseline and this PR).

Pre-existing test failures (`gp_candidate_id`/`gp_candidacy_id` relationships, `election_stage→election`) were verified to exist before this commit by running the same build chain on the prior PR state — none of them are regressions from this PR.

dbt/project/models/marts/civics/candidacy.sql
hhkarimi: leads_win_candidacy and m_techspeed__candidate_to_hubspot duplicating the gate — extracted to a win_icp_date_gate(icp_attribute, primary, primary_runoff, general, general_runoff, effective) macro. leads_win_candidacy now just consumes candidacy.is_win_icp directly (mart is the source of truth, no need for it to even call the macro). m_techspeed__candidate_to_hubspot and election.sql (also had the same old gate) call the macro. candidacy.sql switched to the macro too.

dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
hhkarimi: added `brp.br_gp_election_id` as a fallback in the `gp_election_id` coalesce chain in all three TS intermediates (`candidacy`/`election`/`election_stage`), and surfaced `br_position_database_id` from the TS election intermediate so election.sql can coalesce(br, ts) it. Centralized the lookup CTE in a `br_race_to_position_lookup()` macro to keep the three intermediates in lockstep.

dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
hhkarimi: instead of just relaxing the predicate, I substitute BR-fallback dates into source's `*_election_date_parsed` columns ONLY when both TS dates are NULL. Rows that already had at least one TS date keep their original `gp_candidacy_id` (verified end-to-end on the example candidacy). Rows with neither TS date now flow through with deterministic BR-derived hashes. Same conditional substitution applied to `int__civics_election_techspeed` and `int__civics_election_stage_techspeed` for consistency.

sanjayr1 (approved):
LGTM

PR #355, opened 2026-05-01
DDHQ matcha integration: cluster-based candidacy_stage merge

danpelota:
## Summary

Closes the TS+DDHQ-only-cluster gap in the candidacy_stage mart and a same-stage cluster integrity gap in the ER pipeline. Both came out of the 2026-04-30 matcha audit (companion PR in the matcha repo).

### Cluster-based merge in candidacy_stage

Until now `int__civics_er_canonical_ids` was the only mechanism that gave clustered records a shared join key, and it only handles BR-anchored pairs. Clusters with TS+DDHQ but no BR (1,435 in April 2026 alone) ended up as two separate mart rows because TS and DDHQ each fell back to their self-hashed `gp_candidacy_stage_id`.

The mart now wraps each provider's int model with a left join to `stg_er_source__clustered_candidacy_stages` to expose `cluster_id`, and the FOJ joins on `coalesce(cluster_id, 'self_' || gp_candidacy_stage_id)`. All cluster shapes merge correctly:

| Cluster shape | Before | After |
| --- | --- | --- |
| BR + TS + DDHQ | 1 row, source_systems = `[ballotready, techspeed, ddhq]` | unchanged |
| BR + DDHQ / BR + TS | 1 row | unchanged |
| **TS + DDHQ (no BR)** | 2 rows, `[techspeed]` + `[ddhq]` | **1 row, `[techspeed, ddhq]`** |
| DDHQ only | 1 row, `[ddhq]` | unchanged |

`canonical_ids` is no longer load-bearing for this mart but is still used by the candidacy / candidate / election_stage / election marts. Tracking full removal in DATA-1888.

### Per-provider native keys + relationships tests

The mart exposes:

- `br_candidacy_id`, `ts_source_candidate_id`, `ddhq_candidate_id`, `ddhq_race_id` — populated only when that provider is in the cluster
- `er_cluster_id` — Splink cluster_id when in ER, NULL otherwise (diagnostic only; not stable across reruns)

These drive `relationships` tests declared on each int model's natural key column pointing back to `candidacy_stage`. The DDHQ tests are configured `warn_if: ">0"` / `error_if: ">100"` because Splink occasionally over-clusters DDHQ rows with different `(candidate_id, race_id)` tuples and the mart's dedup keeps one row's natural keys (currently 5 + 20 records — small surface, useful signal).

### Office-name token comparison

Adds the `office_name_tokens` macro and a corresponding column on `int__er_prematch_candidacy_stages` so Splink can use `ArrayIntersectLevel` as a fallback office-overlap match level. Catches cross-source naming variants like DDHQ "Lincoln County R-IV School District" vs BR "Winfield R-4 School Board" — both yield an "r-4" token even though their JW similarity is below 0.88. Token normalization (parens preserved, roman→arabic, "no. N" → "r-N") lives in dbt so the inline string-split fallback that used to live in matcha's `BASE_POST_PREDICTION_FILTER` is gone.

### Same-stage cluster integrity test

`tests/assert_er_clusters_share_election_date.sql` is a singular test that fails if any cluster contains records from multiple `election_date`s. The candidacy_stage grain treats primary / runoff / general as distinct entities, so cross-date clusters are false positives that the mart relies on never happening (cluster_id-derived identifiers would conflate stages).

### Source identifier strategy

`__sources.yaml` now uses `identifier:` to point at the date-suffixed snapshot tables `clustered_candidacy_stages20260430` / `pairwise_candidacy_stages20260430`. Promoting a new matcha run is a single-line change (`identifier:`) without touching the staging models.

## Test plan

- [x] `dbt build -s candidacy_stage int__civics_candidacy_stage_ballotready int__civics_candidacy_stage_techspeed int__civics_candidacy_stage_ddhq` passes locally in `dbt_dball` (38 PASS / 2 WARN — DDHQ relationships tests under threshold / 0 ERROR)
- [x] `assert_er_clusters_share_election_date` passes (0 cross-date clusters)
- [x] Manual diff dbt_dball vs mart_civics: prod-only rows are accountable (DDHQ rows merged into 2-way clusters, plus normal staging-snapshot drift); no phantom missing rows
- [x] Spot-checked April 2026: `[techspeed, ddhq]` count went from 0 → 1,435 rows
- [ ] CI green
- [ ] dbt Cloud deploy build green after merge

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the join key and merge logic for the `candidacy_stage` mart (cluster-based FOJ across BR/TS/DDHQ), which can alter row counts/dedup behavior and downstream dependencies; it also updates ER source snapshot identifiers and adds new relationship/integrity tests that may affect builds.
> 
> **Overview**
> Updates the `candidacy_stage` mart to merge 2026+ BallotReady, TechSpeed, and DDHQ rows by Splink `cluster_id` (falling back to a per-row `self_` key) instead of relying on BR-anchored `canonical_ids`, fixing previously unmerged TS+DDHQ-only clusters. The mart now also surfaces diagnostic/provenance columns (`er_cluster_id` and each provider’s native keys) and adds `relationships` tests from each provider’s intermediate model to ensure those natural keys propagate into `candidacy_stage`.
> 
> Enhances entity-resolution prematch inputs by adding an `office_name_tokens` macro and a new `official_office_name_tokens` column for Splink `ArrayIntersectLevel` office-overlap matching. Points ER source tables at date-suffixed snapshot identifiers and adds a singular test (`assert_er_clusters_share_election_date`) to catch clusters that incorrectly span multiple `election_date`s.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50f05f710bf1d1f556380043ddb54aa51097aead. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/macros/office_name_tokens.sql
+++ b/dbt/project/macros/office_name_tokens.sql
@@ -1,25 +1,15 @@
 {% macro office_name_tokens(col) %}
-    -- Build a normalized array of meaningful tokens from an office-name string,
-    -- for use with Splink's ArrayIntersectLevel as a fallback office-overlap
-    -- match level (mirrors how first_name_aliases drives nickname matching).
+    -- Normalized token array for Splink ArrayIntersectLevel office-overlap.
     --
-    -- Punctuation is preserved on tokens (matches the existing locality-token
-    -- check in scripts/constants.py): "(juneau" stays distinct from "juneau"
-    -- so a WI town-of-X (county) ↔ X-county-supervisor pair correctly fails
-    -- to overlap, even though both contain a "juneau"-ish substring.
+    -- Punctuation is intentionally preserved so e.g. "(juneau" stays distinct
+    -- from "juneau" — a WI town-of-X (county) row must not overlap with an
+    -- X-county-supervisor row even though both contain a "juneau"-ish token.
     --
-    -- Steps:
-    -- 1. Lowercase
-    -- 2. Collapse "r- iii" (DDHQ form with internal space) → "r-iii"
-    -- 3. Inject "r-N" after "no. N" so DDHQ's MO "Reorganized School
-    -- District No. 4" emits an "r-4" token that overlaps with BR's
-    -- "Blue Springs R-4 School Board"
-    -- 4. Split on whitespace
-    -- 5. Per-token CASE: normalize roman-numeral school district codes
-    -- (r-i ↔ r-1, r-ii ↔ r-2, …) so DDHQ's "<county> County R-IV" shares
-    -- an "r-4" token with BR's "<city> R-4 School Board"
-    -- 6. Drop length≤1, pure-digit, and stop-word tokens (stop-word list
-    -- mirrors OFFICE_STOP_WORDS in scripts/constants.py)
+    -- Roman-numeral and "No. N" → "r-N" rewrites let DDHQ "Lincoln County
+    -- R-IV School District" intersect with BR "Winfield R-4 School Board".
+    --
+    -- Stop-word list mirrors OFFICE_STOP_WORDS in the matcha repo's
+    -- scripts/constants.py — keep both in sync when editing.
     filter(
         transform(
             split(
--- a/dbt/project/macros/office_name_tokens.sql
+++ b/dbt/project/macros/office_name_tokens.sql
@@ -1,9 +1,11 @@
 {% macro office_name_tokens(col) %}
     -- Normalized token array for Splink ArrayIntersectLevel office-overlap.
     --
-    -- Punctuation is intentionally preserved so e.g. "(juneau" stays distinct
-    -- from "juneau" — a WI town-of-X (county) row must not overlap with an
-    -- X-county-supervisor row even though both contain a "juneau"-ish token.
+    -- Trailing punctuation is stripped so "county," matches the "county" stop
+    -- word and "r-iv," normalizes to "r-4". Leading punctuation is preserved:
+    -- "(juneau" (parenthetical-county marker on a WI town-of-X row) must stay
+    -- distinct from "juneau" (substantive locality on a county-supervisor row)
+    -- to avoid a false-positive office overlap.
     --
     -- Roman-numeral and "No. N" → "r-N" rewrites let DDHQ "Lincoln County
     -- R-IV School District" intersect with BR "Winfield R-4 School Board".
@@ -12,13 +14,16 @@
     -- scripts/constants.py — keep both in sync when editing.
     filter(
         transform(
-            split(
-                regexp_replace(
-                    regexp_replace(lower({{ col }}), 'r-\\s+', 'r-'),
-                    'no\\.\\s+(\\d+)',
-                    'no. $1 r-$1'
+            transform(
+                split(
+                    regexp_replace(
+                        regexp_replace(lower({{ col }}), 'r-\\s+', 'r-'),
+                        'no\\.\\s+(\\d+)',
+                        'no. $1 r-$1'
+                    ),
+                    ' '
                 ),
-                ' '
+                t -> regexp_replace(t, '\\p{Punct}+$', '')
             ),
             t -> case
                 when t = 'r-i'
@@ -105,7 +110,7 @@
             'metro',
             'and',
             'for',
-            'no.',
+            'no',
             'odd',
             'unexpired'
         )
--- a/dbt/project/macros/office_name_tokens.sql
+++ b/dbt/project/macros/office_name_tokens.sql
@@ -1,0 +1,123 @@
+{% macro office_name_tokens(col) %}
+    -- Build a normalized array of meaningful tokens from an office-name string,
+    -- for use with Splink's ArrayIntersectLevel as a fallback office-overlap
+    -- match level (mirrors how first_name_aliases drives nickname matching).
+    --
+    -- Punctuation is preserved on tokens (matches the existing locality-token
+    -- check in scripts/constants.py): "(juneau" stays distinct from "juneau"
+    -- so a WI town-of-X (county) ↔ X-county-supervisor pair correctly fails
+    -- to overlap, even though both contain a "juneau"-ish substring.
+    --
+    -- Steps:
+    -- 1. Lowercase
+    -- 2. Collapse "r- iii" (DDHQ form with internal space) → "r-iii"
+    -- 3. Inject "r-N" after "no. N" so DDHQ's MO "Reorganized School
+    -- District No. 4" emits an "r-4" token that overlaps with BR's
+    -- "Blue Springs R-4 School Board"
+    -- 4. Split on whitespace
+    -- 5. Per-token CASE: normalize roman-numeral school district codes
+    -- (r-i ↔ r-1, r-ii ↔ r-2, …) so DDHQ's "<county> County R-IV" shares
+    -- an "r-4" token with BR's "<city> R-4 School Board"
+    -- 6. Drop length≤1, pure-digit, and stop-word tokens (stop-word list
+    -- mirrors OFFICE_STOP_WORDS in scripts/constants.py)
+    filter(
+        transform(
+            split(
+                regexp_replace(
+                    regexp_replace(lower({{ col }}), 'r-\\s+', 'r-'),
+                    'no\\.\\s+(\\d+)',
+                    'no. $1 r-$1'
+                ),
+                ' '
+            ),
+            t -> case
+                when t = 'r-i'
+                then 'r-1'
+                when t = 'r-ii'
+                then 'r-2'
+                when t = 'r-iii'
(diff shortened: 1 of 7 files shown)


danpelota:
@bugbot review

danpelota:
@claude review once

dbt/project/macros/office_name_tokens.sql
danpelota, line 1: There's a chance we want to push this down to the Splink layer(?) and it may be a little to specific to catch the recent batch of DDHQ races, but I figured it's probably worth the marginal improvement.

dbt/project/macros/office_name_tokens.sql
```diff
@@ -0,0 +1,113 @@
...
+                when t = 'r-ix'
+                then 'r-9'
+                when t = 'r-x'
+                then 'r-10'
+                else t
+            end
+        ),
+        t -> length(t) > 1
+        and not regexp_like(t, '^[0-9]+$')
+        and t not in (
```
danpelota, line 54: The idea here is to make the DDHQ match more accurate by considering overlapping words in the position name, so order doesn't matter. To do that, though, we want to remove non-helpful stop words.

I think experimenting with an embedding-based rule will help a lot here too.

dbt/project/models/intermediate/civics/int__civics.yaml
danpelota, line 1: These source-native identifiers are now pulled through so we can add some smoke tests: "Do all of these DDHQ/TechSpeed/BallotReady records actually make it into the mart?"

dbt/project/models/intermediate/civics/int__er_prematch_candidacy_stages.sql
```diff
@@ -419,6 +419,11 @@ select
     nullif(phone, '') as phone,
     try_cast(br_race_id as int) as br_race_id,
     nullif(lower(trim(official_office_name)), '') as official_office_name,
+    -- Normalized token array for Splink ArrayIntersectLevel office-overlap
+    -- match. Captures locality + r-N school-district codes (e.g. DDHQ
+    -- "lincoln county r-iv" → ["lincoln","r-4"], BR "winfield r-4 school
+    -- board" → ["winfield","r-4"]) so cross-source naming variants intersect.
+    {{ office_name_tokens("official_office_name") }} as official_office_name_tokens,
```
danpelota, line 426: Again, this mostly helps Splink add an array-intersect rule for words that actually matter, ignoring word order entirely.

dbt/project/models/marts/civics/candidacy_stage.sql
```diff
@@ -117,21 +166,23 @@ with
...
+                    case when br.merge_key is not null then 'ballotready' end,
+                    case when ts.merge_key is not null then 'techspeed' end,
+                    case when ddhq.merge_key is not null then 'ddhq' end
                 )
-            ) as source_systems
-        from {{ ref("int__civics_candidacy_stage_ballotready") }} as br
-        full outer join
-            {{ ref("int__civics_candidacy_stage_techspeed") }} as ts
-            on br.gp_candidacy_stage_id = ts.gp_candidacy_stage_id
+            ) as source_systems,
```
danpelota, line 180: This was the source of the previous bug - when a Splink-matched cluster of candidacy records didn't include a BallotReady row, the old BR-anchored crosswalk silently dropped the match — so TechSpeed-and-DDHQ-only or DDHQ-only candidacies appeared as duplicate rows in the mart instead of merging into one

dbt/project/models/staging/__sources.yaml
```diff
@@ -1520,6 +1512,10 @@ sources:
 
     tables:
       - name: clustered_candidacy_stages
+        # Date-suffixed snapshot from the 2026-04-30 matcha run (er-audit/
+        # prematch-candidacy-stages-2026-04-30 branch). Update the identifier
+        # when promoting a new run; staging models stay unchanged.
+        identifier: clustered_candidacy_stages20260430
```
danpelota, line 1518: This is temporary – I have a couple of tweaks coming in to the `matcha` code, and wanted to store the updated matches in a new table until that code was reviewed and we send the new matches to the original table.

dbt/project/models/marts/civics/candidacy_stage.sql
```diff
@@ -53,7 +57,12 @@ with
             cast(null as boolean) as is_uncontested,
             array_compact(
                 array('hubspot', case when has_match then 'ddhq' end)
-            ) as source_systems
+            ) as source_systems,
+            cast(null as string) as er_cluster_id,
+            cast(null as string) as br_candidacy_id,
+            cast(null as string) as ts_source_candidate_id,
+            cast(null as string) as ddhq_candidate_id,
```
sanjayr1: Should these 2025 archive rows populate `ddhq_candidate_id` and `ddhq_race_id` from `source_candidate_id` / `source_race_id` when `has_match` is true?

Right now `[hubspot, ddhq]` rows get `ddhq` in `source_systems`, but both new DDHQ native-key columns are always NULL for the archive path. In the PR build that affects 2,240 rows, so it seems to break the documented invariant that DDHQ keys are populated when `ddhq` is present.

dbt/project/macros/office_name_tokens.sql
```diff
@@ -0,0 +1,113 @@
...
+                when t = 'r-viii'
+                then 'r-8'
+                when t = 'r-ix'
+                then 'r-9'
+                when t = 'r-x'
+                then 'r-10'
+                else t
+            end
+        ),
+        t -> length(t) > 1
```
sanjayr1, line 52: Should the filter compare against a punctuation-stripped token while still returning the original token shape when needed?

Because the stop-word check is exact after splitting on spaces, punctuation-adjacent stop words survive, e.g. `county:` and `district,`. I also saw literal cases where `R-IV,` stays `r-iv,` and `No. 4,` becomes `r-4,`, so the intended `r-N` overlap can be missed. Would it be safer to normalize a comparison token for filtering/roman conversion, then preserve punctuation only where it is intentionally meaningful?

(5,159 / 117,474 prematch rows have punctuated stop-word tokens, about 4.4%, 1,505 of those rows are in non-singleton ER clusters, so this is not just dead singleton noise.)

sanjayr1 (dismissed):
LGTM

dbt/project/models/marts/civics/candidacy_stage.sql
danpelota: Good catch — fixed in 82e5db3. The archive_2025 CTE now forwards source_candidate_id/source_race_id into ddhq_candidate_id/ddhq_race_id when has_match is true. Verified on the latest build: 2,256/2,256 archive-with-'ddhq' rows populate both columns; the 173,066 archive rows without 'ddhq' stay NULL. Invariant holds.

dbt/project/macros/office_name_tokens.sql
danpelota: Fixed in 82e5db3 — went with the trailing-only strip you suggested.

Added a nested transform calling regexp_replace(t, '\\p{Punct}+$', '') before the stop-word filter and roman-numeral CASE, so 'county,' / 'r-iv,' / 'no. 4,' all normalize before the comparison checks. Stop-word entry 'no.' is now 'no' to match the post-strip form.

Anchored to $ so the leading '(' on '(juneau' is preserved — the parenthetical-county distinction from clean 'juneau' still holds (even though maybe it _shouldn't_ long term).

Spot-check on representative inputs:
- 'lincoln county, r-iv, school district' → ['lincoln','r-4'] (intersects clean BR 'winfield r-4 school board' on 'r-4')
- 'town of x (juneau county)' → ['(juneau']
- 'juneau county supervisor district 1' → ['juneau'] (still distinct from '(juneau')

dbt build of int__er_prematch_candidacy_stages + candidacy_stage: PASS=32 / WARN=6 / ERROR=0, identical warning counts to before.

dbt/project/models/marts/civics/candidacy_stage.sql
```diff
@@ -1,8 +1,12 @@
 -- Civics mart candidacy_stage table
 -- Union of 2025 HubSpot archive and 2026+ merged BallotReady + TechSpeed + DDHQ.
--- TS and DDHQ int models remap clustered IDs to BR canonicals via
--- int__civics_er_canonical_ids, so merging collapses to a full outer join on
--- shared gp_candidacy_stage_id.
+-- The 2026+ merge wraps each provider with its Splink cluster_id and joins
+-- on coalesce(cluster_id, 'self_' || gp_candidacy_stage_id), so any provider
+-- combination (BR+TS+DDHQ, BR+DDHQ, TS+DDHQ-only, DDHQ-only, singletons)
```
hhkarimi, line 5: Also gp-api/prod combinations?

hhkarimi (approved):

PR #364, opened 2026-05-05
feat(civics): use PD's ballotready_race_id for deterministic gp_api stage assignment

hhkarimi:
## Summary

PD asserts the specific BR race a user pledged to in `details:raceId` (base64-encoded `gid://ballotready/Race/<id>`, decoded numeric matches BR's race `database_id`). This PR adds a `ballotready_race_id` column on the `campaigns` mart and uses it to resolve `election_stage` / `br_race_id` / `gp_election_stage_id` for gp_api candidacy_stage rows via a deterministic ID join — replacing the prior closest-date guess against BR's race spine that fed BR-derived signal back into BR comparisons.

## Coverage

In-scope campaigns (`is_latest_version`, pledged, non-demo, `election_date >= 2026-01-01`):

| | Count | % |
|---|---:|---:|
| Total | 8,243 | 100% |
| With `ballotready_race_id` | 7,742 | **94%** |
| With `ballotready_position_id` | 7,656 | 93% |

Verified: 9/10 sampled raceIds decode to BR race database_ids that exist in BR's spine.

## Changes

- **`campaigns.sql`**: extract `ballotready_race_id` from `details:raceId` (`unbase64 + regexp_extract`, parallel to the existing `ballotready_position_id` extraction).
- **`int__er_prematch_candidacy_stages.sql`**: left-join `br_race` on `c.ballotready_race_id = br.br_race_id`. `election_stage` and `br_race_id` populated for ~99% of gp_api rows in dev; null for the ~6% without a raceId.
- **`int__civics_candidacy_stage_gp_api.sql`**: rewritten around the same ID lookup. Inner-join to `int__civics_election_stage_ballotready` on `br_race_id` for a deterministic `gp_election_stage_id`. Prefer ER-resolved canonical IDs when available, otherwise self-derive.
- **`int__civics.yaml`**: scope `not_null` on `election_stage` to non-gp_api sources; descriptions updated.

## Impact (dev)

| | Pre-PR (prod) | This PR |
|---|---:|---:|
| `int__er_prematch_candidacy_stages` gp_api rows | 20,590 | 10,724 |
| `br_race_id` populated on gp_api | closest-date guess | 99% (deterministic from PD) |
| `election_stage` populated on gp_api | closest-date guess | 99% (deterministic from PD) |
| `int__civics_candidacy_stage_gp_api` rows | 20,306 | 10,569 |

Halved row counts come from removing the prior fan-out across stage_types per campaign — each gp_api row now represents the single stage PD pledged to.

## Operational notes

No matcha-side change required: Splink keeps blocking on `br_race_id`, with even higher coverage than before (deterministic vs. closest-date guess).

## Relationship to #357

PR #357 has an earlier closest-date partition fix on `int__civics_candidacy_stage_gp_api`. After this PR merges, that file's rebased state will be the deterministic raceId version. Squash merge recommended for both PRs.

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
+++ b/dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
@@ -1,15 +1,10 @@
 {{ config(materialized="table", tags=["civics", "gp_api"]) }}
 
 -- Product DB campaigns -> Civics mart candidacy_stage rows.
--- Grain: one row per latest-version pledged gp_api campaign with a
--- ballotready_race_id that resolves in BR's election_stage spine. PD
--- asserts the specific race in details:raceId (~94% coverage on in-scope
--- campaigns); we look up gp_election_stage_id by race id deterministically
--- — no closest-date guessing. Campaigns without a raceId are dropped.
---
--- gp_candidacy_stage_id prefers the ER-resolved canonical when available
--- (gp_api clustered with BR/TS/DDHQ on this stage); otherwise self-derives
--- from (gp_candidacy_id, gp_election_stage_id).
+-- Grain: one row per latest-version pledged gp_api campaign whose
+-- ballotready_race_id resolves in BR's election_stage spine.
+-- gp_candidacy_stage_id prefers the ER-resolved canonical when available;
+-- otherwise self-derives from (gp_candidacy_id, gp_election_stage_id).
 with
     latest_campaigns as (
         select *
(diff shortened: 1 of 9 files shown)


dbt/project/models/intermediate/civics/int__er_prematch_candidacy_stages.sql
danpelota: After thinking about it, I'm not totally sure why we're joining BR race/position data to gp-api records at all in this prematch model.

From what I can tell, we're joining on race to get the election date and stage _closest_ to the date the user provided, which we then feed into the match. But ideally, the match determines whether two records are the same based on the input data, and providing this from BR kinda gives us false confidence in the signals we're providing. We're saying "the know this is the stage, position, and election date from gp-api, and use that to match against BR"...but in reality we've just "guessed" those values based on the closest election date.

For matching, we do want the signal we get from a deterministic `br_race_id` in gp-api records. But we do get that in the match process by blocking on `br_race_id` and then later deliberately dropping matches that have _different_ `br_race_id`. 

Any reason to not just drop all the joining on BR here entirely?

hhkarimi: It's a good point. Only `br_position_id` is known from gp-api rows, and that's set in the product so there's no guesswork there.

Going from `br_position_id` ->` br_race_id` is meant to support the downstream splink rules on `br_race_id` and `election_stage`.

But that sort of mixes the purpose of this data model with our specialized ER tool, and the rules on election_date should account for most of the signal related to election_stage.

I'll go ahead and remove them here.

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
```diff
@@ -1,120 +1,69 @@
...
-    -- max() not any_value(): deterministic, must match candidacy_gp_api.
-    er_canonical_candidacy as (
+    -- max() not any_value(): deterministic across runs, must align with
+    -- int__civics_candidacy_gp_api.
+    er_canonical as (
         select
             gp_api_campaign_id,
-            max(canonical_gp_candidacy_id) as canonical_gp_candidacy_id
+            max(canonical_gp_candidacy_id) as canonical_gp_candidacy_id,
+            max(canonical_gp_candidacy_stage_id) as canonical_gp_candidacy_stage_id
```
sanjayr1: This aggregation can pair a canonical candidacy_stage_id with the wrong raceId-derived gp_election_stage_id. In the PR catalog, 1,180 of 1,609 rows that adopt a BR stage ID end up with a gp_candidacy_stage_id whose BR row points to a different gp_election_stage_id than this row carries.

Should we preserve the canonical stage row relationship here instead of taking independent max() values? One option is to carry canonical_gp_election_stage_id through this CTE and only adopt canonical_gp_candidacy_stage_id when canonical_gp_election_stage_id = br.gp_election_stage_id; otherwise fall back to the self-derived salted ID?

sanjayr1 (approved):
LGTM

dbt/project/models/intermediate/civics/int__civics_candidacy_stage_gp_api.sql
hhkarimi: You're right — fixed in 6cb5035. Split back into two CTEs (which is what I had before the simplify-pass merge — I shouldn't have collapsed them):

- `er_canonical_candidacy`: candidacy-grain `max(canonical_gp_candidacy_id)` — same value across stages, so max is safe
- `er_canonical_stage`: carries `canonical_gp_election_stage_id` alongside `canonical_gp_candidacy_stage_id` so the join in `enriched` can match only when the canonical's stage equals the raceId-resolved `br.gp_election_stage_id`. Mispairs fall through to the self-derived salted UUID.

Verified locally: 437 rows now adopt a canonical id, 437 paired correctly, 0 mispaired (was 1,180 mispaired before).

sanjayr1 (approved):
Lgtm

PR #390, opened 2026-05-12
[DATA-1903] Local issue filter for district top issues

sanjayr1:
## Summary
- Expand `m_election_api__district_top_issues` from 27 issues to 68, with four jurisdictional flags (`is_local`, `is_regional`, `is_state`, `is_federal`) per issue.
- Drop the `QUALIFY issue_rank <= 10` cap on the mart so the election-api can filter by flag then re-rank within the filtered set.
- Move the issue universe, labels, and flags into a new `haystaq_issue_tags` dbt seed so non-engineers can iterate without touching SQL.

## Why
Today the mart serves jurisdictionally-mixed top issues to candidates and elected officials. A school-board candidate sees federal-only issues (war in Iran, federal jobs guarantee) alongside locally actionable ones. The new flags let the election-api filter to office-relevant issues per persona. Žak-driven ask via Samuel; Patrick approved the issue set; research-team-tested.

## Files changed
- `dbt/project/seeds/haystaq_issue_tags.csv` (NEW) — 68 rows; source of truth for the issue universe + labels + flags.
- `dbt/project/seeds/seeds_schema.yaml` — appended `haystaq_issue_tags` schema entry with explicit `column_types` for the four booleans, plus tests including an `expression_is_true` assertion that every row has at least one flag TRUE.
- `dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql` — rewritten: column list read from the seed via `dbt_utils.get_column_values` (so a seed edit is a one-file change), no `QUALIFY` cap, deterministic `issue_rank` tie-breaker on `issue asc`, seed join lifted into a CTE to avoid `generate_salted_uuid` ambiguity.
- `dbt/project/models/marts/election_api/m_election_api.yaml` — replaced the hardcoded 27-element `accepted_values` on `issue` with a `relationships` test against the seed; added `not_null` on the four flag columns; updated description to drop the "Top 10" framing and document the EXCLUDE NULLS grain.
- `airflow/astro/dags/sync_election_api.py` — temporary `WHERE issue_rank <= 10` guard on the Postgres sync (with comment pointing to the phase-(c) follow-up).

## Local commands
```bash
# From repo root:
cd dbt/project
dbt build --select "haystaq_issue_tags m_election_api__district_top_issues"
```
The quoted multi-model selector is required so the seed materializes before the model compile reads from it via `get_column_values`.

## Phase plan
- **Phase (a) — this PR**: dbt mart + seed ship; Airflow `WHERE issue_rank <= 10` keeps Postgres row volume bounded (~970K from 97K districts × top 10, up from today's ~470K). No API behavior change.
- **Phase (b) — election-api repo, owned by Patrick's team**: Prisma migration adds `is_local`, `is_regional`, `is_state`, `is_federal` as `BOOLEAN NULL` on `DistrictTopIssue` (nullable, no default — see `.tickets/data-1903/04-design.md` §14 for the column contract). Patrick will be pinged via Slack once this PR opens.
- **Phase (c) — small follow-up PR in this repo**: drop the `WHERE`, extend `DTI_COLUMNS` to 11 entries, swap hardcoded quality checks for shape-aware ones that query the seed row count from Databricks. Wait until Patrick's Prisma migration is in production.

## Validation
Full results: `.tickets/data-1903/07-validation-results.md`.

- dbt build PASS=24 ERROR=0. All 23 mart data tests pass (FK to district mart, FK to seed via `relationships`, `not_null` on every column, `unique_combination_of_columns`, `accepted_range`, `is_uuid`). Plus the seed-level `expression_is_true` flag invariant.
- Math correctness: for AK statewide, spot-checked 3 issues — mart score equals `AVG(hs_*)` over the upstream intermediate exactly (delta 0.0). Aggregation logic is right.
- Universe correctness for AK: dev mart has 92 AK districts. Samuel's prebuilt `private_samuel.district_top_issues_us_all` has 92 AK districts. Exact match at the available scope.
- Flag correctness: 0 mismatches across 6,256 shared rows with Samuel's table.
- Dev validation is AK-only because `dbt_sroberts.int__l2_nationwide_uniform_w_haystaq` was previously materialized with `l2_state_allowlist=AK`; dbt cloud's defer picks it up over prod. Prod-scale validation runs on the first prod dbt run after merge.
- Samuel-snapshot score drift (~0.02 absolute on a 0-100 scale) is normal Haystaq update churn between his build and today; not a logic concern.

## Design + history
- Design doc: `.tickets/data-1903/04-design.md`.
- Plan review findings: `.tickets/data-1903/06-plan-review-findings.md` (3 blockers + 7 strong recs all folded in or explicitly deferred).
- Brainstorming notes: `.tickets/data-1903/01-brainstorming-notes.md`.

## Test plan
- [x] CI passes (dbt build + tests in dbt cloud).
- [ ] Post-merge: prod dbt run produces ~6.6M rows in `goodparty_data_catalog.dbt.m_election_api__district_top_issues`.
- [ ] Post-merge: `sync_election_api` Airflow DAG loads ~970K rows into Postgres (bounded by the `WHERE issue_rank <= 10` guard), quality checks pass, swap succeeds, no downstream alerts.
- [ ] Post-merge: API behavior unchanged from today (top-10 issues per district, same ordering modulo any natural Haystaq drift).
- [ ] Coordinate Slack thread with Patrick for the Prisma migration; queue data-eng team on the schema-stability follow-ups in `04-design.md` §13.

---

> [!NOTE]
> **Medium Risk**
> Changes the shape and row volume of the `m_election_api__district_top_issues` mart (new columns, more issues, no top-10 cap), which can impact downstream syncs and consumers; Postgres sync is temporarily guarded but needs follow-up when schema catches up.
> 
> **Overview**
> **Expands `m_election_api__district_top_issues` to support jurisdiction-based filtering.** The mart now outputs *all* district/issue rows (removing the top-10 `QUALIFY` cap), adds `is_local`/`is_regional`/`is_state`/`is_federal` flags, and makes `issue_rank` deterministic with a tie-break on `issue`.
> 
> **Moves issue metadata into a new dbt seed.** Adds `haystaq_issue_tags` seed as the source of truth for the issue universe, labels, and flags; the model dynamically reads issue columns from the seed and joins it for metadata, and YAML tests switch from hardcoded `accepted_values` to a seed `relationships` check plus `not_null`/range assertions.
> 
> **Keeps Postgres sync stable during rollout.** The Airflow `sync_election_api` DAG temporarily adds `WHERE issue_rank <= 10` when loading `DistrictTopIssue` from Databricks to avoid increased volume until the Postgres schema is migrated to include the new flag columns.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a844d2b8fed82163196fa8b99f983640f368531. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
@@ -175,6 +175,12 @@ with
         -- Join seed metadata in a dedicated CTE so the final SELECT has only one
         -- source in scope, avoiding ambiguity when `generate_salted_uuid` passes
         -- unqualified column names into the SQL.
+        --
+        -- LEFT JOIN (not INNER) so issues in the inline `issue_columns` list
+        -- above that are missing from the seed produce rows with NULL labels
+        -- and flags, which the `not_null` and `relationships` tests on those
+        -- columns surface as loud test failures. With an INNER JOIN the same
+        -- drift would silently drop the issue's rows from the mart.
         select
             d.l2_state,
             d.l2_district_type,
@@ -188,7 +194,7 @@ with
             t.is_state,
             t.is_federal
         from district_issue_long as d
-        inner join {{ ref("haystaq_issue_tags") }} as t on t.issue = d.issue
+        left join {{ ref("haystaq_issue_tags") }} as t on t.issue = d.issue
     )
 
 select
--- a/dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
@@ -17,18 +17,97 @@
     defaults to EXCLUDE NULLS, so districts with NULL average scores for an
     issue emit no row for that issue.
 
-    Issue universe, labels, and jurisdictional flags come from the
-    `haystaq_issue_tags` seed — the seed is the source of truth. Edit the
-    seed to add, remove, or re-tag issues; rebuild the seed before rebuilding
-    this model so `dbt_utils.get_column_values` reads the new column list.
+    Issue universe is the `issue_columns` list below. Issue labels and the
+    four jurisdictional flags come from the `haystaq_issue_tags` seed, joined
+    at runtime.
+
+    Why two places: dbt cloud CI compiles the mart against a fresh PR schema
+    where the seed isn't materialized yet, so a `dbt_utils.get_column_values`
+    lookup against `ref('haystaq_issue_tags')` fails at compile time. The
+    inline list is the compile-time source of truth for the column set; the
+    seed is the runtime source of truth for labels and flags. The singular
+    test `tests/assert_district_top_issues_seed_coverage.sql` asserts every
+    seed issue appears in the mart, catching drift between the two if a new
+    issue is added to one place but not the other.
+
+    To add or remove an issue: edit both this list (keep it alphabetical) and
+    `seeds/haystaq_issue_tags.csv`. The singular test will fail loudly if
+    they drift apart.
 
     Downstream consumers (election-api) filter by jurisdictional flag and
     re-rank within the filtered set. The mart emits the overall `issue_rank`
     only.
 -#}
-{%- set issue_columns = dbt_utils.get_column_values(
-    ref("haystaq_issue_tags"), "issue", order_by="issue"
-) -%}
+{%- set issue_columns = [
+    "hs_abortion_pro_choice",
+    "hs_affordable_housing_gov_has_role",
+    "hs_age_limit_support",
+    "hs_aliens_governenment_hiding_much",
+    "hs_amazon_exploitative",
+    "hs_artificial_intelligence_excited",
+    "hs_autonomous_vehicles_allow",
+    "hs_casino_support",
+    "hs_charter_schools_support",
+    "hs_china_foreign_policy_advesarial",
+    "hs_civil_liberties_support",
+    "hs_climate_change_believer",
+    "hs_college_admissions_consider_race",
+    "hs_community_college_free_support",
+    "hs_critical_race_theory_books_ban",
+    "hs_crypto_increase_restrictions",
+    "hs_death_penalty_support",
+    "hs_defense_spending_increase",
+    "hs_dei_support",
+    "hs_doge_support",
+    "hs_econ_anxiety_very_worried",
+    "hs_family_medical_leave_support",
+    "hs_felon_voting_support",
+    "hs_gas_tax_support",
+    "hs_general_anti_vax_pro_vax",
+    "hs_gentrification_oppose",
+    "hs_gig_work_make_employees",
+    "hs_green_new_deal_support",
+    "hs_gun_control_support",
+    "hs_immigration_undesirable",
+    "hs_income_inequality_serious",
+    "hs_infrastructure_funding_fund_more",
+    "hs_insurance_of_last_resort_government_should_provide",
+    "hs_israel_military_actions_support",
+    "hs_jan_6th_pardons_support",
+    "hs_jobs_guarantee_support",
+    "hs_marijuana_legal_support",
+    "hs_mass_deporations_support",
+    "hs_medicaid_expansion_support",
+    "hs_medicare_for_all_support",
+    "hs_mexican_wall_support",
+    "hs_min_wage_15_increase_support",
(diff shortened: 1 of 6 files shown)


dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
```diff
@@ -141,15 +224,14 @@ select
...
-        {%- for column, label in issues %}
-            when '{{ column }}' then '{{ label }}'
-        {%- endfor %}
-    end as issue_label,
+    issue_label,
     score,
+    is_local,
+    is_regional,
+    is_state,
+    is_federal,
```
hhkarimi, line 240: Just noting that if we want to push these columns to the election-api dbs, we'll need update the schema in those rds tables here: https://github.com/thegoodparty/election-api/blob/develop/prisma/schema/districtTopIssue.prisma

dbt/project/seeds/haystaq_issue_tags.csv
```diff
@@ -0,0 +1,69 @@
+issue,issue_label,is_local,is_regional,is_state,is_federal
+hs_abortion_pro_choice,Pro-Choice on Abortion,True,False,True,True
+hs_affordable_housing_gov_has_role,Government Should Address Affordable Housing,True,True,True,True
+hs_age_limit_support,Support Federal Office Age Limits,False,False,False,True
+hs_aliens_governenment_hiding_much,Government Hiding Information About Aliens,False,False,False,True
```
hhkarimi, line 5: I believe L2 set these, so no typo change needed?
```suggestion
hs_aliens_government_hiding_much,Government Hiding Information About Aliens,False,False,False,True
```

hhkarimi (dismissed):

dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
```diff
@@ -6,51 +6,29 @@
...
+    this model so `dbt_utils.get_column_values` reads the new column list.
 
-{%- set issue_columns = issues | map(attribute=0) | list -%}
+    Downstream consumers (election-api) filter by jurisdictional flag and
+    re-rank within the filtered set. The mart emits the overall `issue_rank`
+    only.
+-#}
+{%- set issue_columns = dbt_utils.get_column_values(
+    ref("haystaq_issue_tags"), "issue", order_by="issue"
+) -%}
```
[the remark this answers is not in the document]
sanjayr1: Resolved in commit 611e897 — switched from `dbt_utils.get_column_values` to an inline Jinja list of the 68 issues exactly because of the compile-time relation-lookup failure this comment describes. The inline list can't be empty by construction, so the guard suggestion no longer applies. The seed is still load-bearing for the runtime join on `issue_label` and the four flags; a singular test (`tests/assert_district_top_issues_seed_coverage.sql`) plus the existing `relationships` test on `issue` catch drift in both directions between the inline list and the seed.

dbt/project/models/marts/election_api/m_election_api__district_top_issues.sql
sanjayr1: Yes, agreed — this is phase (b) in the rollout plan (`.tickets/data-1903/04-design.md` §11 and §14). Opening the election-api PR shortly with the four columns as `Boolean?` (nullable, no default) on `DistrictTopIssue`, positioned between `score` and `issueRank`. Patrick is in the loop and agrees.

The phase-(c) follow-up here drops the temporary `legacy_top_10` Airflow filter and extends `DTI_COLUMNS` to 11 entries once the Prisma migration is live in prod. That's the step that actually surfaces the flags to the API.

dbt/project/seeds/haystaq_issue_tags.csv
sanjayr1: Confirmed — L2 ships these column names with the typos (`hs_aliens_governenment_hiding_much` and also `hs_mass_deporations_support`), and our seed's `issue` values have to match the upstream `hs_*` column names exactly because the mart's AVG aggregation references them as live SQL identifiers. Changing only the seed would break the `relationships` test on `issue` and the AVG call would point at a non-existent column.

Folding the "typos-mirror-upstream" rule into the seed's YAML description and the mart SQL docstring in commit b3412d5, so a future contributor can't `s/governenment/government/` and accidentally break the mart. Keeping the spelling.

sanjayr1:
@delegate review

PR #392, opened 2026-05-13
[DATA-1917] Add HubSpot call disposition GUID -> label seed

hhkarimi:
## Summary

Adds a `hubspot_call_dispositions` seed mapping the 57 `hs_call_disposition` GUIDs currently in `stg_airbyte_source__hubspot_api_engagements_calls` to human-readable outcome labels (e.g. "Connected", "No answer", "Verbal Pledge - EO Waitlist"), plus a thin `int__hubspot_calls` model that joins the seed onto the calls staging so downstream readers can use `disposition_label` directly.

## Why

The Airbyte HubSpot connector ingests the raw disposition GUID but not the label mapping — that mapping lives in HubSpot Settings (Objects > Activities > Calls > "Call and meeting outcomes") and is not synced. Without this seed, computing connect rate or any disposition-segmented metric requires inspecting opaque GUIDs by hand.

Context: [#data-and-metrics thread](https://goodpartyorg.slack.com/archives/C0ATFPZTXGR/p1778693443655329) where the analytics owner flagged this.

## How rows were populated

Each row carries a `source` column so downstream readers can see provenance at a glance:

| source | rows | how |
| --- | ---: | --- |
| `hubspot_docs` | 6 | The documented HubSpot default outcomes (Busy, Connected, Left live message, Left voicemail, No answer, Wrong number). GUIDs are stable across accounts — see [HubSpot's calls API guide](https://developers.hubspot.com/docs/api-reference/legacy/crm/activities/calls/guide). |
| `inferred_from_title` | 22 | Custom outcomes whose label was extracted from the Nooks Call title pattern `[Nooks Call] - <label> - <name> - by <rep>`. Verified by per-GUID title fingerprinting via `dbt show`. |
| `unknown` | 29 | GUID is present in the data but its label could not be inferred from titles or call bodies. The `notes` column flags these for manual fill from HubSpot Settings. |

The 6 docs-verified GUIDs cover ~91% of dispositioned call volume; adding the 22 inferred customs gets to ~96%. The remaining 29 unknown GUIDs together account for ~3% of volume, and ~0.3% of calls have no disposition at all.

## What's in the PR

- `dbt/project/seeds/hubspot_call_dispositions.csv` — 57 rows, columns: `guid`, `disposition_label`, `source`, `notes`
- `dbt/project/seeds/seeds_schema.yaml` — seed description, `unique`+`not_null` on `guid`, `not_null`+`accepted_values` on `source`
- `dbt/project/models/intermediate/hubspot/int__hubspot_calls.sql` — thin left-join exposing `disposition_label` + `disposition_label_source` on the staging passthrough
- `dbt/project/models/intermediate/hubspot/int__hubspot.yaml` — adds the int model entry (description, `unique`+`not_null` on `id`, `accepted_values` on `disposition_label_source`)
- Co-locates existing hubspot intermediate models (`int__hubspot_contacts`, `int__hubspot_contacts_w_companies`, `int__hubspot_contest`) under `intermediate/hubspot/` to match the pattern used by the rest of the intermediate layer. Pure `git mv`; dbt `ref()`s are name-based so no downstream references change.

## Coverage check

Run against the materialized int model:

```sql
select disposition_label, disposition_label_source, count(*) as num
from {{ ref('int__hubspot_calls') }}
group by disposition_label, disposition_label_source
order by num desc
```

Sample of the output (top 15 of 30 rows; total 634,888 calls):

| disposition_label | disposition_label_source | num |
| --- | --- | ---: |
| No answer | hubspot_docs | 515,841 |
| Wrong number | hubspot_docs | 31,579 |
| _NULL_ | unknown | 21,811 |
| Hung up | inferred_from_title | 9,865 |
| Rejected - No need for services | inferred_from_title | 9,231 |
| Left voicemail | hubspot_docs | 9,174 |
| Verbal Pledge - Self-Directed | inferred_from_title | 6,703 |
| Verbal Pledge - Uncontested | inferred_from_title | 4,160 |
| Verbal Pledge - Meeting Booked | inferred_from_title | 3,608 |
| Connected | hubspot_docs | 3,519 |
| Busy | hubspot_docs | 3,368 |
| Rejected - Not Running | inferred_from_title | 2,868 |
| Call back later | inferred_from_title | 2,244 |
| _NULL_ | _NULL_ | 2,162 |
| Rejected-Uncontested | inferred_from_title | 1,372 |

The `NULL / unknown` row (21,811 calls) is the 29 GUIDs we couldn't infer labels for — the analytics owner will fill these from HubSpot Settings. The `NULL / NULL` row (2,162 calls) is rows where `hs_call_disposition` itself is unset.

## Test plan

- [x] `dbt seed --select hubspot_call_dispositions` — loaded 57 rows
- [x] `dbt test --select hubspot_call_dispositions` — 4/4 passed
- [x] `dbt build --select int__hubspot_calls` — view built, 4/4 tests passed (`unique`+`not_null` on `id`, `accepted_values` on `disposition_label_source`)
- [x] `dbt run-operation inspect_data --args '{"model": "int__hubspot_calls"}'` — 634,888 rows; `disposition_label` populated 96.2%, `disposition_label_source` 99.7%
- [ ] Analytics owner to spot-check the 22 `inferred_from_title` rows against HubSpot Settings
- [ ] Follow-up PR to fill the 29 `unknown` rows once the manual mapping is in hand

## Follow-ups

- Fill the 29 `unknown` seed rows from HubSpot Settings (one-time manual step).
- Consider a periodic check (dbt test or analysis) that flags new GUIDs appearing in the calls table that aren't yet in the seed.
- `stg_airbyte_source__hubspot_api_engagements_calls` carries `hs_call_has_voicemail` as a string; per the staging-layer-owns-casts convention, that could be cast to boolean in stg. Out of scope here.


dbt/project/models/intermediate/hubspot/int__hubspot_calls.sql
```diff
@@ -0,0 +1,10 @@
+{{
+    config(
+        auto_liquid_cluster=true,
+        tags=["intermediate", "calls", "hubspot"],
+    )
+}}
+
+select c.*, s.disposition_label, s.source as disposition_label_source
```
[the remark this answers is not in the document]
hhkarimi: Verified against the materialized seed on Databricks — empty CSV cells load as `NULL`, not empty string:

```
| total | is_null | is_empty_string | unknown_rows |
| ----- | ------- | --------------- | ------------ |
|    57 |      29 |               0 |           29 |
```

That matches the `disposition_label`-populated rate on the int model (96.2% = 610,915 / 634,888, the 29 unknown GUIDs leave 21,811 calls with NULL `disposition_label`). No `NULLIF` wrap needed — the YAML contract holds as written. Closing this thread.

dbt/project/models/intermediate/hubspot/int__hubspot.yaml
```diff
@@ -1,6 +1,40 @@
...
+          `inferred_from_title` (custom outcome derived from Nooks Call
+          title pattern), or `unknown` (GUID present but label not yet
+          mapped). Null when the call has no disposition.
+        data_tests:
+          - accepted_values:
+              arguments:
+                values:
+                  - hubspot_docs
+                  - inferred_from_title
+                  - unknown
```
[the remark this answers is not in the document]
hhkarimi: Good catch on the silent-join-failure risk — addressed in 14b5180, but via a different mechanism. The suggested `config.where: "hs_call_disposition is not null"` pattern is something this repo's CLAUDE.md tells us to avoid ("Most dbt tests do *not* need a `config.where: some_column_is_not_null`").

Instead, I added a model-level `dbt_utils.expression_is_true` that captures the same invariant — *if* `hs_call_disposition` is set, `disposition_label_source` must be set — without a `config.where`:

```yaml
data_tests:
  - dbt_utils.expression_is_true:
      arguments:
        expression: "hs_call_disposition is null or disposition_label_source is not null"
```

That fails loudly on the broken-join scenario (rows with a GUID but null source) while letting genuine "no disposition" rows pass. Passes locally.

dbt/project/seeds/hubspot_call_dispositions.csv
```diff
@@ -0,0 +1,58 @@
...
+7e731b63-354b-4beb-9ec4-c842bcb5449a,,unknown,TODO: fill from HubSpot Settings > Objects > Activities > Calls > Call and meeting outcomes
+59c6cd26-61e8-4ec6-abed-78b6e8469ca1,,unknown,TODO: fill from HubSpot Settings > Objects > Activities > Calls > Call and meeting outcomes
+63909de2-8d6e-40dc-8fdc-e5f9c410dd17,,unknown,TODO: fill from HubSpot Settings > Objects > Activities > Calls > Call and meeting outcomes
+af81669c-09fc-473b-aad5-be01193f40f4,Verbal Pledge - EO,inferred_from_title,"derived from ""[Nooks Call] - <label> - <name> - by <rep>"" pattern"
+66258670-5c50-4aa0-8ed4-d0f5b9187cee,Rejected - No Urgency,inferred_from_title,"derived from ""[Nooks Call] - <label> - <name> - by <rep>"" pattern"
+94cb249f-6687-4aa9-b912-cd1457c363cb,,unknown,TODO: fill from HubSpot Settings > Objects > Activities > Calls > Call and meeting outcomes
+a8a7e405-4abc-4442-b0b2-0b14b4906099,Rejected - No Time,inferred_from_title,"derived from ""[Nooks Call] - <label> - <name> - by <rep>"" pattern"
+d3cf69b0-ed13-400e-a113-e5ee7b9aa43d,Rejected - Not EO,inferred_from_title,"derived from ""[Nooks Call] - <label> - <name> - by <rep>"" pattern"
```
sanjayr1, line 37: [P2] I would manually verify this GUID before relying on the inferred label, or temporarily move it back to `unknown` with a note.

Blast radius is bounded but real: this GUID labels 327 calls in `goodparty_data_catalog.dbt_cloud_pr_70471823431465_392.int__hubspot_calls` (0.05% of all 634,888 calls; 0.71% of the 46,195 `inferred_from_title` calls). Among the title-bearing evidence rows for this GUID, 225 match `Rejected - Not EO`, but 46 rows from 2026-04-07 through 2026-04-20 match the existing `Rejected - Not Running` label instead. If the HubSpot setting confirms this GUID is `Rejected - Not EO`, this is fine to merge with a note; without that check, downstream rejection-segment counts can be subtly wrong even though row counts and CI stay green.

dbt/project/seeds/seeds_schema.yaml
```diff
@@ -170,6 +170,43 @@ seeds:
...
+      - name: source
+        description: "How this row was populated (see seed description)"
+        data_tests:
+          - not_null
+          - accepted_values:
+              arguments:
+                values:
+                  - hubspot_docs
+                  - inferred_from_title
+                  - unknown
```
sanjayr1, line 206: [P3] Add a provenance value for the planned manual fill path, e.g. `hubspot_settings`, before the follow-up work starts.

Current blast radius is documentation/test semantics only: the materialized seed still has 57 distinct GUIDs and the int model joins cleanly. The follow-up blast radius is larger: the 29 `unknown` seed rows currently cover 21,811 calls (3.44% of all calls). If those labels are later filled from HubSpot Settings, this enum either forces contributors to keep `source = unknown` on known labels, or makes the accepted-values test fail when they use a more accurate source. Low urgency for production data today, but cheap to address now and it preserves provenance.

dbt/project/models/intermediate/hubspot/int__hubspot_calls.sql
sanjayr1, line 8: [P3] Since this model is now the intended consumer-facing path for readable call outcomes, please also update the existing staging docs for `hs_call_disposition`.

The current staging description still says readers must map GUIDs by hand and that a follow-up may add a seed. Data blast radius is zero: `int__hubspot_calls` has the expected 634,888 rows, preserves unique IDs, and has no unseeded non-null dispositions. The practical blast radius is discoverability: analysts reading the staging docs can keep doing manual mapping instead of using this model. I would address this in the PR because the change is small and directly supports the feature; it is not a merge blocker if docs cleanup is tracked immediately.

dbt/project/seeds/hubspot_call_dispositions.csv
sanjayr1: Additional context for actionability: this was the largest ambiguous inferred-label case, but not the only one from the same validation query. The full mismatch summary I found was:

- `Rejected - Not EO` seed label vs `Rejected - Not Running` title-derived label: 46 rows, 2026-04-07 through 2026-04-20.
- `Verbal Pledge - EO Waitlist` seed label vs `Verbal Pledge - EO` title-derived label: 2 rows.
- `Rejected - No need for services` seed label vs `No answer` title-derived label: 1 row.

Suggested action: spot-check these three GUIDs in HubSpot Settings. If Settings confirms the seed labels, this is fine to merge with a note. If not, move any ambiguous row back to `unknown` until the manual mapping is confirmed.

Repro shape: join calls to `hubspot_call_dispositions` on `hs_call_disposition`, restrict to `source = 'inferred_from_title'` and `hs_call_title LIKE '[Nooks Call] - %'`, then match title prefixes against every non-null seed label with `hs_call_title LIKE concat('[Nooks Call] - ', disposition_label, ' - %')`; mismatches are where the seed label and title-derived label differ.

sanjayr1 (approved):
lgtm

PR #411, opened 2026-05-26
[DATA-1523] Phase 1, wire viability scoring into civics mart

sanjayr1:
## Summary

Phase 1 of three-PR DATA-1523 sequence. Wires the existing MLflow viability scoring (`int__techspeed_viability_scoring`) into `mart_civics.candidacy` via a LEFT JOIN on `techspeed_candidate_code`.

- `int__civics_candidacy_techspeed`: adds `viability_scoring` CTE and LEFT JOIN; replaces `cast(null as float) as viability_score` with the joined value; surfaces `score_viability_automated` (string category).
- `mart_civics.candidacy`: surfaces both columns (`viability_score`, `score_viability_automated`) through `archive_2025`, `merged_since_2026`, the `combined` UNION ALL arms, and the final SELECT.
- `m_civics.yaml`: expands `viability_score` description; adds `score_viability_automated` doc + `accepted_values` test; adds `accepted_range(0..5)` test on `viability_score`.
- `tests/mart_candidacy_techspeed_viability_coverage_warn.sql`: new B3 threshold test — warn at >30% missing (15,524 rows), error at >50% (25,874 rows). Default `severity='error'` so `error_if` actually fires.

PR 2 (to follow) fixes the upstream name-parsing bug that prevents the remaining ~26% of TS rows from matching, taking coverage to ~99.7%.

## Why this PR alone is loose-thresholded

Distinct-code match rate today: ~65% (39,136 / 59,860). The plan expected ~74%; the gap is a freshness mismatch — `int__techspeed_viability_scoring` last refreshed 2026-05-24 while `int__civics_candidacy_techspeed` rebuilds with current TS data. Either the next scheduled run of viability scoring narrows the gap, or PR 2's parser fix replaces the bottleneck entirely (since codes shift under the parser change, post-Phase-2 coverage will be set by whatever matches against the freshly-rebuilt viability table, not by today's snapshot).

## Test plan

Local environment couldn't run the mart-level `dbt build` due to an unrelated dev-schema deferral cascade on `int__civics_candidacy_gp_api`. CI will rebuild from scratch against deferred prod.

After CI builds in its preview schema:
- [ ] `int__civics_candidacy_techspeed` and `candidacy` build cleanly
- [ ] B1 (`accepted_range(0..5)` on `viability_score`) passes
- [ ] B2 (`accepted_values` on `score_viability_automated`) passes
- [ ] B3 (coverage threshold) passes — well below warn threshold
- [ ] Spot check: ≥10 TS-source rows on candidacy with non-null `viability_score` and matching string category

What I verified locally before pushing:
- [x] `int__civics_candidacy_techspeed` builds; pre-existing tests still pass
- [x] 10-row spot check on `int__civics_candidacy_techspeed` returns non-null `viability_score` (0..5) with matching `score_viability_automated`
- [x] `dbt parse` + `dbt compile --select mart_candidacy_techspeed_viability_coverage_warn` succeed
- [x] yaml `accepted_values` list matches the branching in `int__techspeed_viability_scoring.py` (5 categories, exact strings/order)

## Design

`.tickets/DATA-1523/20_viability_completeness_design_v2.md` (canonical spec) and `.tickets/DATA-1523/24_viability_completeness_plan_v3.md` (this PR = Tasks 2-6).

---

> [!NOTE]
> **Medium Risk**
> Adds new mart columns and changes how `viability_score` is populated for merged candidacies; impact is analytical/data-contract rather than runtime security, with partial TS coverage expected until a follow-up PR.
> 
> **Overview**
> **Phase 1 (DATA-1523)** connects existing TechSpeed MLflow viability output into the civics **`candidacy`** mart instead of leaving `viability_score` null for TechSpeed rows.
> 
> `int__civics_candidacy_techspeed` adds a `viability_scoring` CTE and **LEFT JOIN** on `techspeed_candidate_code`, mapping `viability_rating_2_0` → `viability_score` and exposing **`score_viability_automated`**. Minor qualify/filter fixes qualify `candidate_code` via `source`.
> 
> `candidacy` passes both fields through archive and 2026+ merge paths: **`viability_score`** becomes `coalesce(br.viability_score, ts.viability_score)` (BR stays null upstream; TS supplies the score), and **`score_viability_automated`** comes from TS only (2025 archive gets `cast(null)`).
> 
> `m_civics.yaml` documents the columns and adds **`accepted_range(0..5)`** on `viability_score` and **`accepted_values`** on the five automated category strings. New test **`mart_candidacy_techspeed_viability_coverage_warn`** fails if more than ~30% of TechSpeed-tagged mart rows lack `viability_score` (loose until a follow-up PR tightens join coverage).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a530bdd4333f01bb6d359b88759a18e52397c5aa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics_candidacy_2025.sql
+++ b/dbt/project/models/intermediate/civics/int__civics_candidacy_2025.sql
@@ -87,6 +87,7 @@ with
 
             -- assessments
             viability_scores.viability_rating_2_0 as viability_score,
+            viability_scores.viability_rating_automated as score_viability_automated,
             cast(cast(tbl_companies.win_number as float) as int) as win_number,
             tbl_companies.win_number_model,
 
@@ -180,6 +181,7 @@ select
     general_runoff_election_date,
     br_position_database_id,
     viability_score,
+    score_viability_automated,
     win_number,
     win_number_model,
     has_ddhq_match,
--- a/dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
+++ b/dbt/project/models/intermediate/civics/int__civics_candidacy_techspeed.sql
@@ -108,6 +108,11 @@ with
             = 1
     ),
 
+    viability_scoring as (
+        select techspeed_candidate_code, viability_rating_2_0, score_viability_automated
+        from {{ ref("int__techspeed_viability_scoring") }}
+    ),
+
     candidacies as (
         -- gp_candidate_id and gp_election_id use WINDOW propagation so all raw
         -- rows of the same person/election adopt BR's canonical if ANY candidacy
@@ -148,7 +153,7 @@ with
             cast(null as string) as hubspot_company_ids,
 
             'techspeed' as candidate_id_source,
-            techspeed_candidate_code as candidate_code,
+            source.techspeed_candidate_code as candidate_code,
 
             party as party_affiliation,
             is_incumbent,
@@ -178,7 +183,8 @@ with
 
             source.br_position_database_id,
 
-            cast(null as float) as viability_score,
+            vs.viability_rating_2_0 as viability_score,
+            vs.score_viability_automated,
             cast(null as int) as win_number,
             cast(null as string) as win_number_model,
 
@@ -189,8 +195,11 @@ with
         left join
             canonical_candidacy as xw
             on source.techspeed_candidate_code = xw.ts_source_candidate_id
+        left join
+            viability_scoring as vs
+            on source.techspeed_candidate_code = vs.techspeed_candidate_code
         where
-            techspeed_candidate_code is not null
+            source.techspeed_candidate_code is not null
             -- After the source CTE's BR fallback substitution, this passes
             -- rows that had EITHER a TS date OR a BR-resolved date.
             and coalesce(general_election_date_parsed, primary_election_date_parsed)
--- a/dbt/project/models/marts/civics/candidacy.sql
+++ b/dbt/project/models/marts/civics/candidacy.sql
@@ -65,6 +65,7 @@ with
             is_incumbent,
             office_type,
             br_position_database_id,
+            cast(null as string) as score_viability_automated,
             array_compact(
                 array('hubspot', case when has_ddhq_match then 'ddhq' end)
             ) as source_systems
@@ -141,6 +142,7 @@ with
             br.viability_score,
             br.win_number,
             br.win_number_model,
+            ts.score_viability_automated,
             array_compact(
                 array(
                     case when br.gp_candidacy_id is not null then 'ballotready' end,
@@ -193,6 +195,7 @@ with
             is_incumbent,
             office_type,
             br_position_database_id,
+            score_viability_automated,
             source_systems
         from archive_2025
         union all
@@ -226,6 +229,7 @@ with
             is_incumbent,
             office_type,
             br_position_database_id,
+            score_viability_automated,
             source_systems
         from merged_since_2026
     ),
@@ -351,6 +355,7 @@ select
             effective_date="icp.icp_win_effective_date",
         )
     }} as is_win_supersize_icp,
+    deduplicated.score_viability_automated,
     deduplicated.source_systems,
     deduplicated.created_at,
     deduplicated.updated_at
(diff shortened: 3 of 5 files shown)


sanjayr1:
Thanks to Cursor Bugbot and delegate-reviewer for catching these. Here's the disposition:

### Blocker 1: Mart was discarding `ts.viability_score` (line 142) — fixed in a530bdd

Replaced `br.viability_score,` with `coalesce(br.viability_score, ts.viability_score) as viability_score,` and rewrote the now-misleading "BR-only viability fields" comment. BR doesn't actually compute viability_score (always NULL upstream), so the coalesce simply lets the TS MLflow score reach the mart.

After the fix, B3 coverage test should WARN (~35% TS nulls remain due to the upstream candidate_code mismatch — see Blocker 3) but not ERROR.

### Blocker 2: WoE fan-out risk in `int__techspeed_viability_scoring.py:88–92` — separate ticket

The fan-out concern is real in principle, but the model is outside DATA-1523's scope (ML scoring pipeline, not civics mart). Row count of `int__techspeed_viability_scoring` matches expected distinct `techspeed_candidate_code`s, so no observed fan-out today. Filing as its own ticket for the ML pipeline maintainers.

### Blocker 3: `techspeed_candidate_code` mismatch with viability_scoring (line 78-87) — deferred to PR 2

This is the exact 26% gap that DATA-1523 Phase 2 targets via the parser fix in staging (`stg_airbyte_source__techspeed_gdrive_candidates` replaces `trim(last_name)` with a chained regex parser, and `int__techspeed_candidates_clean` refactors to consume the already-parsed `last_name`). Once Phase 2 lands, the inline `generate_candidate_code(... ts.last_name ...)` in this file produces the same code as `int__techspeed_candidates_clean` and `int__techspeed_viability_scoring` — coverage rises to ~99%.

Reviewer's alternative (JOIN to `int__techspeed_candidates_clean` directly) would close the gap in PR 1 but routes around the divergence rather than unifying the codebase on parsed staging. Sticking with the staged plan keeps PR 1 mechanical and Phase 2 architecturally clean.

Plan reference: `.tickets/DATA-1523/24_viability_completeness_plan_v3.md` Tasks 8-13.

delegate review

sanjayr1:
bugbot run

sanjayr1:
Round 2 dispositions:

### delegate-reviewer line 5: severity="warn" required — disagreeing

The bot is reading dbt's severity semantics inverted, and the empirical record on this PR contradicts it:

- Initial commit \`a121a595\` (no severity set, default = error): ERRORED with "Got 51748 results, configured to fail if >25874" — proves \`error_if\` IS evaluated.
- Fix commit \`a530bdd4\` (no severity change): WARNED at 15,679 missing — between the 15,524 warn threshold and the 25,874 error threshold. Proves \`warn_if\` is ALSO evaluated.

Per dbt's severity docs, the default \`severity: error\` evaluates \`error_if\` first then \`warn_if\`; \`severity="warn"\` would CAP max severity at warn and silently turn the \`>25874\` error gate into a no-op. The other repo tests that set both thresholds AND \`severity="warn"\` are effectively running only their warn_if (their error_if is dead code). I'll leave a note on whether those want a follow-up.

I've expanded the inline comment on the test to make this explicit so future readers don't re-litigate.

### Independent review pass — 3 fixes shipped in 34711d7

- **archive_2025 score_viability_automated** populated. ~9,928 archive rows had \`viability_score\` non-null but \`score_viability_automated\` was hardcoded NULL. The upstream \`stg_model_predictions__viability_scores\` already exposes \`viability_rating_automated\`; just had to surface it through \`int__civics_candidacy_2025\`.
- **m_civics.yaml:400** model-level precedence docs updated: viability_score is no longer "BR-only" after Blocker 1; new "TS-derived (BR/DDHQ/gp_api don't compute viability)" group covers viability_score and score_viability_automated.
- **B3 comment** clarified that thresholds are absolute counts pegged at PR-1 baseline (~51,748 TS rows); PR 2 Task 12 explicitly recomputes them. Percentages-from-denominator would mask regressions as TS volume grows, which defeats the test's purpose.

delegate review

PR #430, opened 2026-06-01
Fix misleading win_number comments and add coverage guard

hhkarimi:
## What

- Corrects misleading comments that claimed `win_number` comes from BallotReady, in `int__civics_candidacy_ballotready.sql` and `m_election_api__race.sql` (comment-only, no logic change).
- Adds `tests/mart_election_api_race_win_number_estimate_coverage_warn.sql`, a warn/error coverage guard.

## Why

`civics_win_number` is null for 100% of races the election_api serves, which surfaced as confusion downstream (whether `win_number_estimate` was redundant, whether the field was supposed to be null). Root cause, verified against the warehouse:

- **BallotReady supplies no win number.** No BR source (API race, S3 candidacies, viability_scores) carries one. The `cast(null as int) as win_number` comment ("hardcoded until we join viability/p2v in followup") implied a planned join that has no source to join to.
- The only `win_number` values that ever existed are HubSpot-archive rows for 2023-2025 elections (`int__civics_candidacy_2025`). The race mart's forward-looking date filter excludes those, so `win_number` is null for every row the mart emits.
- The `gp_election_id` join in the race mart is also low-coverage, but that is a symptom; even a perfect join finds nothing to attach because there is no current-cycle source.

So the served `win_number_effective` is always `win_number_estimate`, which is computed from projected turnout. That estimate must stay (it is the only live win number), and the thing worth guarding is its input.

## The guard test

Tracks the share of upcoming races whose district has no positive projected turnout for the election year (the races that would serve a null win number and null `contacts_needed_estimate`). Warns at >15%, errors at >25%. Baseline is ~7.7%, so it passes today and will catch a projected-turnout ingestion regression that would otherwise silently null the win number for the whole endpoint. Verified locally: `dbt build --select` -> `PASS=1`.

## Companion

`thegoodparty/election-api#184` fixes the matching `Race.winNumber` field comment in the Prisma schema.

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics_candidacy_ballotready.sql
+++ b/dbt/project/models/intermediate/civics/int__civics_candidacy_ballotready.sql
@@ -275,8 +275,13 @@ with
             br_candidacy_id,
             br_race_id,
 
-            -- Assessment fields (hardcoded until we join viability/p2v in followup
-            -- work)
+            -- Assessment fields. BallotReady provides none of these: it does not
+            -- compute a viability_score, and supplies no win number at all.
+            -- Downstream, the candidacy mart fills viability_score from
+            -- TechSpeed's model; win_number / win_number_model have no live
+            -- source (only the 2023-2025 HubSpot archive in
+            -- int__civics_candidacy_2025 carries win_number), so they stay null
+            -- on the BR path.
             cast(null as float) as viability_score,
             cast(null as int) as win_number,
             cast(null as string) as win_number_model,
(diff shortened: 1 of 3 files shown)


hhkarimi:
Good catch, fixed in a89702e. Confirmed against the data: the `unmatched_br_positions` branch (`district_id = null`) accounted for 10,701 of the 20,678 "missing" rows, so the metric was conflating position-match coverage with turnout-ingestion coverage. Scoped the `races` CTE to `district_id is not null`, which drops the baseline from ~7.7% to ~3.9% and makes the thresholds guard turnout ingestion specifically. Comment updated to match.

delegate review

PR #442, opened 2026-06-04
Fail fast on non-retryable Databricks OAuth errors

hhkarimi:
## What

`get_databricks_connection` (used by every Databricks read in the Airflow DAGs, e.g. `sync_election_api`) retried **every** exception up to 20 times with a 30s delay and a "warehouse may be starting" message.

When a service-principal OAuth secret is rotated or expires, the connection fails with `invalid_client: Client authentication failed` — which Databricks itself tags `"no-retry-reason": "non-retryable error"`. Retrying can't fix it, so each task burned ~10 minutes looping, and the real cause was buried behind 20 misleading cold-start warnings.

This is what made `sync_election_api` (both `zip_to_position.load_staging` and `district_top_issues.load_staging`) hard to diagnose — it had been failing this way since mid-May.

## Change

- Classify non-retryable OAuth credential failures (`invalid_client`, `client authentication failed`) and raise immediately with an actionable message: rotate the service principal's OAuth secret in Databricks and update the Airflow connection.
- Transient / cold-start errors retry exactly as before.
- Adds `tests/test_databricks_utils.py` covering the classifier and the fast-fail / retry-then-succeed / exhaust-retries paths.

## Testing

`pytest tests/test_databricks_utils.py` — 9 passed. Full pre-commit suite (black, isort, flake8, mypy, pytest) green.

## Context

Surfaced while operating DATA-1958. Separate from that data change; no ticket yet — happy to link one if preferred.


hhkarimi:
delegate review

Fixed: moved the fail-fast auth tests out of `airflow/astro/tests/` (not collected by CI's `pytest tests`) into the existing CI-collected `tests/airflow/test_databricks_utils.py`, alongside the other databricks_utils coverage. Dropped the inline `sys.path.insert` and module-stub block since `tests/airflow/conftest.py` already puts `airflow/astro` on the path and the root tree imports the real packages. 25 tests pass (16 existing + 9 new).

PR #443, opened 2026-06-05
[DATA-1950] Fix G5420 school-district county/demographics

sanjayr1:
## What
Fix wrong county + demographics on school-district pages (mtfcc `G5420`) in `int__place_fast_facts`. School districts resolved their county via the non-G4110 ILIKE name-match, which hit boilerplate words in the district-type suffix: "X Union Free School District" matched the town of Union, Broome County NY (93 NY districts pinned to Broome); "X Community Unit School District" matched Unity, Alexander IL (54 IL districts). Roughly a third of the county-assigned user-facing G5420 had a wrong county AND wrong demographics on SEO-indexed pages.

## How
Mirror the merged G4110 fix: add a dedicated `G5420` branch and carve G5420 out of the non-G4110 filter, so every other branch (including G4110) stays byte-identical. The G5420 branch strips the district-type suffix to a candidate place name, then resolves: county-name match (or NULL) for "X County ..." names; else city primary-strip; else city retry-strip; else NULL. Reuses `deduped_cities_by_name` (city) and a new `deduped_counties_by_name` (county). Demographics: city match takes the matched city's own; county match takes the county's most-populous-city; no match is NULL.

## Impact counts (clarified)
- **11,042** G5420 at the intermediate grain (`int__place_fast_facts`); **~10,194** user-facing in the mart (`m_election_api__place`).
- Resolution at the intermediate grain: **71.2% resolved** (53.7% city-primary + 7.2% city-retry + 10.2% county), **28.8% honest NULL**.
- **5,656** G5420 rows change `county_fips` vs current prod; ~5,825 differ on any field. The unchanged ~5,200 already resolved correctly via the old path or were NULL in both.

## Coverage + safety (measured on a same-snapshot dev build; CI is the authoritative build/test)
- Non-G5420 same-snapshot diff = **0/0** (G4110 and every other branch unchanged).
- `equal_rowcount` preserved (73,470 == 73,470, no fan-out; the county join is 1:1).
- Zero new sinks (max resolved (state,county,population) cluster 7; the old bug pinned 50-93).
- Examples: Amagansett UFSD to Suffolk (own pop 663, was Broome/56,047), Mount Union Area to Huntingdon, Brown County CUSD to Brown, Dodge City USD to Ford, San Diego City Unified to San Diego.

## Independent (Codex) PR review — addressed
A code-only review by a different model flagged that ambiguous county names (12 keys where a county shares its name with an independent city, e.g. Baltimore, St. Louis MO, Richmond/Roanoke/Franklin VA) could pick the independent **city** over the **county**. Fixed: `deduped_counties_by_name` now orders by `county_fips asc` (US FIPS numbers independent cities 510+, so the lower FIPS is the actual county) and drops null-population cross-state stray rows. Verified: Baltimore County PS -> Baltimore County (24005, was city 24510), Richmond County VA -> 51159, Roanoke -> 51161, Fairfax stays 51059. Added Baltimore County and Richmond County VA as known-value fixtures.

## Tests
5 singular tests: known-values (geo_id-keyed, 15 fixtures incl. 2 NULL anchors + 2 ambiguous-county, red to green), null-ceiling (<= 3,600), classification-stable (anti-join), no-new-sink (>= 12), strip-mangle (pins the suffix-strip contract).

## Out of this PR / pre-merge
Spec, plan, measurement, and the diagnostic query are in `.tickets/DATA-1950/`. **Pre-merge gate:** a stratified correctness audit against an independent Census/NCES source (30 city-primary + 30 county + 40 city-retry). Propagation to election-api Postgres needs a full-refresh of the place chain + the `write__election_api_db` writer, owned by a teammate with prod write perms (per the runbook), with a direct-Postgres rollback. Follow-up: v2 GEOID-to-census/NCES SD-county crosswalk to close the ~29% NULL.

## Post-deploy (required)
After merge, run a one-time `dbt run --full-refresh --select int__place_fast_facts` (then the rest of the place chain) so the ~10k historical G5420 rows recompute through the new branch; the incremental `merge` will not backfill rows BallotReady has not re-synced. The prod build handles this. Verify live afterward: `ny/amagansett-union-free-school-district` should return Suffolk / 663.

### DIFF
--- a/dbt/project/models/intermediate/int__place_fast_facts.sql
+++ b/dbt/project/models/intermediate/int__place_fast_facts.sql
@@ -62,6 +62,29 @@ with
         where rn = 1
     ),
 
+    -- One row per (state, normalized county name), most-populous — for the G5420
+    -- county match.
+    -- Distinct from deduped_cities (1 row per county_fips): a few states have a
+    -- county AND an
+    -- independent city sharing a name (e.g. VA Richmond city vs Richmond County), so
+    -- (state, county_name) is NOT unique in deduped_cities. Deduping by (state,
+    -- county_name)
+    -- here keeps the county join 1:1 (protects equal_rowcount).
+    deduped_counties_by_name as (
+        select * except (rn)
+        from
+            (
+                select
+                    *,
+                    row_number() over (
+                        partition by state_id, lower(county_name)
+                        order by population desc nulls last, county_fips asc
+                    ) as rn
+                from {{ ref("stg_airbyte_source__ballotready_s3_uscities_v1_77") }}
+            ) ranked
+        where rn = 1
+    ),
+
     -- mtfcc is read only to split the branches below; it is not in the final output
     places as (
         select database_id, geo_id, name, slug, state, updated_at, mtfcc
@@ -75,7 +98,9 @@ with
     -- added) so the same-snapshot non-G4110 diff stays 0. A null mtfcc routes here too
     -- (the legacy path); assert_place_mtfcc_present fails loudly if one appears, since
     -- mtfcc is the only branch signal and is non-null for all places today.
-    non_g4110_places as (select * from places where mtfcc <> 'G4110' or mtfcc is null),
+    non_g4110_places as (
+        select * from places where (mtfcc not in ('G4110', 'G5420')) or mtfcc is null
+    ),
     joined_by_geo_id as (
         select
             p.database_id,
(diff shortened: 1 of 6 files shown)


sanjayr1:
delegate review

sanjayr1:
Thanks for the second pass. Addressed in `d932b6d`:

**Fixed — county null-population finding.** `deduped_counties_by_name` now drops cross-state stray rows by the real symptom (the row's `county_fips` 2-digit prefix not matching the state's own prefix) rather than `where population is not null`. That keeps legitimate population-less counties the proxy was silently NULLing — GA Echols (13101), GA Webster/Quitman, VA James City/Greensville, the VI districts. The prefix filter runs *before* `row_number()`, so a stray can't win the dedup. Note the suggested "drop the `where` and rely on `county_fips asc`" would mis-resolve TN Monroe (its KY stray `21171` sorts below the correct TN `47123`); the prefix filter avoids that. Added Echols -> 13101 as a known-value fixture.

**Declining — pass-3 `[0-9]+` removal.** Keeping it; it is required for the Missouri "Town NN" district-numbering convention, where it correctly strips the trailing number so the town name matches: Columbia 93 -> Columbia (Boone), Independence 30 -> Independence (Jackson), Kansas City 33 -> Kansas City, Perry County 32 -> Perry County, plus 18 more. Removing it NULLs all 22. "Region 10 School District" is NULL with or without the strip ("region" is not a city), so the bare `[0-9]+` never over-fires into a *wrong* match here. Added Region 10 -> "region" and Columbia 93 -> "columbia" strip-mangle fixtures.

**Full-refresh.** Agreed; added to the PR description. After merge, `int__place_fast_facts` needs a one-time `--full-refresh` so the ~10k historical G5420 rows recompute (the incremental merge won't backfill them); the prod build handles this.

All five tests green; the non-G5420 same-snapshot diff is still 0/0 (the change is isolated to the G5420 county relation and strip).

sanjayr1:
delegate review

sanjayr1:
Addressed in `b7ea637`. Confirmed the edge case on data: **VI** is the one state where every uscities row (194) has null population, so `max(case when population is not null ...)` returned NULL and the `WHERE` dropped the entire state.

Switched to a `state_fips_prefixes` CTE that takes the **most-common prefix per state (mode by row count)**, which `deduped_counties_by_name` joins to. Mode is robust where the simpler aggregates aren't: the suggested `min(prefix)` would pick TN's Kentucky stray (`21` < `47`) and mis-county every Tennessee "X County" district; `max`-over-populated nulls VI. Verified mode gives TN->47, GA->13, VI->78 (each the dominant prefix; strays are 1-3 rows). Coverage unchanged (7,859 resolved), known-values + fan-out green.

sanjayr1:
delegate review

sanjayr1:
delegate review

PR #475, opened 2026-06-09
[DATA-1972] office_type quick win: prefer BR in mart coalesce

sanjayr1:
First of two PRs for DATA-1972. Reorders the candidacy mart office_type coalesce to prefer BallotReady over gp_api, so matched product candidacies stop showing as 'Other'. Fixes a stale comment in office_standardizations.sql. PR2 (#478) adds the position-keyed crosswalk. See DATA-1972.

Measured impact in prod (Delta time travel, before vs after the on-merge build):

- Target population, the 576 product candidacies Splink-merged with a BallotReady row: 'Other' 568 -> 3. Redistribution: County Supervisor 0 -> 217, City Council 0 -> 214, School Board 0 -> 82, Mayor 0 -> 35, Judge 8 -> 9, Sheriff 0 -> 7, Clerk/Treasurer 0 -> 5, Town Council 0 -> 4.
- Correctness: every changed value equals the BallotReady classification derived from the official position name (0 mismatches against the BR intermediate; e.g. City Legislature -> City Council, Local School Board -> School Board, City Executive//Mayor -> Mayor).
- Safety: 0 rows lost a clean value; NULL count and total row count flat; ICP flags identical row-for-row.
- Downstream: users_win_candidacy 'Other' 23,569 -> 22,477 (-1,092 rows at campaign-version grain); leads_win_candidacy 'Other' 37,205 -> 36,642 (-563); m_election_api__race flipped 51 races from 'Other' to specific office types (26 County Supervisor, 13 City Council, 10 Clerk/Treasurer, 1 School Board, 1 Mayor).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #478, opened 2026-06-10
[DATA-1972] Position-keyed office_type crosswalk

sanjayr1:
Second of two PRs for DATA-1972. Adds int__civics_position_office_type (one office_type per BR position, derived from the normalized position name via the existing two-stage macro). The candidacy AND election marts inherit office_type from it by br_position_database_id via a hybrid coalesce, falling back to the per-source value when there is no position or the crosswalk can only say 'Other'. See DATA-1972.

Verified against the dbt Cloud CI preview schema (full plan battery, re-checked after the review-driven fallback-flag commit):

Crosswalk (int__civics_position_office_type)
- 285,394 rows, exactly one per BallotReady position; 0 NULL office_type; 0 fallback rows (every position currently resolves a normalized name).
- Reproduces the BallotReady candidacy derivation on 31,087 of 31,087 shared positions (0 disagreements), so BR-sourced rows are unchanged by construction.

candidacy mart, preview vs prod baseline
- gp_api candidacies 'Other': 93.2% before -> 8.2% after. The remaining 908 are all at positions whose normalized name genuinely maps to Other.
- TechSpeed-with-position NULL office_type: 13,493 before -> 0 after.
- Positions carrying 2+ distinct office_types across their candidacies: 3,455 before -> 1 after (the residual is a crosswalk-Other position where per-source values survive by design).
- Marquee races (US Senate TX/NC/IL; Governor CA/FL/GA/TX/CO): 502 candidacies, every one now Congressional or Statewide/Governor, zero 'Other' remaining.
- Safety: 0 rows that had a clean office_type became 'Other'; row count exactly equals prod; is_win_icp / is_serve_icp identical row-for-row; 48 positioned rows reference positions absent from the API position table and keep per-source values (expected, plan anticipated ~51).

election mart
- Wiring verified active (all 396,244 positioned rows join the crosswalk) with zero rows changed: the mart was already fully consistent with the crosswalk because its positioned rows derive from the same BR position path. The join is a consistency guard so candidacy and election stay aligned for the election-api surfaces if non-BR election sources appear.

Context: PR1 (#475), already in prod, fixed the 576 merged product candidacies (568 'Other' -> 3) and its downstream surfaces; this PR fixes the gp_api-only and TechSpeed populations and makes office_type position-stable.

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -2486,6 +2486,41 @@ models:
         data_tests:
           - not_null
 
+  - name: int__civics_position_office_type
+    description: >
+      One row per BallotReady position with the canonical office_type, derived
+      from the normalized position name (DATA-1972). Source of office_type for
+      positioned candidacies in the civics marts.
+    columns:
+      - name: br_position_database_id
+        description: BallotReady position database id (grain).
+        data_tests:
+          - not_null
+          - unique
+      - name: office_type
+        description: Canonical office_type for the position (map_office_type output).
+        data_tests:
+          - not_null
+          - accepted_values:
+              arguments:
+                values:
+                  - Alderman
+                  - Attorney
+                  - City Council
+                  - Clerk/Treasurer
+                  - Congressional
+                  - County Supervisor
+                  - Judge
+                  - Mayor
+                  - President
+                  - School Board
+                  - Sheriff
+                  - State House
+                  - State Senate
+                  - Statewide/Governor
+                  - Town Council
+                  - Other
+
   - name: int__civics_candidate_gp_api
     description: >
       Product Database (gp_api_db) users transformed into the civics mart
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -2520,6 +2520,23 @@ models:
                   - Statewide/Governor
                   - Town Council
                   - Other
+      - name: is_office_type_fallback
+        description: >
+          True when no candidate_office could be derived for the position (no
+          normalized position name available and no position-name pattern
+          matched), making office_type the NULL-input fallback 'Other' rather
+          than a real classification. True rows mean the incremental
+          normalized-position model is lagging the position universe; the mart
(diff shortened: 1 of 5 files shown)


sanjayr1:
Re the reviewer finding on normalized-position join misses: the mechanism is real and is now exposed. Measured against current prod, the miss rate is zero (every position in the staging universe has a normalized_position struct that matches a row in the normalized-position model), so all crosswalk 'Other' values today are genuine classifications, not NULL fallthrough. Since the incremental Python model could lag future positions, the crosswalk now emits is_office_type_fallback (candidate_office is null) with a warn-severity all-false test, so any future lag surfaces in CI without blocking builds. office_type semantics are unchanged, and the mart coalesce already prevented fallback 'Other' from overriding per-source values.

delegate review

sanjayr1:
Triage of the three new findings, against the reviewed implementation plan for DATA-1972:

1. elected_official_terms / elected_officials wiring: this is a recorded scope decision in the reviewed plan, not an oversight. The EO marts derive office_type from their own BallotReady officeholder lineage (position_name / normalized_position_name through the same macros), so wherever the crosswalk classifies a position the EO derivation already agrees with it by construction (the crosswalk reproduced the BR derivation exactly on every shared position in verification). This PR does not introduce new divergence; wiring the EO marts to the crosswalk is documented in the plan as an optional follow-up, and will be noted on DATA-1972.

2 and 3. accepted_values on candidacy.office_type / election.office_type: as proposed this does not fit the marts' data contract. The 2025 archive rows are frozen by design and carry historical free-text office_type values (the election mart in particular has many distinct historical values, including casing and spelling variants). An enum test at the mart boundary would warn permanently on legitimately frozen data, which is noise rather than observability. The new derivation path is already guarded where the guard is meaningful: the crosswalk enforces a strict enum plus a not_null and unique key on its own output, the warn-severity is_office_type_fallback test surfaces NULL-input fallthrough, and the mart coalesce cannot emit a value that neither the crosswalk nor a source produced. If a mart-boundary guard is wanted, it would need a config.where excluding archive rows, which is best decided by the human reviewer as a follow-up rather than bolted on here.

sanjayr1:
delegate review

PR #480, opened 2026-06-10
analytics: pipeline cost profiler + per-run decision flags + baseline benchmark

datawithtristan:
## Summary

Adds a deterministic **pipeline cost profiler** for the Win-analytics pipeline: it reads a Claude Code session transcript and attributes per-stage (framing / execution / review / calibration) **token counts and wall-clock time, split into model-active vs human-idle**, so we can see where a run's time and tokens actually go. Plus the decision flags that explain *why* a run cost what it did, a reusable baseline benchmark, and an opt-in invocation step in the runbook.

Stdlib-only, no Databricks dependency. Built test-first (73 tests).

## What's included

- **`analytics/diagnostics/pipeline_profiler.py`** — marker-based stage segmentation (skill load → `_brief.yaml` write → first reviewer dispatch → `CALIBRATION_*.md`), per-stage token totals and model-active/human-idle time split, parallel-reviewer max-span, a `confidence` flag that drops to `low` on missing/out-of-order markers, and a Markdown report.
- **Per-run decision flags** — deliverable type (`.ipynb` / analysis `.py` / `build_*_nb.py`), reviewer dispatch counts, and process-design edits (process-owning files touched at/after calibration). Rendered as a separated `**Decisions:**` block so choices sit next to the time they produced.
- **Report columns** — `model_active_min`, `human_idle_min`, `wall_clock_min`, `%_of_active`, token columns, and `%_of_tokens`, with a per-run total row and a combined mean summary. Reviewer-internal tokens are labeled as a known blind spot (sub-agent sidechains aren't in the transcript).
- **Per-run logging** — each run appends its full report to a gitignored `analytics/diagnostics/logs/profile_log.md` for later review.
- **`analytics/diagnostics/baseline_question.md`** — a fixed, reusable benchmark question + the runnable arm protocol (framing depth, deliverable, review, calibration) for comparing how pipeline decisions affect runtime.
- **`analytics/runbook/run-product-analysis.md`** — an opt-in "profile this run?" question at the start (default **no**), executed automatically at the closing step if opted in; also refreshes the stale Databricks prerequisite to the profile-auth setup (post Spec A / #474).

## Stage definitions

The four profiler stages map 1:1 to the pipeline stages in the `win-analytics-process` skill's `pipeline.md`. The two human-approval gates fold into the preceding stage (framing-approval → framing; results checkpoint → execution); the module docstring documents this.

## Verification

- 73 tests pass (`analytics` suite); ruff check + format clean; mypy clean (pre-commit).
[...]

PR #482, opened 2026-06-10
fix(analytics): unblock repo-wide pre-commit (RUF001/SIM108 in pipeline_profiler)

hhkarimi:
## Summary

#480 (`analytics: pipeline cost profiler`) landed with lint violations that fail CI's repo-wide `pre-commit run --all-files`, so **every open PR is currently red** (the failure shows up on unrelated PRs because pre-commit lints the whole tree on the merge ref). This clears them:

- 6× **RUF001** — ambiguous `×` (MULTIPLICATION SIGN) in the diagnostic report strings (`notebook ×3`, etc.).
- 1× **SIM108** — an `if/else` ruff wants as a ternary.

## Approach

Targeted `# noqa` rather than changing behavior or output:
- `# noqa: RUF001` keeps the intentional `×N` ("N times") glyph in the report — readable and deliberate; `× → x` would give `alicex2`.
- `# noqa: SIM108` keeps the `calibration` block as an `if/else`, parallel to the structurally identical `review` block right above it (which ruff doesn't flag only because its ternary would be too long).

No logic or output changes — comment-only.

## Test Plan
- [x] `pre-commit run ruff --files <both files>` → Passed
- [x] `pre-commit run ruff-format --files <both files>` → Passed

cc @ the #480 author — happy to switch to `× → x` instead of noqa if you'd prefer the ASCII glyph.

### DIFF
--- a/analytics/diagnostics/pipeline_profiler.py
+++ b/analytics/diagnostics/pipeline_profiler.py
@@ -210,8 +210,8 @@ def segment_stages(records: list[dict]) -> tuple[dict, str]:
         review = (reviewer, max(reviewer, rev_end))
     else:
         review = (exec_end, exec_end)  # empty
-    # calibration
-    if calib is not None:
+    # calibration  (kept as an if/else to parallel the `review` block above)
+    if calib is not None:  # noqa: SIM108
         calibration = (calib, _first_human_after(records, calib))
     else:
         calibration = (n, n)  # empty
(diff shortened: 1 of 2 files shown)


hhkarimi:
Folding this fix into #477 instead (the loader PR needs it to be green anyway, and one PR is simpler). The same commit (RUF001/SIM108 noqa in pipeline_profiler) is now on the #477 branch. Closing.

PR #468, opened 2026-06-09
uv-standardize airflow + bump people-api-loader to Python 3.14

hhkarimi:
Final dir of the uv-standardization spec (airflow), plus a small follow-on bump of people-api-loader so every directory lands on Python 3.14.

## airflow (the main change)
airflow has two dependency worlds, so this is deliberately narrower than the other conversions.
- **Local-dev / CI env** converted poetry -> uv on **Python 3.14** (`pyproject.toml` rewritten PEP-621, `poetry.lock` -> `uv.lock`, `.python-version` 3.14). Uniform `setup-uv` + `uv sync`/`uv run` workflow, fuller ruff, per-dir mypy, pre-push hook on `uv run pytest`.
- **Deploy stays on `astro/requirements.txt`** — Astro has no uv-native project model; it already resolves requirements.txt *with* uv internally. The only deploy change is the Dockerfile tag: `runtime:3.2-2` -> `runtime:3.2-2-python-3.14`. `astro/requirements.txt` + `astro/packages.txt` untouched.
- Verified: 50 passed / 2 skipped on 3.14.5; ruff/format/mypy clean.
- **Deploy caveat:** 3.14 is ahead of Apache Airflow's official matrix (only Astronomer's image extends to it). CI doesn't run an Astro deploy, so verify the next deploy out-of-band; one-line fallback to `-python-3.13` if needed.

## people-api-loader (the follow-on bump)
Already on the Astral toolchain (uv + ty), so this is just a version bump: `.python-version`, `requires-python`, the `[tool.ty.environment]` version, and the ruff target. Verified on 3.14.5: **ty clean**, ruff/format clean, 15 tests pass. No type-checker change.

With this, every directory in the repo is on Python 3.14 (people-api-loader, matcha, genie-tools, genie-slack-bot, dbt, analytics, airflow).

Note: lightly conflicts with the other in-flight uv PRs (#464, #466, #467) on `CLAUDE.md` + the central mypy `exclude` — trivial to reconcile.

---

> [!NOTE]
> **Medium Risk**
> Deploy moves to a Python 3.14 Astro runtime ahead of upstream Airflow’s official matrix, and ingestion DAGs depend on stricter Databricks connection checks that could surface misconfigured connections in prod.
> 
> **Overview**
[...]

PR #464, opened 2026-06-08
uv-standardize the genie apps (genie-tools + genie-slack-bot, Python 3.14)

hhkarimi:
uv-standardization spec, applied to **both** apps under `apps/` in one PR (they're the genie apps).

## genie-tools
- uv project on **3.14**, uniform `setup-uv` + `uv sync`/`uv run` workflow, fuller ruff set, per-dir mypy, pre-push hook on `uv run pytest`. CLAUDE.md updated.
- Added 4 tests for `_parse_serialized_space` (the prior review's finding).

## genie-slack-bot
- **Drops `requirements.txt`** -> `pyproject.toml` + `uv.lock` (3.14). This switches the **Databricks App** deploy to uv-native (the platform installs via uv when no `requirements.txt` is present). `app.yaml` is unchanged — it still runs `python app.py`.
- Uniform uv workflow, fuller ruff set (1 `SIM103` fixed by hand, rest auto), per-dir mypy, pre-push hook on `uv run pytest`. CLAUDE.md updated.
- Verified: 22 tests pass on 3.14.5; ruff/format clean; mypy clean.

### Deploy verification needed (out of CI scope)
Dropping `requirements.txt` changes how the genie-slack-bot Databricks App installs deps (pip -> uv). CI exercises the test suite but not the Databricks runtime install path, so the **next app deploy should be verified out-of-band** (staging or next deploy cycle) to confirm the bot still starts.

Note: lightly conflicts with the dbt (#466) / analytics (#467) PRs on `CLAUDE.md` + the central mypy `exclude` — trivial to reconcile (each adds its own dir).

PR #484, opened 2026-06-11
ci: reusable uv-subproject workflow + best-practice fixes

danpelota:
Follow-up to the uv-standardization PRs (#464, #466, #467, #468), which left six per-subproject CI workflows with byte-identical job bodies. Dedups them and applies a few CI best practices.

- Adds `.github/workflows/uv-subproject.yml` (`workflow_call`) with the shared job: checkout, `setup-uv`, `uv sync --frozen`, pytest, optional mypy. Inputs: `directory`, `mypy-targets`, `pytest-args`.
- `airflow`, `analytics`, `dbt`, `genie-slack-bot`, `genie-tools`, `matcha` collapse to triggers plus a one-line call. Triggers/path filters stay per-caller (the part that differs); each caller adds `uv-subproject.yml` to its `paths`.
- `permissions: contents: read` and a `concurrency` group (cancels superseded PR runs, never main) folded into the reusable workflow; all six callers inherit them.
- `pre-commit.yml`: runner Python 3.11 -> 3.14 (the old `must match prod airflow` pin was stale; prod airflow is on 3.14). Same permissions/concurrency added.
- `people-api-loader.yml`: `setup-uv` v3 -> v6, drop `ruff check`/`format` steps already covered repo-wide, sync `--frozen --extra dev`, add permissions/concurrency. Stays standalone (uses `ty`, not mypy).

`pre-commit` was already red on main from ruff debt in the analytics profiler (#480). Cleared it: ignore `RUF001` in analytics (matching dbt) for the `×` glyphs, and collapse one `SIM108` if/else to a ternary.

### DIFF
--- a/.github/workflows/people-api-loader.yml
+++ b/.github/workflows/people-api-loader.yml
@@ -1,5 +1,8 @@
 name: people-api-loader
 
+# Standalone (not via uv-subproject.yml): this project uses the Astral type
+# checker `ty` instead of mypy, and keeps its dev deps in an optional-deps
+# extra, so it needs `--extra dev`. Ruff is covered repo-wide by pre-commit.yml.
 on:
   pull_request:
     paths:
@@ -11,6 +14,13 @@ on:
       - 'people-api-loader/**'
       - '.github/workflows/people-api-loader.yml'
 
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


.github/workflows/people-api-loader.yml
```diff
@@ -1,5 +1,8 @@
 name: people-api-loader
 
+# Standalone (not via uv-subproject.yml): this project uses the Astral type
+# checker `ty` instead of mypy, and keeps its dev deps in an optional-deps
```
hhkarimi, line 4: We'll eventually get `ty` throughout, at the moment I had some issues with the airflow project where `mypy` works better.

hhkarimi (approved):

PR #489, opened 2026-06-11
[DATA-1582] Enrich election_stage mart for DDHQ coverage-gap analysis

hhkarimi:
## Summary

Enriches the `election_stage` civics mart with `office_level`, `office_type`, and `br_candidate_count` so DDHQ-vs-BR election coverage-gap analysis can be expressed as a **filter over the mart** — the mart's `source_systems` already reflects DDHQ coverage — rather than a standalone crosswalk-joining model.

This **supersedes #446** (the standalone `int__civics_ddhq_election_stage_coverage_gaps` model), per review feedback that the mart + a filter is the cleaner home.

- `office_level` / `office_type`: BR-first coalesce from the per-source election_stage models.
- `br_candidate_count`: distinct BR candidacies per race from the race-grain S3 candidacies source (avoids the `any_value(br_race_id)` collapse in the rolled-up candidacy model).
- All additive; grain unchanged (one row per `gp_election_stage_id`).

## Why the mart is the right (and more accurate) source

The mart's `source_systems` is built from the per-source `gp_election_stage_id` alignment, which turns out to be **more complete than the Splink ER crosswalk for DDHQ**: it identifies **260 real DDHQ matches** (all with a `ddhq_race_id`) within the actionable subset that the crosswalk missed (the crosswalk caught 64 the mart missed). So the gap count drops and gets more trustworthy.

## The coverage-gap query

DDHQ-vs-BR local coverage gaps, bounded to DDHQ's data horizon and scoped to the agreement (partisan primaries excluded):

```sql
with ddhq_horizon as (
    select max(election_date) as last_ddhq_election_date
    from mart_civics.election_stage  -- or the int model: int__civics_election_stage_ddhq
    where array_contains(source_systems, 'ddhq')
)
select
    e.gp_election_stage_id,
    upper(substring(e.race_name, 1, 2)) as state,  -- 2-letter race_name prefix
    e.race_name,
    e.office_level,
    e.election_date,
    e.stage_type,
    e.br_candidate_count,
    array_contains(e.source_systems, 'techspeed') as has_techspeed_match
from mart_civics.election_stage e, ddhq_horizon h
where not array_contains(e.source_systems, 'ddhq')      -- DDHQ has no match
  and array_contains(e.source_systems, 'ballotready')   -- BR has the race
  and e.election_date between date '2026-01-01' and h.last_ddhq_election_date
  and e.office_level not in ('State', 'Federal')         -- local
  and not coalesce(e.partisan_type = 'partisan' and e.is_primary, false)  -- agreement scope
  and e.br_candidate_count > 0                            -- real races only
order by state, e.election_date, e.race_name;
```

## Funnel (current data)

- **13,433** scoped local gaps (within DDHQ window, agreement-scoped)
- → **813** with BR candidates (real races)
- → **364** also TS-confirmed (a second source confirms the race exists; highest-confidence "DDHQ missed a real race")

### DIFF
--- a/dbt/project/analyses/ddhq_election_stage_coverage_gaps.sql
+++ b/dbt/project/analyses/ddhq_election_stage_coverage_gaps.sql
@@ -1,0 +1,44 @@
+-- DDHQ vs BallotReady local election coverage gaps.
+--
+-- This is a dbt *analysis*: it is compiled (refs resolved, validated against the
+-- schema) but never materialized. Run the compiled SQL ad hoc to inspect which
+-- BallotReady local races DDHQ has not matched.
+--
+-- Definition of a gap: BallotReady has the race (`ballotready` in source_systems)
+-- but DDHQ does not (`ddhq` not in source_systems). Scoped to be a fair,
+-- actionable comparison:
+-- - bounded to DDHQ's data horizon (max DDHQ election_date) so far-future BR
+-- races DDHQ has not loaded yet are not counted as gaps;
+-- - local only (office_level not in State/Federal);
+-- - partisan primaries excluded (out of scope for the DDHQ coverage agreement:
+-- independent / 3rd-party candidates are ineligible);
+-- - real races only (br_candidate_count > 0) -- races with no filed candidates
+-- are largely uncontested or BR implied-future races that never reach DDHQ.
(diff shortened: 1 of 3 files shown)


hhkarimi:
delegate review

PR #491, opened 2026-06-12
[DATA-86ahzvddm] Add 
  elected_official_support civics mart

melecia-gp:
## What

Adds a new civics mart table, `elected_official_support`, with one row per elected official per office, describing the vote support behind that office's most recent general-election win. It carries the votes the official received, the total votes cast in that race, the number of seats in that race, the count of registered voters in the district from the L2 voter file, and a projected count of registered voters who support them.

## Why

The Serve product wants to show each elected official a support figure: how many people in their jurisdiction support them, against the total population of that jurisdiction. We do not yet have a constituents dataset, so as an interim placeholder this table projects the official's general-election vote share onto the count of registered voters in their district. When a real constituents count arrives, the same vote share will be applied to it instead. Closes [#86ahzvddm](https://app.clickup.com/t/86ahzvddm).

## How

- **Grain and scope.** One row per (`gp_elected_official_id`, `br_position_id`), taken from that office's most recent general-election win. Primaries are excluded because winning a primary does not put someone in office. "General" includes general, general special, general runoff, and general special runoff stages. An official who holds more than one distinct office gets one row per office; re-elections to the same office collapse to the most recent win.

- **Every column populated.** The table keeps only contested wins that have a coherent vote tally and an L2 registered-voter count. This is a deliberate decision: the support figure is the point of the table, and a row that cannot produce one is not shown. The result is a small, fully-populated table rather than a sparse roster.

- **Join path.** `elected_official_terms` to `candidacy_stage` on `br_candidacy_id` (cast because it is an integer on one side and a string on the other), then to `election_stage` on `gp_election_stage_id` for the total, and a left join to the `positions` mart on `br_position_database_id` for the L2 voter count (`icp_voter_count`). We source from `positions` rather than the `int__icp_offices` intermediate so the L2 count uses the same mart and column name (`icp_voter_count`) that `positions` and `election` already use.

- **Sources.** `votes_received` is DDHQ-canonical via `candidacy_stage`. `total_votes_cast` is BallotReady-first (DDHQ fallback) via `election_stage`. We confirmed that where both vendors carry a total they are identical in 100 percent of overlapping cases, so the BallotReady-first total is used per repo standard at no accuracy cost.

- **Excluded rows.** Uncontested winners have no numeric total, so they are dropped (a support fraction is not meaningful without an opponent). Rows where votes exceed the reported total are genuine source errors under both vendors and are dropped whole. Rows with no L2 district match are dropped because no support figure can be projected.

- **Projected supporters.** `icp_voter_count * (votes_received / total_votes_cast)`, rounded to a whole number. This applies the official's general-election vote share to the full registered-voter base. It assumes non-voters would split the same way as voters. It is fine for a display estimate and is not for analysis. It is explicitly an interim placeholder for a future constituents-based figure.

- **Multi-seat races.** About 47 percent of rows (1,030 of 2,164) are multi-seat or at-large races. In those, the electorate is split across several winners, so a single winner's votes over the race total is a structurally low share and the support fraction is not comparable to a single-seat race. We surface `number_of_seats` so the consumer can label or adjust the figure rather than silently dropping these officials.

- **Low fractions from in-progress vote loads.** A few very recent winners have implausibly low recorded vote counts because the source vote data is still loading (for example a 3-seat school board where two winners show 5 and 6 votes against 1,017 cast). By decision these rows are kept, not filtered, and will correct themselves as the source data finishes loading. The model documents this rather than hiding it.

## What this affects

- Adds one new table to `mart_civics` (`elected_official_support`). No existing models or tables are changed.
- Updates `m_civics.yaml` with the model description, column docs, and tests.
- Reads from existing marts (`candidacy_stage`, `elected_official_terms`, `election_stage`, and `positions`). No upstream changes.

## Testing

Materialized to `goodparty_data_catalog.sandbox.elected_official_support` and validated against `mart_civics` in Databricks. Current counts:

- 2,164 rows across 2,160 officials (4 officials hold two offices and get two rows)
- Composite key (`gp_elected_official_id`, `br_position_id`) is unique with no collisions
- Every column is fully populated (zero nulls)
- Support fraction ranges from 0.005 to 1.0, median 0.5

Checks that passed:

- [x] Materialized to sandbox and row-level integrity validated
- [x] No null or duplicate composite key
- [x] No null in any column
- [x] No row where `votes_received > total_votes_cast`
- [x] No row where `projected_registered_supporters > icp_voter_count`
- [ ] `dbt build` (run + test) to be run in CI or by a teammate with dbt Cloud access; no local dbt access at drafting time

Note: the new `votes_received <= icp_voter_count` test is set to warn and currently flags 9 rows. These are real upstream L2-match errors (the matched district is too small), not a problem in this model. They are left visible on purpose.

## Tests added in m_civics.yaml

- `not_null` on every column
- `dbt_utils.unique_combination_of_columns` on (`gp_elected_official_id`, `br_position_id`)
- `dbt_utils.accepted_range` on the numeric columns (`number_of_seats` min 1, the rest min 0)
- `relationships` to `elected_officials` (warn)
- `votes_received <= total_votes_cast` and `projected_registered_supporters <= icp_voter_count` regression guards (warn)
- `votes_received <= icp_voter_count` data-quality check (warn), which surfaces too-small upstream L2 district matches

## Known limitation

Requiring an L2 voter count excludes roughly 270 officials who have a usable vote tally but no L2 match. The gap is not random: it is concentrated in school boards and special districts, and over-represents Texas and Missouri. The cause is upstream, in how the BallotReady position to L2 district match works (the same match `positions` and `int__icp_offices` rely on). For most of these the L2 data appears to exist under a differently-formatted district name. This is worth a separate ticket and does not block this table.


hhkarimi:
delegate review

PR #492, opened 2026-06-12
Fix hubspot_calls disposition handling and disable stale techspeed freshness

hhkarimi:
## Summary

Fixes the two failing jobs from the latest dbt Cloud run, and properly resolves the underlying disposition-seed gap rather than just silencing it.

### 1. `dbt build` test failure — `int__hubspot_calls` (FAIL 35)

The test `hs_call_disposition is null or disposition_label_source is not null` failed on 35 rows (4 distinct GUIDs) whose disposition GUID was not in the `hubspot_call_dispositions` seed, so the left join left `disposition_label_source` null.

Resolved in three layers:

**a. Map the missing GUIDs (the real fix).** Three of the four are recognizable dispositions; leaving them unmapped would have mis-bucketed 22 rows as `unmapped`, undercounting `connected_pledge` / `connected_other` downstream. Labels confirmed against the HubSpot "Edit call outcomes" list:

| GUID | label | outcome_family |
|---|---|---|
| `546b672d…` | Connected - Pledged + Offer Accepted | `connected_pledge` |
| `5d083c6b…` | Connected - No Pledge | `connected_other` |
| `6782cb6c…` | Connected - Pledge, No Offer Accepted | `connected_pledge` |
| `8a9d9833…` | — (manually-logged calls, no title/body signal) | `unknown` / `unmapped` |

`546b672d` also carried a "Verbal Pledge - EO Waitlist" title, but that label already maps to GUID `bbdca77e` in the seed, so by elimination `546b672d` is the "Offer Accepted" outcome. `8a9d9833` has no label signal and stays `unknown` pending the HubSpot API.

**b. Safety-net coalesce.** `int__hubspot_calls` now resolves any GUID still missing from the seed to `'unknown'` / `'unmapped'` instead of null, so a single new disposition can never hard-break the 2,392-node monthly build.

**c. Warn-severity detector.** A `relationships` test (`warn_unmapped_hs_call_disposition_guid`) on `hs_call_disposition` → `hubspot_call_dispositions.guid` surfaces future unseen GUIDs as a warning (not a silent `unmapped` bucket, not a build-breaking error). This was a sanctioned follow-up from the seed's original PR (#392).

Seed maintenance is documented in the seed schema description (authoritative GUID→label only via the HubSpot API; title-inference caveats; the warn-test workflow) and a pointer comment in `int__hubspot_calls.sql`.

### 2. `dbt source freshness` — `techspeed_gdrive_marketing_data_enrichment` (ERROR STALE)

Confirmed against Databricks: last loaded 2025-12-02 (192 days ago), past the 180-day `error_after`, with no downstream consumers. Disabled freshness on this source, mirroring the `gp_api_db_path_to_victory` precedent.

## Verification

- `dbt seed --select hubspot_call_dispositions --full-refresh` → loaded
- `dbt build --select int__hubspot_calls` → PASS=8 (incl. the new warn test)
- Post-fix: the 4 GUIDs classify correctly; unmapped rows dropped 35 → 13 (only the signal-less `8a9d9833`)
- `dbt source freshness --select source:airbyte_source.techspeed_gdrive_marketing_data_enrichment` → no longer evaluated

## Follow-ups (not in this PR)

- Bind `8a9d9833` authoritatively via the HubSpot API (`GET /crm/v3/properties/calls/hs_call_disposition`) and replace its `unknown` placeholder.
- `techspeed_gdrive_officeholders` is also drifting (101 days, same thresholds) and will hit the same STALE error in ~11 weeks if that feed is likewise retired — worth a data-team confirmation.

### DIFF
--- a/dbt/project/models/intermediate/hubspot/int__hubspot.yaml
+++ b/dbt/project/models/intermediate/hubspot/int__hubspot.yaml
@@ -14,6 +14,19 @@ models:
           - unique
       - name: hs_call_disposition
         description: "HubSpot internal GUID for the call outcome"
+        data_tests:
+          # Surfaces disposition GUIDs that have appeared in the calls data but
+          # are not yet mapped in the hubspot_call_dispositions seed. Warn-only:
+          # the model coalesces unmapped GUIDs to 'unknown'/'unmapped', so a
+          # missing seed row should not break the build — it just needs a
+          # follow-up seed entry (see the seed's maintenance notes).
+          - relationships:
+              name: warn_unmapped_hs_call_disposition_guid
+              arguments:
+                to: ref('hubspot_call_dispositions')
+                field: guid
+              config:
+                severity: warn
       - name: disposition_label
         description: >
           Human-readable outcome label resolved from
--- a/dbt/project/models/intermediate/hubspot/int__hubspot_calls.sql
+++ b/dbt/project/models/intermediate/hubspot/int__hubspot_calls.sql
@@ -20,6 +20,11 @@ with
         select
             c.*,
             s.disposition_label,
+            -- Safety net: a disposition GUID not yet in the hubspot_call_dispositions
+            -- seed resolves to 'unknown'/'unmapped' rather than null, so a missing
+            -- seed row never breaks the build. New GUIDs are surfaced by the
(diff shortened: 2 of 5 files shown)


hhkarimi:
delegate review

PR #495, opened 2026-06-12
Partition district_top_issues load by issue (sync OOM follow-up to #493)

hhkarimi:
## Why

#493 fixed the `sync_election_api` OOM for `zip_to_position` by partitioning its Databricks read. But the sync **still OOMs** (`SIGKILL`): the two `load_staging` tasks (`zip_to_position`, `district_top_issues`) run **concurrently on the same worker**, and `district_top_issues` still reads its ~5.1M-row mart unpartitioned. Its client-side buffer dominates worker memory, so the combined load dies — the logs show zip_to_position getting a few states in (AZ, SC, LA, ~50k rows) before the worker is killed.

## What

Partition `district_top_issues.load_staging` by `issue` (68 distinct values, ≤~96k rows each) using the `partition_column` mechanism already added in #493. Each read is now bounded to one issue's rows, so the two concurrent loads fit in worker memory.

No new util code — this is the one-line application of the existing, tested `partition_column` parameter. The generic partition behavior (chunking, NULL handling, quote escaping, rollback-on-failure, generator close) is already covered by `test_election_api_utils.py`; existing `test_sync_election_api` / DAG-integrity suites still pass.

### DIFF
--- a/airflow/astro/Dockerfile
+++ b/airflow/astro/Dockerfile
@@ -1,2 +1,1 @@
 FROM astrocrpublic.azurecr.io/runtime:3.2-5-python-3.14
-# touch: force an image deploy so the dev image picks up include/ (partition_column).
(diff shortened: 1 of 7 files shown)


hhkarimi:
delegate review

PR #513, opened 2026-06-18
[DATA-1993] People Served metrics + per-official constituents

sanjayr1:
## What

The first product-facing consumers of the district/census substrate (epic DATA-1359; this is the last v1 build):

- **`int__serve_block_coverage`** (intermediate/civics): the count-once engine. Scans the L2 voter file once to emit per-`(block, served_set)` distinct served-voter counts. Count-once is a distinct-person union across overlapping district types, which the counts-only substrate cannot produce (cross-type voter dedup needs voter IDs), so this is the "scanned, never surfaced" node from TDD 7.1. Counts-only output (PII-free). Scoped to the same 22 district types as the substrate, so the ordering invariant holds.
- **`people_served`** (mart/analytics): the People Served metrics, one row per `(metric_variant, office_type)`. count_once plus count_multiple per district / seat / org. Statewide officials read the exact T5 `State` rows and are reported separately (`office_type='State'`), never folded into the local `all` headline (TDD 4.5). Thin by design.
- **`official_constituents`** (mart/analytics): one row per resolved serve official with `constituents`, `registered_voters`, the gap, and `has_census_match`.

## Key decisions

- **count-once = voter-union** (the distinct-person North Star), not the whole-block reading. Whole-block breaks the ordering invariant: it sums a split block's full population while count-multiple is on the substrate's allocated basis. A spec inconsistency caught at plan time and resolved with two independent model reviews plus live data.
- **requires_review is carried through, never filtered.** Wrong-scope matches are fixed upstream (DATA-1989 overrides), not patched in the model (a blunt flag filter over-excludes legitimate districts). The flag rides on `official_constituents` for the product to caveat.
- **High-undercount flag deferred** to a follow-up (a correct voterless/invisible-population measure needs geographic assignment of zero-voter blocks, out of v1 scope); v1 ships the honest `constituents_minus_voters` gap.
- **Seat multiplier hardened** with an `organization_slug` fallback so `1 <= n_seats <= n_orgs` per district (guarantees per-district <= per-seat <= per-org).
- Codex plan-gate review: GO after revisions (statewide redesign, seat-id namespacing, unique test, reconciliation additions); 6 other findings verified as non-issues against data.

## Tests

- Ordering invariant (singular): count_once <= per_district <= per_seat <= per_org within every office_type.
- served_voters <= total_voters; unique keys on both marts; range checks; a wide North-Star sanity band. Bands, not fixed counts (the cohort is live and the L2 file refreshes).

## Verification

count-once and count-multiple were validated on prod data before commit (ordering holds for all 21 binding types; statewide join confirmed). Reconciliation on the CI preview schema will follow as a comment once the build completes. Local dbt is Fusion and cannot parse the project, so verification is on CI.

Not self-merged; opened for human review.

Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> New heavy L2 nationwide materialization and population metrics that drive product OKRs; correctness is guarded by ordering and key tests but wrong-scope matches still flow into headline numbers until upstream fixes.
> 
> **Overview**
> Introduces the first **Serve analytics consumers** of the district/census substrate: headline People Served metrics and per-official constituent counts.
> 
> **`int__serve_block_coverage`** (new intermediate table) is the **count-once engine**: one scan of `int__l2_nationwide_uniform`, unpivot across the same 22 L2 district types as the substrate, join to cohort districts from `int__serve_district_resolution`, and emit PII-free `(block_geoid, served_set)` rows with `served_voters` / `total_voters` for `'all'` and per-type sets. Statewide offices stay out of this model.
> 
> **`people_served`** (new mart table) exposes one row per `(metric_variant, office_type)`: **count_once** from block coverage × census block population; **count_multiple** variants from `district_census_stats` with seat/org multipliers (namespaced seat ids + `organization_slug` fallback). Local **`office_type = 'all'`** is separate from **`State`** statewide totals (TDD 4.5).
> 
> **`official_constituents`** (new view) is one row per resolved Serve org with `constituents`, `registered_voters`, `constituents_minus_voters`, and `has_census_match`, joining local districts on normalized names or statewide on `district_type = 'State'`. **`requires_review`** is passed through, not filtered.
> 
> **Tests/docs:** YAML for the new models; singular **`assert_people_served_ordering_invariant`**; uniqueness, ranges, and a warn-level North Star sanity band on `count_once` / `all`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7944a77ccebc63bb9719bd1244d5e61862d357a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -3413,3 +3413,57 @@ models:
             expression: >
               length(block_geoid) = 15 and length(block_group_geoid) = 12
               and length(tract_geoid) = 11
+
+  - name: int__serve_block_coverage
+    description: >
+      The count-once engine (DATA-1993, epic DATA-1359). One row per (census block,
+      served_set) with the count of DISTINCT cohort-served voters and the block's total
+      voters; people_served reads it as count-once(set) = sum(served_voters /
+      total_voters * block_population) over served blocks. Count-once is a
+      distinct-PERSON union across overlapping district types, which the counts-only
+      substrate cannot produce (cross-type voter dedup needs voter IDs), so this model
+      scans int__l2_nationwide_uniform once -- the "scanned, never surfaced" node from
+      TDD 7.1 -- and emits only block-grain counts (PII-free). Scoped to the same 22
+      district types as the substrate (get_l2_major_district_columns), so count-once <=
+      count-multiple holds. served_set is 'all' (served by any cohort district) or one
+      of the 22 L2 district types; statewide officials are read from the exact T5
+      'State' rows in people_served, not here.
+    columns:
+      - name: block_geoid
+        description: 15-char zero-padded 2020 census block GEOID
+        data_tests:
+          - not_null
+      - name: served_set
+        description: >
+          'all' (voters served by any cohort district) or an L2 district_type (voters
+          served by a cohort district of that type)
+        data_tests:
+          - not_null
+      - name: served_voters
+        description: Distinct cohort-served voters in this block for this served_set
+        data_tests:
+          - not_null
+          - dbt_utils.accepted_range:
+              arguments:
+                min_value: 0
+                inclusive: true
+      - name: total_voters
+        description: >
+          Distinct voters in this block carrying >=1 of the 22 major district values
+          (the count-once denominator; ~= the block's total voters since County covers
+          essentially all)
+        data_tests:
+          - not_null
+          - dbt_utils.accepted_range:
+              arguments:
+                min_value: 1
+                inclusive: true
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          name: serve_block_coverage_unique_key
+          arguments:
+            combination_of_columns: [block_geoid, served_set]
+      - dbt_utils.expression_is_true:
+          name: serve_block_coverage_served_le_total
+          arguments:
+            expression: "served_voters <= total_voters"
--- a/dbt/project/models/intermediate/civics/int__serve_block_coverage.sql
+++ b/dbt/project/models/intermediate/civics/int__serve_block_coverage.sql
@@ -1,0 +1,129 @@
+{{ config(materialized="table", tags=["intermediate", "civics", "serve"]) }}
+
+-- int__serve_block_coverage (DATA-1993): the count-once engine. One row per
+-- (census block, served_set) carrying the count of DISTINCT served voters and the
+-- block's total voters. people_served reads it: count-once for a set =
+-- sum(served_voters / total_voters * block_population) over served blocks.
+--
+-- WHY a voter scan (and not the substrate): count-once is a distinct-PERSON union
+-- across overlapping district types. Within one type the cohort districts partition a
+-- block's voters, but a voter served by BOTH their city and their county cannot be
+-- de-duplicated across types from counts alone -- that cross-type overlap is exactly
+-- the "represented by 2+ officials" gap. So the union needs voter IDs, which the
+-- counts-only substrate (and int__l2_block_district_map) dropped at the
+-- count(distinct lalvoterid) step. This model scans int__l2_nationwide_uniform once --
+-- the "scanned, never surfaced" node from TDD 7.1 -- and emits ONLY block-grain counts
+-- (PII-free). It is a model, not an inline CTE, because the ~230M-row unpivot is heavy
+-- (Dan's "break out only if it grows complex" clause); it restores the original TDD
+-- DAG.
+--
+-- Scoped to the SAME 22 district types as the substrate
+-- (get_l2_major_district_columns),
+-- so count-once <= count-multiple holds while the substrate is at 22 types. state is
+-- derived from the block's geoid FIPS prefix (the substrate's rule), matching the
+-- substrate block-for-block; the '06'=6 numeric coercion is proven (the merged
+-- substrate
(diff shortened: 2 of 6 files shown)


sanjayr1:
delegate review

Both review findings are addressed in 9172c5e:

**delegate-reviewer (ordering test passes vacuously on a missing variant):** made the invariant test strict. It now flags any office_type missing a metric variant, in addition to ordering violations. Every office_type ('all', each L2 type, and 'State') carries all four variants by construction, and I verified 0 offending office_types on the built preview table.

**Cursor Bugbot (empty-normalized district filter):** added the `trim(normalize_l2_district_name(district_value)) != ''` filter to `int__serve_block_coverage`, mirroring `int__l2_block_district_map`. On a point of mechanism: `voter_any` de-duplicates per `lalvoterid`, so an empty-normalized type-row for an already-counted voter cannot inflate `total_voters`. I confirmed this on data: coverage `total_voters` equals the substrate per-block voter total exactly (0 differing blocks across 714,954 served blocks). The filter is a no-op on current data; it keeps the voter grain provably identical to the substrate across future refreshes.

Full reconciliation on the preview schema to follow once CI rebuilds the new commit.

sanjayr1:
## Reconciliation (CI preview schema, Category C verified in Claude)

Reconciled on the preview build (green). The latest commit (9172c5e) adds the empty-normalized filter and the strict invariant test, a verified no-op on the data, and is rebuilding now.

### People Served ladder (`people_served`)

Local North Star (`office_type = 'all'`), ordering holds:

| variant | people |
|---|---|
| count_once | 55,964,976 |
| count_multiple_per_district | 63,882,759 |
| count_multiple_per_seat | 68,712,704 |
| count_multiple_per_org | 71,437,338 |

1,571 cohort districts. Statewide (`office_type = 'State'`, reported separately, never folded into `all`): 28,323,715 across 2 officials. The strict ordering-plus-completeness test returns 0 offending office_types on the built table.

These are the live merged-pipeline numbers, not the validation's run-C figures (50.6M / 56.6M). The merged ladder reads higher because the cohort still includes 3 known wrong-scope matches (NYC SD, Maricopa, Miami-Dade, all flagged `requires_review`) pending the DATA-1989 upstream overrides, and it covers the 22 v1 district types. Per the band-not-fixed rule, tests assert the ordering invariant and wide bands, never fixed counts.

### `official_constituents`
1,790 officials = 1,790 distinct slugs (unique), 1,708 with a census match, 82 non-binding (78 deferred special-district type, 4 name-drift), 2 statewide, 12 `requires_review`, constituents >= 0.

### Codex #5 (collision-immunity), proven
9 real L2 normalized-key collisions exist. count-once de-dups on `lalvoterid` and joins the substrate at block grain (never the district-level aggregations on name), so a collision cannot inflate it. The count_multiple to count_once gap (7.9M) is exactly the cross-type plus collision de-dup.

### Codex #2 (basis consistency), exact
The count-once denominator (`coverage.total_voters`) equals the substrate per-block voter total exactly: 37,021,122, with 0 differing blocks across 714,954 served blocks.

### Reviews
Codex plan and PR gates: GO after revisions (statewide redesigned as a separate `office_type='State'` read from T5's exact rows; seat id namespaced; strict invariant test). Other findings verified as non-issues with data. Delegate-reviewer blocker and Cursor Bugbot medium both addressed in 9172c5e (responses in the thread above).

### Follow-ups (not blockers)
- DATA-1989: file the wrong-scope overrides (moves the live count-once down toward the conservative ~42M).
- DATA-2009: the deferred high-undercount flag (needs voterless-block geographic assignment).
- Substrate type-widen (T3 follow-up): add the cohort's deferred major types (community college, county board of education) to recover count-multiple and per-official coverage.

Not self-merged; opened for human review.

sanjayr1:
On the latest delegate-reviewer finding (the strict ordering+completeness test could fire `missing_variant` if a local type appears in count-multiple but not in count_once): this is a verified non-issue, and I am not applying the suggested `coalesce(count_once, 0)` because it would introduce a wrong number and mask a real bug. Reasoning, with data:

1. **The two type sets are identical by construction.** count_once (`int__serve_block_coverage`) and count-multiple (`district_census_stats`, a substrate rollup) both derive from the same upstream: a scan of `int__l2_nationwide_uniform` with the same `get_l2_major_district_columns` unpivot, the same `normalize_l2_district_name`, and the same resolver cohort. A `(state, type, name)` that binds `district_census_stats` (so the type is in `cm`) necessarily has voters in its blocks who match the cohort in the coverage scan (so the type is in count_once). Verified: 0 offending office_types on both CI builds, and the 21 binding local types match exactly between count_once and count-multiple-per-district.

2. **The suggested fix would be incorrect.** Left-joining `cm` into count_once and coalescing a missing type to 0 would emit `count_once = 0` for a type whose districts have population (`cm > 0`). count_once is the true distinct-person union, which is greater than 0 whenever the districts have people, so a 0 there is simply wrong. Worse, it would silently mask the divergence the completeness check exists to catch: if count_once ever lacked a type that count-multiple has, that is a real inconsistency (the coverage scan and the substrate diverging from a shared source) that should fail loudly, not be coalesced away.

The original blocker (the null-tolerant test passing vacuously) is resolved by the strict test in 9172c5e, which folds in the null check exactly as the bot's "alternatively" suggestion described.

Not re-triggering, to avoid a review loop. Flagging for the human reviewer.

PR #523, opened 2026-06-22
[DATA-1734] Add race-key guard to candidacy ER

danpelota:
In candidacy-stage entity resolution, same-person records were bridging a candidate's distinct races into one cluster. Example: one person's Racine school board, city council, and county board candidacies all merged into a single cluster.

The existing `br_race_id` guard only worked when both sides carried a `br_race_id`, so ddhq (which carries none) slipped through and bridged races.

This change replaces the office-name-only `br_race_id` guard with a single race-key post-prediction filter. Keep a pair only if:

- a strong office-name match (gamma 3 == JW >= 0.88), or
- a shared `br_race_id`, or
- it conflicts on none of `br_race_id` / `district_identifier` / `office_type` (each null-wildcards).

Also normalizes `office_type` in the prematch model so the equality is not defeated by noise.

## Validation

Built the prematch model and ran the matcher against it. Versus baseline:

- 504 pairs dropped (439 involve ddhq), 0 gained, 0 same-`br_race_id` regressions
- clusters larger than 4: 57 -> 38; max cluster size 8 -> 7
- coverage flat: ddhq 72.7 -> 71.9%, techspeed 35.8 -> 35.6%, ballotready 21.4 -> 21.2%, gp_api 8.0%
- the Racine example splits into 3 correct clusters

---

> [!NOTE]
> **Medium Risk**
> Changes core Splink post-prediction logic and prematch `office_type` values, which can drop previously linked pairs and alter cluster membership across civics ER outputs.
> 
> **Overview**
> Stops candidacy-stage entity resolution from merging **one person’s different races** into a single cluster when identity signals are strong but race signals are weak (notably **DDHQ**, which has no `br_race_id`).
> 
> **Matcher (`candidacy.py`):** Replaces the old **`br_race_id` + office-name** drop rule with a **race-key** post-prediction filter. A pair is kept only if there is a **strong office-name match** (`gamma_official_office_name >= 3`, JW ≥ 0.88), **matching `br_race_id`**, or **no conflict** on `br_race_id`, `district_identifier`, and `office_type` (nulls wildcard on each field).
> 
> **Prematch (`int__er_prematch_candidacy_stages.sql`):** **`office_type`** is normalized for that race-key equality—local legislature labels collapse to **`City Council`**, broad buckets (`other`, `local`, `county`, `state`) become **NULL** so they do not falsely split or block pairs, and remaining values are trimmed/`initcap`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 723e59d422a50a8aca466c80c61e2d64a7b53612. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/intermediate/civics/int__er_prematch_candidacy_stages.sql
+++ b/dbt/project/models/intermediate/civics/int__er_prematch_candidacy_stages.sql
@@ -405,7 +405,25 @@ select
     party,
     candidate_office,
     office_level,
-    office_type,
+    -- Normalize office_type into a clean candidacy-ER race-key discriminator.
+    case
+        when
+            lower(trim(office_type)) in (
+                'alderman',
+                'alderperson',
+                'town council',
+                'town chair',
+                'township council',
+                'city commission',
+                'city commissioner',
+                'city council'
+            )
+        then 'City Council'
+        -- Non-discriminating buckets -> NULL so the race-key clause wildcards.
+        when lower(trim(office_type)) in ('other', 'local', 'county', 'state')
+        then null
+        else nullif(initcap(trim(office_type)), '')
+    end as office_type,
     nullif(district_raw, '') as district_raw,
     district_identifier,
     election_date,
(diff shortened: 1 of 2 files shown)


datawithtristan (approved):
Looks good, no comments for changes

PR #532, opened 2026-06-24
Standardize pandas pin and stop dependabot floor churn

danpelota:
## What

Standardizes the pandas requirement to `>=2.3.1,<3.0.0` across the monorepo and changes dependabot so it stops editing the version floor on every release.

This is the monorepo + uv follow-up to #499. That PR tried to widen the dbt pin to `<4.0.0` (allowing pandas 3.0) and was correctly closed. Since then we moved to a monorepo and standardized on uv, so the pin lives in several subprojects.

## Pin changes

- `dbt/pyproject.toml`: `>=2.2.3` to `>=2.3.1`
- `analytics/pyproject.toml`: `>=2.2.3` to `>=2.3.1`
- `matcha/pyproject.toml`: unpinned `pandas` to `>=2.3.1,<3.0.0`
- `airflow/pyproject.toml` and `airflow/astro/requirements.txt` already used `>=2.3.1,<3.0.0`

The `>=2.3.1` floor is honest for `requires-python >=3.14` (pandas 2.2.3 predates 3.14 wheels). The `<3.0.0` ceiling stays because pandas 3.0 is breaking (copy-on-write default, new PyArrow string dtype, removed APIs), so we do not want it pulled in automatically. Resolved version stays 2.3.3 in all four lockfiles, so no runtime change.

## Long-term fix

Added `versioning-strategy: increase-if-necessary` to the three pip blocks in `.github/dependabot.yml`. Minor and patch updates now flow through `uv.lock` only, and the pyproject floor is left alone as long as the new version satisfies the existing range. Dependabot opens a manifest-editing PR only when a version falls outside the range, i.e. a real major like pandas 3.0, which we then review deliberately. This removes the recurring need to hand-merge floor bumps while keeping the protective ceiling and the major-version signal.

## Note

`uv lock` in dbt also synced stale pyspark lock metadata to `<5.0.0`, matching the already-merged pyproject from #521. The committed lock was just out of date.

## Out of scope

Dependabot only watches `/`, `/airflow/astro`, and `/dbt`. It does not watch `airflow/`, `analytics/`, or `matcha/`, so those pins are now consistent but unmonitored. Expanding coverage would add PRs, so it is left for a separate decision.

---

> [!NOTE]
> **Low Risk**
> Dependency constraint and automation-only changes; resolved pandas stays 2.3.3 with no application logic touched.
> 
> **Overview**
> Aligns **pandas** across `analytics`, `dbt`, and `matcha` to **`>=2.3.1,<3.0.0`** (raising floors where needed and pinning matcha), with matching **`uv.lock`** metadata. The **`<3.0.0` cap** is unchanged so pandas 3.x is not pulled in automatically.
> 
> Adds **`versioning-strategy: increase-if-necessary`** to all three pip entries in **`.github/dependabot.yml`**, so minor/patch updates should stop repeatedly bumping manifest floors when the existing range already allows the new version.
> 
> **`dbt/uv.lock`** also updates the recorded **pyspark** specifier to **`<5.0.0`**, matching `dbt/pyproject.toml` (lock metadata catch-up, not a separate dependency change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f4bcf1dae1852e0be10e1772071398eb0584727. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danpelota:
delegate review

hhkarimi (approved):

PR #549, opened 2026-07-01
[DATA-2063] Codify active serve users as a gold mart model

sanjayr1:
## What

Adds `users_serve_active` (mart_analytics): the canonical gold-layer membership table for "active serve users" — one row per user who completed Serve onboarding by sending an SMS poll AND has pledged. It is a thin, filtered projection of `int__serve_active_user`, which owns the definition (DATA-1359). Also adds a `users_serve` semantic model + `active_serve_users` metric (mirror of `sem_analytics__users_win.yml`), so the count exists in the metric layer as well.

Why now: the dashboard tool is about to lose access to staging/intermediate schemas, and the active-user definition is currently re-derived in dashboard SQL that mixes mart + staging + intermediate tables. This gives dashboards one gold-schema source of truth; the analytics owner repoints the dashboard query after this lands (handoff note prepared separately).

## Why it is safe

- Purely additive: no existing model changes. The resolver (`int__serve_district_resolution`) and `users_serve_base` keep reading the intermediate, exactly as before — no dependency cycle.
- The model body is a projection with a flag filter; it cannot express a second definition of "active", so it cannot drift from the intermediate.
- View materialization (like `users_serve_base`): zero-copy over the intermediate table, always consistent within a build.
- The semantic model only exposes existing `users_serve_base` columns; the measure is a case-when count over the governed flag.
- Definitional note: the model's pledge notion is "any pledged campaign" (owned by `int__serve_active_user`, proven equivalent to the dashboard's latest-candidacy pledge on this cohort with zero membership delta). The model is canonical going forward.
- Access-model assumption for the DE lead: Unity Catalog views read underlying tables with owner rights, so the dashboard principal needs SELECT only on `mart_analytics` (same mechanism `users_serve_base` relies on today). Flagging explicitly since the Sigma schema-access cut is the motivation — please confirm the dashboard principal retains `mart_analytics` SELECT under the new grants.

## Review and verification

- Independent adversarial review (a different model) was run on the implementation plan; its findings (prod-gate before the dashboard handoff, metric verification beyond YAML parse, YAML anchor placement) are incorporated. A diff-level pass runs after CI builds.
- After the CI preview builds, these checks will be posted below: membership parity (preview view vs prod intermediate, exact match expected), dashboard-cohort reconciliation (rewritten mart-only query vs the current staging-based dashboard query, anti-join both directions, zero diff expected), and metric-expression equivalence (case-when count over `users_serve_base` == membership count).
- dbt tests: `unique` + `not_null` on `user_id`.

Refs DATA-1359, DATA-2063 (https://app.clickup.com/t/86ajb360m). Related: #529.

---

> [!NOTE]
> **Low Risk**
> Additive analytics views and YAML only; definition stays in `int__serve_active_user`, so no change to upstream business logic or existing marts.
> 
> **Overview**
> Adds **`users_serve_active`** in `mart_analytics`: a view that lists one `user_id` per **active serve user** (SMS poll sent **and** pledged), by filtering `int__serve_active_user`—so dashboards can use gold schema only instead of re-deriving the rule from staging/intermediate SQL as Sigma loses those grants.
> 
> Also adds **`sem_analytics__users_serve.yml`**: a semantic model on `users_serve_base` with Serve engagement dimensions and the **`active_serve_users`** metric (sum of `is_active_serve_user`), aligned with the Win semantic-layer pattern.
> 
> **`m_analytics.yaml`** documents the new mart (`unique` / `not_null` on `user_id`) and how it relates to `users_serve_base` and `users_win_candidacy` for funnel vs empowered-EO cohorts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5139e9fc6c0db31e4612a2297684ea99537c519. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/marts/analytics/m_analytics.yaml
+++ b/dbt/project/models/marts/analytics/m_analytics.yaml
@@ -580,6 +580,31 @@ models:
           arguments:
             expression: "last_activity_at >= first_activity_at"
 
+  - name: users_serve_active
+    description: >
+      Canonical gold-layer membership table for "active serve users" (epic
+      DATA-1359): one row per user who completed Serve onboarding by sending
+      an SMS poll AND has pledged. Thin projection of int__serve_active_user,
+      which owns the definition; exists so dashboards read the definition
+      from a gold schema (Sigma is losing staging/intermediate access).
+
+      Grain: one row per user_id, active serve users only — joining this
+      model IS the active-user filter.
+
+      For component flags and correct funnel denominators over all users,
+      use users_serve_base (same upstream definition). For the "empowered
+      elected officials" dashboard cohort, join users_win_candidacy and
+      apply icp_office_serve and the internal-email exclusion there.
+
+    columns:
+      - name: user_id
+        description: >
+          Product-DB user id (primary key). Joins to users_serve_base.user_id,
+          users_win_candidacy.user_id, and mart_civics.users.user_id.
+        data_tests:
+          - unique
+          - not_null
+
   - name: users_serve_base
     description: >
       User-grain snapshot for Serve product metrics and onboarding funnel.
--- a/dbt/project/models/marts/analytics/m_analytics.yaml
+++ b/dbt/project/models/marts/analytics/m_analytics.yaml
@@ -582,7 +582,7 @@ models:
 
   - name: users_serve_active
     description: >
-      Canonical gold-layer membership table for "active serve users" (epic
+      Canonical gold-layer membership view for "active serve users" (epic
       DATA-1359): one row per user who completed Serve onboarding by sending
       an SMS poll AND has pledged. Thin projection of int__serve_active_user,
       which owns the definition; exists so dashboards read the definition
--- a/dbt/project/models/marts/analytics/sem_analytics__users_serve.yml
+++ b/dbt/project/models/marts/analytics/sem_analytics__users_serve.yml
@@ -55,9 +55,11 @@ metrics:
     description: >
       Count of active serve users: completed Serve onboarding by sending an
       SMS poll AND pledged. Definition owned by int__serve_active_user,
-      surfaced through users_serve_base; the gold membership table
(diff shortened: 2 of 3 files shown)


sanjayr1:
## Preview verification (CI run 70471886449261, schema `dbt_cloud_pr_70471823431465_549`)

All checks pass exactly.

| Check | Result |
|---|---|
| Membership parity | preview `users_serve_active` = 1,038 rows == prod `int__serve_active_user where is_active_serve_user` = 1,038 |
| Metric-expression equivalence | prod `users_serve_base where is_active_serve_user` = 1,038 == membership count (the semantic measure counts the same population) |
| Dashboard-cohort reconciliation | rewritten marts-only query = 718 == current staging-based dashboard query = 718; anti-join both directions = 0 and 0 (user-for-user identical) |
| dbt tests | `unique` + `not_null` on `user_id` green in CI |

**Access model, now verified rather than assumed:** the dashboard service principal reads through the `mart_analytics_readers` / `mart_civics_readers` groups (marts-only read model). The new view lives in `mart_analytics`, and reading a view over an intermediate is the same Unity Catalog owner-rights mechanism `users_serve_base` already relies on.

**Semantic-layer note (routed to DATA-2030, not a blocker here):** `dbt sl query --metrics active_serve_users` compiles, but the semantic-layer environment resolves against `dbt_staging.users_serve_base`, which predates the column added in #529 — so live SL queries fail until that environment rebuilds or is repointed at prod. The metric's correctness is covered by the SQL equivalence above and the CI parse.

First CI run failed on MetricFlow semantic-manifest validation (the `user` + `registered_at` entity-dimension pairing is owned by `sem_analytics__users_win`); fixed in 3c85536 by using the serve-native `first_sms_poll_sent_at` as the agg time dimension. Re-validated with a remote parse before pushing.

sanjayr1:
delegate review

PR #553, opened 2026-07-02
Source the people-api voter district columns from the shared macro

danpelota:
Rebased onto main. The original version of this PR also merged the two L2 district-column macros; #812 has since done that, so this PR is now scoped to the part that is still missing: the people-api voter marts hand-maintain their own copy of the district list, and it has drifted.

## What changed

- `m_people_api__voter` sources its district columns from `get_l2_district_columns()` in both CTEs instead of a hand-maintained inline list (223 lines per CTE). The ~19 district-adjacent columns that are not in the macro (`AddressDistricts_Change_Changed_*`, `Proposed_*`, `Precinct`, `FIPS`, the party-adjacent zone/convention columns) stay listed explicitly, since they are part of the people-api Voter contract but not the electoral-district list.
- `get_l2_district_columns` gains an optional `table_alias`, so the joined final CTE can emit `tbl_updated.` + "\`Col\`" without a second copy of the list. Default is a no-op; every existing call site is byte-identical.
- `m_people_api__districtvoter` DISTRICT_COLUMNS is synced with the same set. It is a Python model, so it cannot call the macro; the list stays hand-maintained but now matches.
- `LOADER_ADDED_COLUMNS["Voter"]` in people-api-loader declares the seven new columns. They are absent from the prod serving baseline, so without this the loader's `validate` schema-diff reads them as drift.

Intentionally not touched: `l2_district_columns()` in `l2_column_classifications.sql`. The earlier version of this PR rewired it to the macro; it is currently unused and its siblings all read the seed, so leaving it alone keeps that file coherent.

## Why the drift matters

The voter mart is where the people-api Voter table gets its district columns, and `m_people_api__districtvoter` unpivots that same set into the DistrictVoter rows the app queries. Seven L2 district types are present in the source and in `m_people_api__district` (which already unpivots the macro) but never reached either voter mart, so no voter is linked to a district of those types.

## Verification

Columns added, none dropped. Compiled `m_people_api__voter` on main and on this branch and diffed the final select list:

```
main:   356 output columns
branch: 363 output columns
added:  4H_Livestock_District, Community_College, Judicial_Chancery_Court,
        Judicial_Justice_of_the_Peace, Soil_and_Water_District,
        Soil_and_Water_District_At_Large, State_Board_of_Equalization
dropped: none
duplicates: none
```

All seven exist in the upstream source and in none of the current prod mart:

```sql
select table_name, column_name
from goodparty_data_catalog.information_schema.columns
where table_schema = 'dbt'
  and table_name in ('int__l2_nationwide_uniform_w_haystaq', 'm_people_api__voter')
  and column_name in (<the seven>)
-- 7 rows, all int__l2_nationwide_uniform_w_haystaq; 0 rows for m_people_api__voter
```

`DISTRICT_COLUMNS` in the districtvoter model is now exactly the 230 macro types plus the same 19 non-macro columns the voter mart keeps, with no duplicates.

`people-api-loader`: `ruff check` clean, `pytest tests/test_validate.py` 79 passed.

## Before merge

`m_people_api__voter` is a ~5-hour full rebuild tagged `monthly`, so it was compile-verified here, not built. It should get a `dbt build` to confirm the seven added columns populate, and `m_people_api__districtvoter` should be built after it (it reads the voter mart and is incremental on `updated_at`).

The DistrictVoter rows for the seven new types only appear once the voter mart carries them, so run the two in that order.


danpelota:
delegate review

danpelota:
delegate review

danpelota:
delegate review

PR #557, opened 2026-07-02
[DATA-2055] Widen election-api race date window to 2-month grace period

sanjayr1:
## Summary

The election-api Race table loses a race the day after its election date because both dbt-side date filters only kept races from 1 day in the past forward. Campaign plan sections look a race up by `br_hash_id` with no date filter, so the lookup 404s the morning after an election (DATA-2055). Decision on the ticket: widen the lower bound to a 2-month grace period.

- `m_election_api__race.sql`: lower bound of the `election_date` window is now `current_date() - interval '2 months'` (was `- interval '1 day'`). Upper bound unchanged at 2 years out.
- `write__election_api_db.py`: race filter lower bound is now `add_months(current_date(), -2)` (was `date_sub(current_date(), 1)`). `add_months` keeps the lower bound on the same calendar-month arithmetic as the SQL interval rather than a day-count approximation. The lower bound is inclusive (`>=`), matching the mart's `between` exactly (tightened after review; the first revision used strict `>`). The writer's upper bound keeps its pre-existing 730-day cap vs the mart's calendar 2 years, harmless because the writer never deletes inside the 2-year retention. Comments in both files say to keep the lower bounds in sync.
- The 2-year retention DELETE is untouched; only its stale comment (which still said 1 day) is corrected.

election-api itself needs no change: it does not date-filter races by default, and the zip/onboarding search already defaults `electionDateFrom` to today app-side, so past races will not surface there.

## Tests

- New singular test `assert_election_api_race_retains_recent_past_races`: every upstream race inside the grace window that passes the mart's other gates (place retained, position matched) must be retained. This fails loudly if the window is ever re-narrowed, and passes vacuously if the upstream population is legitimately empty. Window edges padded a day against build-vs-test-time date skew.
- `mart_election_api_race_win_number_estimate_coverage_warn` now scopes explicitly to upcoming races. Its documented metric was always "upcoming races"; before this PR the mart window guaranteed that implicitly. The explicit filter keeps the metric and its 15/25 percent thresholds meaning the same thing after the widen.

## Verification (PR preview schema)

- Confirm the 11 known recently-passed races (May/June 2026 election dates) appear in the preview `m_election_api__race` by `br_database_id`.
- Confirm the row-count delta vs prod is bounded: recently-passed races added, no change to future rows.

Results will be posted as a PR comment after the CI build.

## Observation for a follow-up ticket (not changed here)

The writer's incremental `updated_at` gate is currently a no-op: the filtered dataframe inside the watermark loop is discarded (`df = df.filter(...)` rebinds only the loop-local name; the load loop uses the original dataframes). Every run therefore writes the full window, which is why no one-time backfill is needed after this merge: the missing recently-passed races reappear on the first prod run. Worth noting before anyone "fixes" that gate: activating it would skip rows whose BallotReady `updated_at` predates the Postgres watermark, exactly the backlog this PR recovers, so it needs a deliberate design pass rather than a drive-by repair.

### DIFF
--- a/dbt/project/models/marts/election_api/m_election_api__race.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__race.sql
@@ -133,9 +133,9 @@ left join
 where
     tbl_race.place_id in (select id from {{ ref("m_election_api__place") }})
     -- 2-month grace period keeps recently-passed races serveable (the campaign
-    -- plan reads races after election day). Keep anchored to the same bounds as
+    -- plan reads races after election day). Keep the lower bound in sync with
     -- the race filter in write__election_api_db.py, which otherwise re-narrows
-    -- this window (boundary-day membership differs: between vs strict >)
+    -- this window
     and tbl_race.election_date
     between current_date() - interval '2 months' and current_date() + interval '2 years'
     -- Race -> Position -> District -> ProjectedTurnout is the chain the API
--- a/dbt/project/models/marts/election_api/m_election_api__race.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__race.sql
@@ -42,8 +42,9 @@ with
     -- election cycle. win_number is carried here too but has no live source:
     -- BallotReady supplies no win number, and the only values that exist come
     -- from the 2023-2025 HubSpot archive (int__civics_candidacy_2025), whose
-    -- past elections fall outside this mart's forward-looking date filter, so
-    -- win_number is null for every row this mart emits, and consumers fall back
+    -- past elections fall well outside this mart's date window (a 2-month
+    -- grace period at the oldest), so win_number is null for every row this
+    -- mart emits, and consumers fall back
     -- to a computed estimate. Aggregate with any non-null value; downstream Race
     -- rows that share a gp_election_id (i.e. multiple BR race stages for the
     -- same election cycle) will all carry the same values.
@@ -132,8 +133,9 @@ left join
 where
     tbl_race.place_id in (select id from {{ ref("m_election_api__place") }})
     -- 2-month grace period keeps recently-passed races serveable (the campaign
-    -- plan reads races after election day); must match the race filter in
-    -- write__election_api_db.py, which otherwise re-narrows this window
+    -- plan reads races after election day). Keep anchored to the same bounds as
+    -- the race filter in write__election_api_db.py, which otherwise re-narrows
+    -- this window (boundary-day membership differs: between vs strict >)
     and tbl_race.election_date
     between current_date() - interval '2 months' and current_date() + interval '2 years'
     -- Race -> Position -> District -> ProjectedTurnout is the chain the API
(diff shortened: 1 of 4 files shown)


sanjayr1:
Verification on the PR preview schema (`dbt_cloud_pr_70471823431465_557`) vs prod (`dbt`), as promised in the description. All checks pass.

**1. The 11 known dropped races are recovered.** Every `br_database_id` from the ticket (2094742, 2266274, 2225675, 2266413, 2266638, 2913201, 2815927, 2252083, 2910469, 2106624, 2915827; election dates 2026-05-12 through 2026-06-09) is present in the preview `m_election_api__race` and absent from the prod mart.

**2. The delta is bounded to exactly the grace window.**

| bucket | preview | prod | delta |
|---|---|---|---|
| future window (today .. +2y) | 270,396 | 270,396 | 0 |
| grace past (-2mo .. yesterday) | 41,307 | 0 | +41,307 |

No rows older than the grace window or beyond 2 years in either environment. Preview `min(election_date)` is 2026-05-02 (exactly today minus 2 months, inclusive); `max(election_date)` is identical to prod.

**3. Completeness: zero eligible grace races missing.** Every upstream race in the grace window that passes the mart's place and position gates is retained (41,307 expected, 41,307 retained, 0 missing). This is the same invariant the new singular test enforces.

**4. No grain or null regression.** `id` and `br_database_id` are unique with zero nulls (also `br_hash_id`) in both preview (311,703 rows) and prod (270,396 rows).

**5. Candidacy coverage improves, not degrades.** Grace races carry candidacies for 37.3% of rows vs 2.9% for future races (BallotReady fills candidacies as filings close), so the widen adds the best-covered rows.

**6. The documented boundary-day edge is quantified and harmless.** 3,101 preview rows sit on exactly 2026-05-02 (the inclusive `between` day the writer's pre-existing strict `>` skips; it is the Texas May uniform election date, hence the size). Steady state is unaffected: a race only reaches that boundary after 2 months of being written, and the writer never deletes inside the 2-year retention.

Independent adversarial review (Codex, code plus these aggregates): APPROVE-WITH-NITS, no critical or major findings. The one nit is the optional `>=` lower-bound parity, deliberately not taken per the ticket decision to keep the pre-existing edge semantics.

---

**About the red dbt Cloud check (not caused by this PR):** the only failing node in the run is `prospects` (analytics mart), at 22:47Z, with `TABLE_OR_VIEW_NOT_FOUND: goodparty_data_catalog.dbt.organizations`. This PR touches neither model. `prospects` was swept into `state:modified+` because main is ahead of the last prod build, and CI's deferral manifest resolves `ref("organizations")` to `dbt.organizations`, a relation that exists nowhere (prod materializes it at `mart_civics.organizations`, which is present and healthy). Every model and test belonging to this PR built and passed in the same run: `m_election_api__race` and `m_election_api__candidacy` rebuilt at 22:10Z, `write__election_api_db` completed at 22:42Z, and both changed tests pass when run manually against the rebuilt preview (0 missing grace races; win_number coverage 3.88% vs the 15% warn threshold). Other open PRs sweeping `prospects` should fail identically until a successful prod build refreshes the deferral baseline; a CI rerun after that should go green here.

sanjayr1:
delegate review

PR #560, opened 2026-07-06
[DATA-2015] Fix even-year local election data to allow for "no election happened" precincts

nigel-gp:
## What this fixes

The turnout model for even-year local elections (city council, school board, etc.
held the same year as a presidential or midterm election) couldn't tell the
difference between two very different situations:

1. A precinct really did have a local election that year, and a given voter
   chose not to vote
2. A precinct had no local election at all that year

Both cases were being counted the same way ("didn't vote"), which inflated
local-election turnout numbers everywhere - including in places that never
had a local race at all that year.

This fix uses whether a precinct actually recorded a vote in that type of
election as the signal for whether an opportunity to vote existed there in
the first place. If nobody in a precinct ever voted in that election type in
a given year, it's now treated as "no election happened" (excluded) rather
than counted as a missed vote.

This mirrors a fix already made and validated on the model-training side.

## Background

As part of the same effort, a new model, `even_year_primary`, was also built
and trained (on the training side) to predict turnout in statewide primary
elections, separate from local elections. That model is not wired into this
nationwide inference pipeline yet - this PR only covers the even-year local
election fix. Primary-election inference support (routing, feature building,
`election_code` mapping for `even_year_primary`) would be a follow-up PR.

## Test plan
- [x] Updated tests to check the new even-year-local behavior
- [x] All existing tests still pass (23 passed)
- [x] Code style checks (ruff) pass
- [ ] Not yet tested against real production data (no dbt Cloud CLI access in
      this session)

---

> [!NOTE]
> **Medium Risk**
> Changes core vote-history SQL for `even_year_local` predictions nationwide; wrong opportunity logic would skew projected ballots, but behavior is aligned with training and covered by updated unit tests.
> 
> **Overview**
> Fixes **even-year local** (`OtherElection`) vote-history eligibility in nationwide LightGBM turnout inference so precincts with **no local race** are excluded instead of counted as eligible non-voters.
> 
> **Before:** eligibility used `year % 2 == 0` and treated every even-year eligible non-voter as `0.0`, which inflated `even_year_local` features/targets nationwide. **After:** routing is by column prefix (`AnyElection` vs `OtherElection`). `_odd_op_years` becomes `_op_years` and includes even years for `OtherElection`; `turnout_historical_precincts` supplies `opp_{year}` for both odd and even local years. `OtherElection` non-voters are eligible only when `opp_{year} = 1` (no odd-year state-list shortcut). Odd-year `AnyElection` behavior is unchanged.
> 
> Unit tests add `test_even_year_opportunity_has_no_state_shortcut` and update fixtures/assertions for `opp_2024` and combined opportunity years.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1dfa0ffb64673330bda086011296e99f6f8050c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/intermediate/l2/int__voter_turnout_lgbm_inference.py
+++ b/dbt/project/models/intermediate/l2/int__voter_turnout_lgbm_inference.py
@@ -610,9 +610,7 @@ def model(dbt, session):
     for inference_year in inference_years:
         op_years = _op_years(election_cols, l2_col_set, inference_year)
         if op_years:
-            session.sql(_opp_view_sql(op_years, catalog, precincts_schema)).createOrReplaceTempView(
-                "_hp_opp"
-            )
+            session.sql(_opp_view_sql(op_years, catalog, precincts_schema)).createOrReplaceTempView("_hp_opp")
         # ~206k precincts per year (~150 MB) — toPandas is safe; eagerly materialized here so
         # the per-year _hp_opp view is consumed before the next year overwrites it.
         pdf = session.sql(
--- a/dbt/project/models/intermediate/l2/int__voter_turnout_lgbm_inference.py
+++ b/dbt/project/models/intermediate/l2/int__voter_turnout_lgbm_inference.py
@@ -9,9 +9,17 @@ ballots_projected. Output feeds int__model_prediction_voter_turnout.
 Collapse-specific changes vs the per-state macro (all verified against the
 nationwide table on 2026-06-29):
   - State comes from the real `state_postal_code` column (no lit(state_code)).
-  - The odd-year opportunity logic is ROW-LEVEL (state_postal_code IN (...)).
+  - The odd-year (`AnyElection`) opportunity logic is ROW-LEVEL (state_postal_code IN (...)).
   - The vote-history "voted" test is BOOLEAN (cols are BooleanType nationwide),
     not = 'Y'.
+
+Even-year (`OtherElection`, local) eligibility is guided solely by per-precinct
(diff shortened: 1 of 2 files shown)


sanjayr1:
Ran an independent adversarial review pass (Codex) on this before merging. It raised two blocking-level concerns; I verified both are non-issues:

- **Shared `opp_{year}` flag mixing prefixes**: checked `turnout_historical_precincts` directly — `AnyElection` rows are 100% odd-year (2001–2025) and `OtherElection` rows are 100% even-year (2000–2024), zero mixed-parity rows. So the collapsed flag is exactly equivalent to a prefix-specific one on current data, and it deliberately mirrors the training-side vote_history_sql, which is the parity that matters.
- **`PresidentialPrimary_*` falling into the OtherElection branch**: can't happen — `_VH_COL_RE` is anchored to exactly the four prefixes `(General|Primary|OtherElection|AnyElection)_YYYY`, so `_detect_election_cols` never emits anything else.

One follow-up worth filing (not blocking): a defensive assertion or data test on the parity invariant (AnyElection odd / OtherElection even) in `turnout_historical_precincts`, so a future mixed-parity load fails loudly instead of silently changing eligibility. If we ever want a prefix-specific opportunity flag it has to be a joint training+inference change.

Merging into the feature branch. Old-vs-new output validation runs on PR #543's preview build next; expectation is even-year Local coverage drops (odd-cycle-local states fall back to the legacy feed via the carve) and General is unchanged.

PR #577, opened 2026-07-08
Map TechSpeed Oregeon state typo to OR in clean_states

sanjayr1:
## What

Adds one mapping row to the `clean_states` seed: `OREGEON,OR`.

## Why

The nightly build is failing two error-level tests:

- `is_state_abbreviation_candidate_state` (mart_civics.candidate): 139 rows
- `is_state_abbreviation_election_state` (mart_civics.election): 4 rows

Root cause: TechSpeed's 2026-07-07 file (`techspeed_to_goodparty/20260707/candidates_20260707_clean_utf8_nobom.csv`) spells Oregon as "Oregeon" on 139 rows. The value is not in the `clean_states` seed, so the staging fallback `coalesce(state_cleaned_postal_code, <raw state>)` passes the raw string through as `state_postal_code`. All 139 people are TechSpeed-only net-new candidates, so the TechSpeed value wins the provider coalesce in the candidate mart. Four TechSpeed-only races leak the value into the election mart; BallotReady's clean `OR` wins everywhere else.

The affected rows are genuine Oregon records (Corvallis, Ashland, Eugene, Gresham, Clackamas County, and more; all 2026-11-03 general; every row has `br_race_id`), so mapping the spelling recovers them correctly. This mirrors the existing `VIRGINA,VA` typo row in the same seed. A vendor re-delivery would not fix the marts because stage dedupe keeps the earliest source file per candidate.

An all-time scan of TechSpeed staging shows "Oregeon" is the only unmapped state value that has ever passed through the fallback.

## Verification

- CI builds `state:modified+` from the seed, so the corrected value cascades through staging, the civics intermediates, and the candidate/election marts in the PR preview schema, where the two failing tests rerun.
- After merge, the scheduled jobs rebuild prod and the nightly build returns green.

## Follow-up (not in this PR)

Consider routing unmapped state values to the `_invalid` quarantine model (a new `unmapped_state` reason) instead of the raw passthrough, so future misspellings quarantine loudly rather than reaching the marts.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 75f1c1f120c6875577d2c15613f2f80d8c3fe002. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sanjayr1:
Superseded: the identical one-row fix (OREGEON,OR in clean_states) merged to main tonight via #580 while unblocking the DATA-2015 release train (stale deferral manifest was pulling the failing civics tests into every PR's CI). Same diff, so closing this one. The tag:monthly CI-job exclude noted on this PR's run was the other half of the fix and is already live on the job config.

PR #578, opened 2026-07-08
[DATA-2091] Surface requested HubSpot contact properties in hubspot_contacts

sanjayr1:
## What

Surfaces the analytics/Growth-requested HubSpot contact properties on
stg_airbyte_source__hubspot_api_contacts, and therefore on
mart_analytics.hubspot_contacts (Sigma's surface) via its existing select *
passthrough: 231 net-new generated columns; the other 46 requested
properties were already exposed and are recorded in the registry instead of
re-emitted.

## How

- seeds/hubspot_contact_property_columns.csv: one row per requested property
  (status existing|generated, column_name, cast_type). Adding a future
  property = adding a row. Doubles as the warehouse-queryable
  property -> column dictionary.
- macros/hubspot_contact_property_columns.sql: run_query accessor over the
  seed (execute-guarded, seed ordered by sort_order) + typed extraction
  macro. HubSpot sends '' for unset values, so typed casts go through
  nullif(..., ''); all typed casts are try_cast/cast_to_boolean on purpose —
  staging and the mart are views, so a plain cast would raise inside Sigma
  at query time on a future malformed value instead of failing a build.
  Loss visibility comes from the rerunnable parity check instead.
- Staging model: -- depends_on hint + loop emitting the generated columns.
  Mirrors the l2_column_classification registry pattern (#552).
- YAML: generated column docs composed from HubSpot's property label +
  description (persist_docs pushes them to the catalog).

## Verification

- Full-table pre-flight on raw before building: every bool/number/date/
  datetime property's values checked castable (counts of non-empty vs
  try_cast survivors: all lossless) — cast map is data-verified, not assumed.
- Remote parse green; dev-environment build green (seed + view + all tests);
  exhaustive render scan (count over every generated column, full table) and
  sample-row checks ran BEFORE this PR — the loop is execute-guarded, so
  parse alone never renders it.
- CI preview parity (comment to follow): per-column full-table value
  checksums (xxhash64 over id+value) on the preview relation vs the same
  extraction computed on raw — catches swapped mappings, not just missing
  ones; plus row-count parity, a mart column-inheritance check, and spot
  value samples per cast type.

## Notes

- Current-state must never come from snapshot__hubspot_api_contacts (its
  wide columns froze 2026-03 when the sync slimmed; it is history-only).
- A subset of requested properties is empty in HubSpot today; they ship as
  typed all-NULL columns and populate as HubSpot fills them.
- Generic property names got DE-chosen disambiguating column names
  (high_priority_issue_id/_summary/_title, number_property); everything else
  keeps its HubSpot internal name. The seed records the mapping for all
  requested properties, including the 46 pre-existing aliases.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 2e3ca99fd9d83b38f2ac36538a9e40ce84fc529c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sanjayr1:
delegate review

sanjayr1:
## Preview verification (against the fc7cb8d CI build)

Value-level parity, not just counts: for every generated column, a full-table
checksum `sum(xxhash64(id, <value as string>))` on the preview relation vs the
same extraction+cast replicated in SQL on raw, both sides inside a single
statement (one Delta snapshot, no sync race).

- Row counts: staging 328,706 = raw 328,706
- **231/231 generated columns checksum-verified — 0 mismatches**
- **Mart inheritance: 231/231 generated columns present on the preview
  `hubspot_contacts`** (the `select *` view re-created correctly via
  `state:modified+`; the merge job selects the same way)
- Spot values per cast type all plausible (booleans, doubles, ISO dates,
  timestamps, enum strings)
- Latency signal: ~40-column dual-side full-scan chunks ran 90–218s; the raw-side
  JSON extraction dominates, the built view itself is what Sigma reads

## On the red dbt Cloud check

The failure is far downstream of this diff, ~2.5h into the civics DAG
(everything this PR touches built green at the start of the run; the preview
relations above ARE this PR's build). PR #576 failed its run in the same
window, and #577 has a states-data fix in flight — this looks like the known
late-DAG source-data flake, not this change. Retriggering; if it reproduces
identically I'll escalate with the node-level evidence.

sanjayr1:
## Adversarial-review disposition (independent reviewer, 2 passes)

An independent model reviewed the diff twice — once diff-only, once grounded
in the preview verification data. Every finding and its disposition:

**Pass 1** (all fixed in fc7cb8d):
1. hs_object_id surfaced as double → now **bigint** (it is HubSpot's Record
   ID; verified exact against `id` on all rows; max ~2.3e11 already exceeds
   int32).
2. No committed guard against silent cast loss → added
   `assert_hubspot_generated_casts_lossless` (per typed generated column:
   raw non-empty count must equal cast survivors).
3. Registry under-guarded → column_name identifier-regex test added; a
   collision with a hand-written column already fails view creation loudly.
4. double branch skipped nullif — aligned with the documented contract.

**Pass 2** (with preview data as context):
1. "Checksum parity can't prove losslessness" — correct by design (both
   sides apply the same cast). Independent proof now attached: lossless
   aggregate ON THE PREVIEW BUILD = **76/76 typed generated columns, 0
   lossy** (matches the pre-flight on raw before any code existed).
2. "warn-only guard can pass CI while values null out" — accepted: severity
   flipped to **error** (next commit). Rationale: HubSpot enforces property
   types at entry, so real loss = format/schema drift = block-worthy.
3. "Mart carries no column docs" — **refuted empirically**: Databricks
   `create view as select *` inherits column comments, and the preview
   catalog shows **231/231 generated columns documented on BOTH staging and
   the mart**. The m_analytics.yaml "columns documented at the upstream
   staging model" convention holds in the catalog Sigma reads.

Reviewer verdict was REQUEST-CHANGES on items 1–2 (now closed with data and
the severity commit) and the docs claim (refuted with catalog evidence).

sanjayr1:
## CI failure root-cause evidence (runs 2 and 3)

Both failed runs died at the same terminal node region, ~2.5h in:
`write__election_api_db` is the last-written relation in this PR's preview
schema for BOTH failures (22:15Z run 2, 00:51Z run 3 — identical code, the
second was an empty-commit retrigger). Everything this PR actually changes
builds green in the first ~10 minutes of the run and is value-verified
against raw (comments above).

PR #576's failed run shows the same terminal cluster in its preview schema
(the election-api write chain). Two unrelated PRs failing at the same
reverse-ETL node — the one that writes to the election Postgres database —
points at the election-api write path, not at either diff. Run 1 of this PR
(same staging/seed code, pre-review-fix) passed earlier in the day, which
brackets the breakage window to this afternoon.

Run 4 (the blocking-severity test commit) is queued. If it fails at the same
node, this PR is blocked on the write-path issue, not on its own changes.
Node-level logs need the dbt Cloud run page:
https://yx558.us1.dbt.com/deploy/70471823433531/projects/70471823438007/runs/70471887539580/

sanjayr1:
Run 4 (the blocking-severity commit): same failure signature, third time —
`write__election_api_db` is again the last-written relation in the preview
schema (03:02Z). Three runs, two different commits, one node.

Corroborating scope: PRs that do NOT sit upstream of the election-api write
(#570–572, #579) merged green today; the failures hit exactly the PRs whose
`state:modified+` selection reaches `write__election_api_db` (this one via
HubSpot staging → civics candidacies → election-api marts; #576 directly).

Holding this PR on the write-path fix. Everything the PR itself changes is
built, value-verified, and documented above; will retrigger once the write
path is healthy.

sanjayr1:
delegate review

sanjayr1:
On the delegate-reviewer blocker (parse-time render of
assert_hubspot_generated_casts_lossless): the premise doesn't hold for dbt's
toolchain — parse renders Jinja for dependency extraction only and never
validates SQL text; compilation happens at execution time with execute=true,
where the registry is loaded and the SQL is complete. Evidence over theory:
this exact test executed and PASSED in the dev build and again inside the
green CI run on cde0982 (a compile failure would have failed that run).

The suggested `{% if execute %}` wrapper is a no-op here (the accessor it
calls is already execute-guarded and returns [] at parse either way). The
genuinely useful hardening is an empty-registry SQL fallback so the rendered
text is valid under any future static analysis — folding that into the
upcoming registry-migration PR, which touches this test anyway.

PR #579, opened 2026-07-09
Register race_opponent staging models

hhkarimi:
## What

Registers four newly ingested `gp_api_db` tables (from the opponent research feature) into the dbt staging layer:

- `gp_api_db_race_opponent_contrast`
- `gp_api_db_race_opponent_research`
- `gp_api_db_race_opponent_standout_action`
- `gp_api_db_race_opponent_summary`

## Changes

- Added each table to the `airbyte_source` source in `models/staging/__sources.yaml`.
- Added `stg_airbyte_source__gp_api_db_race_opponent_*` staging models following the existing gp_api_db staging pattern (passthrough of Airbyte metadata columns, timestamp casting in the staging layer).
- Documented columns and added tests (`not_null`/`unique` on `id`, `not_null` on `campaign_id`, plus `not_null` on `run_id`/`opponent_name` where fully populated) in the model yaml.

## Verification

`dbt build` on the four new models passed: PASS=20 (4 view models + 16 data tests), 0 errors.

Note: `contrast` and `research` are currently empty in the source (0 rows); `standout_action` (7 rows) and `summary` (5 rows) have data. Tests were chosen to hold as those tables populate.

### DIFF
--- a/dbt/project/models/staging/__sources.yaml
+++ b/dbt/project/models/staging/__sources.yaml
@@ -193,6 +193,14 @@ sources:
         description: raw data load from gp_api_db postgres db from table `website_contact`
       - name: gp_api_db_website_view
         description: raw data load from gp_api_db postgres db from table `website_view`
+      - name: gp_api_db_race_opponent_contrast
+        description: raw data load from gp_api_db postgres db from table `race_opponent_contrast`
+      - name: gp_api_db_race_opponent_research
+        description: raw data load from gp_api_db postgres db from table `race_opponent_research`
+      - name: gp_api_db_race_opponent_standout_action
+        description: raw data load from gp_api_db postgres db from table `race_opponent_standout_action`
+      - name: gp_api_db_race_opponent_summary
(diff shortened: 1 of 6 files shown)


hhkarimi:
delegate review

hhkarimi:
Fixed in fd52f65. Moved `config:` to be a sibling of `arguments:` in all four `expression_is_true` tests. Confirmed against the compiled test SQL — each now runs on a subquery filtered by `where updated_at is not null and created_at is not null`, so the null guard is applied. `dbt build` on the four models passes (PASS=32).

delegate review

hhkarimi:
Fixed in dc6ca4d. Switched ```order``` to `{{ adapter.quote("order") }}` to match the repo convention for reserved identifiers (the only other `order` references, in `campaign_position` and the stripe charges staging model, both use it). Compiles to the same backtick-quoted identifier on Databricks, so no behavior change. Build passes.

delegate review

PR #582, opened 2026-07-09
Loosen candidacy election_date match to a 10-day window

danpelota:
## What

Loosen the candidacy-stage entity resolution to treat election dates within 10 days of each other as the same election, instead of requiring an exact date match. Two sources routinely report the same election a few days apart (data-entry / reporting lag), and the exact-date requirement was dropping those cross-source matches.

## Changes

**matcha** (`scripts/configs/candidacy.py`)
- `election_date` now uses `AbsoluteDateDifferenceAtThresholds` (null / exact / within-10-days / else) instead of `ExactMatch`. Exact still scores highest (gamma 2); a within-window match is a distinct middle level (gamma 1).
- The four date-bearing blocking rules use a matching 10-day window (`abs(epoch(try_strptime(...)) - ...) <= 864000`) so near-duplicate-date pairs actually get generated.
- The `gamma_election_date > 0` same-stage guard is preserved: it admits within-window matches but still drops gaps over 10 days, so a candidate's primary, runoff, and general stay in distinct clusters (the candidacy_stage grain holds).
- Uses the `*AtThresholds` builder rather than the raw `AbsoluteDateDifferenceLevel`, which mutates its shared `ColumnExpression` on each resolution and double-wraps `try_strptime` to NULL. Added a test pinning the idempotent SQL.

**dbt** (`models/staging/__sources.yaml`)
- Repoint the `er_source.pairwise_candidacy_stages` source at the live table by dropping the stale `pairwise_candidacy_stages20260527` identifier. This matches the `clustered_candidacy_stages` source (no override) and the freshly reloaded all-time table.

## Audit

Reran the match on the full all-time cohort (~525k records). The window added exactly 214 pairs (205 at a consistent 7-day cross-source offset), all high-confidence:
- ~208 clear true positives — same person, same office, date reported a few days apart across sources.
- ~5-6 false positives from a pre-existing office-name JW collision (school board vs county council), already present in the baseline (31 exact-date instances) — not introduced by this change.
- Zero real regressions; the one differing baseline pair is a correct same-stage drop.
- 0 within-source duplicate clusters; match rates and cluster sizes unchanged.

## Note

The two `er_source` candidacy tables have already been reloaded all-time from this config (backups: `*_bak_predatewindow_20260709`). This PR is the code + source-identifier change behind that reload.

---

> [!NOTE]
> **Medium Risk**
> Changes core Splink blocking, scoring, and post-filters for candidacy crosswalks; mis-tuned windows could merge distinct stages or miss matches, though the 10-day cap and gamma guard are designed to limit that.
> 
> **Overview**
> Candidacy-stage entity resolution now treats **election dates within 10 days** as the same election for cross-source matching, instead of requiring an exact date.
> 
> In **matcha** (`candidacy.py`), `election_date` scoring moves from `ExactMatch` to `AbsoluteDateDifferenceAtThresholds` (null / exact / within-window / else), with a shared `_ELECTION_DATE_WITHIN_WINDOW` constant applied to all date-bearing blocking rules so pair generation and scoring use the same definition. The existing **`gamma_election_date > 0`** same-stage guard still blocks primary/runoff/general merges when dates are more than 10 days apart. **Tests** assert the within-window comparison level and that date-window SQL stays idempotent when resolved repeatedly.
> 
> In **dbt**, the `er_source.pairwise_candidacy_stages` source drops the stale `pairwise_candidacy_stages20260527` table identifier so it points at the live reloaded table.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8eaa258262af8b6ed3bb0f5adc94a933ea51bf88. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danpelota:
delegate review

danpelota:
delegate review

danpelota:
delegate review

PR #580, opened 2026-07-09
Map TechSpeed state typo Oregeon to OR in clean_states

sanjayr1:
One-row addition to the `clean_states` seed: `OREGEON,OR`.

**Why now:** the 2026-07-07 TechSpeed load introduced `state = 'Oregeon'` on 139 `candidate` rows and 4 `election` rows (sample: Eagle Point, OR officials). The `is_state_abbreviation` tests on the civics marts have failed every prod run since 2026-07-07 23:14 UTC, which keeps the prod deferral manifest stale and drags unrelated models into every PR's CI scope. The crosswalk already maps raw-state typos (`VIRGINA,VA`), so this follows the established pattern; the staged column falls back to the raw value only when unmapped.

**Verification:** `Oregeon` is the only non-2-letter value in either mart (checked live), and both failing tests trace to the same staged `state_postal_code` column via `stg_airbyte_source__techspeed_gdrive_candidates`.

Upstream correction of the source rows can be requested from the vendor separately; this keeps the marts and test suite correct either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Single seed lookup row with no logic changes; only affects how a known vendor typo is normalized in downstream joins.
> 
> **Overview**
> Adds **`OREGEON,OR`** to the `clean_states` seed so TechSpeed rows with `state = 'Oregeon'` normalize to Oregon’s postal code instead of passing the raw typo through staging.
> 
> This matches existing typo crosswalks (e.g. `VIRGINA,VA`) and should clear **`is_state_abbreviation`** failures on civics marts tied to `stg_airbyte_source__techspeed_gdrive_candidates` / elections staging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 045e5e43efee36d156faa6735ee76d4652886033. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #587, opened 2026-07-09
[DATA-2095] Fix clustered_elected_officials source_name accepted_values

hhkarimi:
## Why

The `accepted_values` test on `stg_er_source__clustered_elected_officials.source_name` is wrong on main. It lists `["ballotready", "techspeed", "ddhq", "gp_api"]`, but the elected-officials ER pipeline emits a **combined** source. Prod `source_name` values:

| source_name | rows |
|---|---|
| ballotready_techspeed | 524,168 |
| ddhq | 28,520 |
| gp_api | 1,538 |

None are `ballotready`/`techspeed` alone, so the test fails against ~524k prod rows. It hasn't tripped recent main builds only because the model wasn't in their `state:modified+` selection.

## Root cause

`ballotready_techspeed` is intentional — `int__er_prematch_elected_officials.sql` emits `'ballotready_techspeed' as source_name`, and three sibling `accepted_values` tests in this same `__er_source.yaml` already use `["ballotready_techspeed", "gp_api", "ddhq"]`. Only the `clustered_elected_officials.source_name` test had the wrong split list.

## Change

Set the accepted values to `["ballotready_techspeed", "gp_api", "ddhq"]`, matching the pipeline and its siblings.

## Verification

`dbt test --select stg_er_source__clustered_elected_officials` → all 6 tests PASS against prod. Pre-commit passes.

## Note

Surfaced from CI run `70471887646242`. The other three failures there (`experiment_run.status`, `is_state_abbreviation` on candidate/election state) are stale-branch artifacts already fixed on main (#561, #580); those branches just need main merged in. This PR is only the genuine main bug.

### DIFF
--- a/dbt/project/models/staging/er_source/__er_source.yaml
+++ b/dbt/project/models/staging/er_source/__er_source.yaml
@@ -42,7 +42,7 @@ models:
           - not_null
           - accepted_values:
               arguments:
-                values: ["ballotready", "techspeed", "ddhq", "gp_api"]
+                values: ["ballotready_techspeed", "gp_api", "ddhq"]
       - name: first_name
         description: Candidate first name (lowercased, trimmed in prematch).
       - name: last_name


hhkarimi:
Closing: this change is incorrect and introduces a regression. I misread which yaml block belonged to which model. The edit landed on `stg_er_source__clustered_candidacy_stages.source_name`, whose split values [ballotready, techspeed, ddhq, gp_api] are correct (prod: ballotready 310k, ddhq 102k, techspeed 74k, gp_api 37k) — this PR broke that test. The intended target, `clustered_elected_officials.source_name`, was already correct on main ([ballotready_techspeed, gp_api, ddhq]). There is no real bug here: run 70471887646242's failures were all stale-branch artifacts already fixed on main; the branch just needs main merged in.

PR #589, opened 2026-07-10
[DATA-2097] Expose BallotReady URL type (government vs campaign) in civics marts

hhkarimi:
## What and why

Two changes, bundled per request.

### 1. Expose BallotReady URL type on civics marts (DATA-2097)

BallotReady's flat files collapse every website URL under `type = 'website'`, so we cannot tell an official/government site from a campaign or personal one. The CivicEngine (BallotReady) GraphQL API, which we already call in `int__ballotready_person`, returns `urls` with a richer `type` discriminator that includes `government` as distinct from `website`. Example:

- `government` -> ashevillenc.gov/government/meet-city-council/kim-roney
- `website` -> kimroney4asheville.com

This PR carries the full typed array `array<struct<type, url>>` through to the `candidate` and `elected_official_terms` marts as a new `urls` column. Consumers can then filter `type = 'government'` vs `type = 'website'`. No regex or LLM classification needed.

Both marts source the array from the API person feed. The S3 `office_holders` feed carries a government URL for only 101 of ~533k terms, versus ~169k via the API, so the flat file is not a viable source for the office-holder side. The misleading `elected_official_terms.website_url` description was tightened to point at `urls` for the government site.

Files: `int__civics_candidate_ballotready.sql`, `int__civics_elected_official_ballotready.sql`, `candidate.sql`, `elected_official_terms.sql`, `m_civics.yaml`.

### 2. Register gp_api_db_artifact_review staging model

Adds the airbyte source declaration, staging model, and column docs/tests for the `gp_api_db` `artifact_review` table (review verdicts on EO artifacts). Follows the existing `gp_api_db` convention; freshness is disabled to match the sporadically-written `artifact_feedback` sibling.

Files: `__sources.yaml`, `stg_airbyte_source__gp_api_db.yaml`, `stg_airbyte_source__gp_api_db_artifact_review.sql`.

## Validation

- `elected_official_terms`: builds; `urls` populated with real government URLs; non-pre-existing tests pass.
- `candidate`: the intermediate carries `urls` (27,626 rows with a government URL) and the `union all` type reconciles cleanly. Both marts are `table`-materialized, so the new column lands on the next prod run with no schema-evolution work.
- `stg_airbyte_source__gp_api_db_artifact_review`: builds and passes all data tests (`not_null`/`unique` on `id`, `not_null`/`accepted_values` on `verdict`, `not_null` on `resource_id`, `resource_type`, `reviewer_clerk_sub`).
- All pre-commit lint/format hooks pass.

## Notes

- There are 3 pre-existing relationship-test failures in the civics intermediates/marts (`gp_candidate_id` / `gp_elected_official_id` integrity) that fail identically with or without this change. Not introduced here.

ClickUp: DATA-2097

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics_candidate_ballotready.sql
+++ b/dbt/project/models/intermediate/civics/int__civics_candidate_ballotready.sql
@@ -29,6 +29,12 @@ with
             get(filter(urls, x -> x.type = 'twitter'), 0).url as twitter_url,
             get(filter(urls, x -> x.type = 'facebook'), 0).url as facebook_url,
             get(filter(urls, x -> x.type = 'instagram'), 0).url as instagram_url,
+            -- Full typed URL array (normalized to {type, url}); lets marts
+            -- distinguish government/official sites from campaign websites.
+            transform(
+                filter(urls, x -> x.url is not null),
+                x -> struct(x.type as type, x.url as url)
+            ) as urls,
             get(filter(contacts, x -> x.email is not null), 0).email as api_email,
             nullif(
                 regexp_replace(
@@ -93,6 +99,7 @@ with
             person_urls.twitter_url as twitter_handle,
             person_urls.facebook_url as facebook_url,
             person_urls.instagram_url as instagram_handle,
+            person_urls.urls,
 
             coalesce(
                 person_urls.person_created_at, candidacies._airbyte_extracted_at
@@ -141,6 +148,7 @@ select
     twitter_handle,
     facebook_url,
     instagram_handle,
+    urls,
     created_at,
     updated_at
 from deduplicated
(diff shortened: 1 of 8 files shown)


hhkarimi:
Looked at the actual data before applying this.

The premise is inverted: `reviewer_clerk_sub` is 100% populated across all 41 rows, including every row with a null `reviewer_email` (0 nulls in that group). The null-email rows are not automated reviews; they are a single human reviewer whose identity is stored as an email in the `clerk_sub` field. So the suggested `where: reviewer_email is not null` guard would skip exactly the null-email rows that currently do carry a `clerk_sub`, and pass vacuously.

That said, the finding surfaced two real issues, fixed in 268daa8:
- Dropped the `not_null` on `reviewer_clerk_sub`. Not via the email guard, but because it is a non-key reviewer-identity field on a sporadically-written external source and should not be hard-asserted.
- Corrected the `reviewer_email` description (it was not "null for automated reviews") and documented the clerk_sub quirk.

Model and tests rebuild green (7 nodes).

hhkarimi:
delegate review

PR #595, opened 2026-07-13
[DATA-2015] Add election_calendar lookup table for determineElectionCode

nigel-gp:
## Summary

Backs a companion omni PR (#725) that replaces the placeholder ConsolidatedGeneral logic in election-api's determineElectionCode with a lookup against a new Election_Calendar table.

**Architecture:** General is computed once and never changes (fixed federal rule), so it's a plain seed value. Primary is different -- primary dates are legislated and do drift across cycles -- so Primary is served by a **live dbt model** (`int__election_calendar_primary_ballotready`) that recomputes every run directly from BallotReady's scheduled-election data. An edit to BallotReady's table (a corrected date, a newly scheduled state) flows through to `Election_Calendar` on the next daily sync automatically -- no script to re-run. The `election_calendar` seed stays a full, human-readable snapshot of everything currently known (General + Primary, past and future); its own Primary rows are used only as a fallback for a `(state, year)` the live model doesn't resolve. BallotReady is authoritative for Primary, full stop -- even a seed row sourced from L2 ground truth gets superseded if the live model disagrees, by design (one code path, not two).

## Primary-day rule (validated 59/59 against L2 ground truth)

Filter BallotReady elections named `%Primary%` (excluding Runoff/Consolidated/Special), then per `(state, year)` the row with the **most races** wins -- except **Wisconsin**, which gets the opposite (fewest races): WI statute runs two structurally distinct primaries every even year on two fixed windows (Wis. Stat. secs. 5.02(21)/8.05 nonpartisan spring primary in Feb; sec. 8.15 partisan primary in Aug), both literally named "Wisconsin Primary Election" with nothing else distinguishing them, and the spring one always structurally outsizes the partisan one. Confirmed against 3 independently L2-verified cycles before trusting it as a rule rather than a coincidence.

If that filter finds zero rows for a `(state, year)` -- confirmed to happen for DC in 2024, which BallotReady split into party-specific "Republican/Democratic Presidential Primary Election" rows -- retry including Presidential-named rows. Generic fallback, not DC-specific.

Also excludes a Primary candidate whose date exactly equals that year's computed November general day: Louisiana runs an all-candidate "jungle primary" on the standard election day itself (confirmed across every BallotReady cycle, 2018-2030 -- only LA ever hits this), which would otherwise collide with LA's General row on the same `(state, date)` key.

Restricted to even years only: Primary is specifically an even-year concept in this system (serves the `even_year_primary` turnout model).

## What's in this PR

- `dbt/project/seeds/election_calendar.csv` -- full snapshot (458 rows conceptually once combined with the live model: 102 General + up to 356 Primary), with a `source` column (computed/l2/ballotready) documenting provenance.
- `dbt/project/models/intermediate/l2/int__election_calendar_primary_ballotready.sql` -- the live model implementing the rule above.
- `dbt/project/models/marts/election_api/m_election_api__election_calendar.sql` -- combines General (seed, unconditional) with Primary (live model, seed as fallback).
- `airflow/astro/dags/sync_election_api.py` -- new `election_calendar` task group, same atomic-swap delivery pattern as `Projected_Turnout` (and for the same reason: this data gets corrected between cycles, so upsert-by-id would strand superseded rows).
- Cold-start floor tuned to the table's actual current size (410, tracking ~458 real rows) per earlier review feedback that the original floor (50) was far too loose to catch a truncated load.

No maintenance script is included on purpose: by the time an election is over and L2 has confirmed it, candidates have already selected their race, used the turnout models, etc. against whatever BallotReady said beforehand -- there's nothing useful left to backfill after the fact.

## Test plan

- [x] pytest tests/test_sync_election_api.py tests/test_election_api_swap.py -- 26 passed
- [x] ruff check --fix / ruff format (pinned 0.7.4, matching CI) -- clean
- [x] Validated the live model's exact SQL logic directly against Databricks: 59/59 against L2 ground truth, 0 duplicate (state, election_date) keys in the full combined output (458 rows: 102 General + 356 Primary)
- [ ] dbt build of the seed + models against a dev target
- [ ] First sync_election_api DAG run creates Election_Calendar cleanly (cold start, no prior table)


nigel-gp:
delegate review

nigel-gp:
delegate review

sanjayr1:
delegate review

danpelota (changes requested):
Just gating the PR so we can chat before merging 😄

sanjayr1:
Closing in favor of the signed turnout serving design (TDD linked from DATA-2107 / DATA-2117). The primary-day derivation model from this PR is adopted into the warehouse implementation; the calendar delivery surface (seed, mart, DAG group, API lookup) is superseded: no calendar ships outside the warehouse. The derivation, its quirk handling, and its tests carry forward.

PR #596, opened 2026-07-13
Align ER cluster date test with the 10-day match window; fix race_opponent test contracts

sanjayr1:
## Why

The nightly prod build (job `dbt build`) has been red on two steps, and both trace to test contracts that drifted from deliberate changes rather than to broken data or models:

- `dbt build` fails `assert_er_clusters_share_election_date` on 155 clusters every run since the Jul 9 all-time re-cluster. matcha now matches candidacy `election_date` within a 10-day window (`ELECTION_DATE_WINDOW_DAYS` in `matcha/scripts/configs/candidacy.py`), but the test still required one exact date per cluster. Because the test sits on the er staging layer with error severity, its failure skips ~780 downstream nodes nightly (the person spine, `users`, and the civics marts).
- `dbt source freshness` fails `gp_api_db_race_opponent_research` and `_contrast`. Both tables are empty (the product writes rows only when campaigns exercise those flows), and freshness over an empty table reads as ERROR STALE. The sibling `race_opponent` streams on the same connection sync fine.
- `not_null` on `stg_airbyte_source__gp_api_db_race_opponent_standout_action.opponent_name` fails on a small, growing set of rows. The column is nullable upstream by design (actions not tied to a named opponent); confirmed by the data team on the source schema.

## What

- `assert_er_clusters_share_election_date` now fails only when a cluster's election dates spread wider than the 10-day window (the matcha config is the contract's source of truth). Stage labels stay uncompared: sources label the same stage inconsistently (5k+ clusters carry mixed labels today), so the date window is the stage-identity contract.
- Updated the stale "clusters are single-date" phrasing where it was documented: the minted-candidacy-ids model header, its yaml description, and the CLAUDE.md id-derivation section.
- `config: freshness: null` on the two empty race_opponent sources, matching the existing treatment of the sporadically-written annotation tables, with a comment to revisit once the product writes rows.
- Dropped the `not_null` on `standout_action.opponent_name` and documented the nullable contract. The summary table's `opponent_name` stays tested: it is required upstream.
- matcha README: the Databricks delivery examples now target dated `er_source` tables with the real table names dbt reads, plus a note on why `--overwrite` must never point at the live tables (`CREATE OR REPLACE` + `COPY INTO` is not atomic).

## Verification

- New test SQL run against prod `er_source.clustered_candidacy_stages`: 0 rows (all 155 previously failing clusters sit within the window; max spread 9 days; none exceed it).
- `assert_multi_br_clusters_never_adopt_cluster_mint` re-run in prod context: 0 failures (its CI-only failures on open PRs are preview-vs-prod deferral mixes, not a prod issue).
- Remote `dbt parse` via the Cloud CLI: clean (only the two pre-existing OSI metric warnings).
- `pre-commit run` on all changed files: green.

## Sequencing note for CI on this PR

A separate data fix is in flight: the Jul 9 cluster snapshot references 111 BallotReady candidate ids retired in Sunday's BR sync, which fails `not_null_int__civics_person_canonical_ids_first_seen_at` (this is what broke the 4-hourly job). Until the cluster refresh (matcha rerun) is swapped in, this PR's CI will hit that failure when it rebuilds the person spine, and cross-era asserts can false-fail on preview-vs-prod mixes until prod completes one green build. Retrigger CI after the cluster swap lands.

---

> [!NOTE]
> **Low Risk**
> Changes are test contracts, source freshness overrides, and documentation; no transformation logic or minting SQL behavior changes. The relaxed cluster test is intentional alignment with existing matcha matching rules.
> 
> **Overview**
> Unblocks nightly `dbt build` and source freshness by aligning tests and docs with **matcha**’s candidacy-stage contract instead of stricter assumptions that no longer hold.
> 
> **`assert_er_clusters_share_election_date`** now fails only when a cluster’s `election_date` spread exceeds **10 days** (matching `ELECTION_DATE_WINDOW_DAYS` in matcha), and reports `spread_days`. Stage labels are intentionally not compared. Comments clarify that clusters are **single-stage** (dates may differ within the window), not single exact date.
> 
> Documentation in `CLAUDE.md`, `int__civics_minted_candidacy_ids` SQL header, and `int__civics.yaml` is updated to the same **single-stage / 10-day window** wording.
> 
> **gp_api Airbyte contracts:** `freshness: null` on empty `race_opponent_contrast` and `race_opponent_research` sources; **`not_null` removed** from `race_opponent_standout_action.opponent_name` with nullable upstream documented.
> 
> **matcha README:** CLI examples target dated `er_source` tables; adds guidance to validate then swap into live tables and avoid `--overwrite` on live tables during upload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c5c12eb1a003abdd558f41f76f46e357c600ead. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/CLAUDE.md
+++ b/dbt/project/CLAUDE.md
@@ -265,9 +265,10 @@ cluster mints from itself (single-member semantics).
     from the person group's earliest member (person grain). `gp_candidate_id` is
     kept as an alias of `gp_person_id`.
 *   `gp_candidacy_id` — minted in `int__civics_minted_candidacy_ids` from the
-    candidacy-stage cluster's earliest member. Clusters are single-date (stage
-    grain), so the candidacy-grain id is the min minted id over the candidacy's
-    stages; co-clustered candidacies converge because they share clusters.
+    candidacy-stage cluster's earliest member. Clusters are single-stage
+    (election dates may differ within matcha's 10-day window), so the
+    candidacy-grain id is the min minted id over the candidacy's stages;
+    co-clustered candidacies converge because they share clusters.
 *   `gp_election_stage_id` — minted in `int__civics_minted_election_stage_ids`
     from the election-stage cluster's earliest member (already stage grain).
 
--- a/dbt/project/dbt_project.yml
+++ b/dbt/project/dbt_project.yml
@@ -21,6 +21,11 @@ clean-targets: # directories to be removed by `dbt clean`
   - "target"
   - "dbt_packages"
 
+vars:
+  # Must match ELECTION_DATE_WINDOW_DAYS in matcha/scripts/configs/candidacy.py
+  # (the matcher's contract source of truth). Update both together.
+  er_election_date_window_days: 10
+
 # Configuring models
 # Full documentation: https://docs.getdbt.com/docs/configuring-models
 
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -3399,10 +3399,11 @@ models:
       Minted gp_candidacy_id per clustered candidacy-stage record. One row per
       candidacy-stage unique_id. The id is the salted-hash of the cluster
       member earliest by first_seen_at (ties: source_name, source_id), shared by
-      all members, so it is stable under cluster churn. Clusters are single-date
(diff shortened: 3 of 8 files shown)


sanjayr1:
delegate review

sanjayr1:
delegate review

PR #597, opened 2026-07-13
[DATA-2093] Migrate hand-written HubSpot contact columns into the property registry

sanjayr1:
## What

Completes the DATA-2091 registry end-state the DE team asked for: all 118
property-derived hand-written columns of stg_airbyte_source__hubspot_api_contacts
move into the hubspot_contact_property_columns seed registry. The model body is
now the four Airbyte record columns (id, companies, created_at, updated_at) plus
the generation loop, and no column reads Airbyte-flattened properties_* columns.
Zero renames, zero type changes: every output keeps its exact name and current
warehouse type (column-set/type identity vs prod verified: zero differences
across all 353 columns).

## How

- Seed regenerated: 349 rows, all status=generated (46 former existing rows
  flipped with their metadata kept; 72 rows added). One row per OUTPUT column:
  dual-output properties (pledge_status -> is_pledged + pledge_status; the
  *_raw pairs) have two rows sharing internal_name, so the internal_name unique
  test is dropped; column_name unique remains the output contract.
- Cast vocabulary extended for exact type preservation: int (hand-written int
  casts stay int, not bigint), decimal(38,10) (the *_raw columns mirror the
  Airbyte-typed flattened source columns), int_via_float (preserves
  number_of_seats_available's float round-trip so decimal-formatted strings
  still land as ints).
- Custom word-to-boolean maps live in the seed as data: boolean_true_value /
  boolean_false_value (partisan/nonpartisan, uncontested/contested), passed to
  cast_to_boolean by the expression macro.
- Legacy plain casts become the registry's try_cast(nullif(..., '')) semantics:
  generated columns can never raise at query time inside Sigma (staging and
  mart are views); the blocking lossless test turns silent cast loss into a
  build failure instead.
- New cast_loss_expected seed flag on five columns whose source properties
  legitimately carry non-conforming values the cast nulls BY DESIGN, unchanged
  from the hand-written expressions (br_race_id and br_position_id hold
  free-text workflow markers, is_confirmed_candidate and has_verbal_pro carry
  out-of-map enum words, goodparty_user_id tolerates free text). The lossless
  test skips exactly those rows and strictly zero-loss-tests every other typed
  column (~137).
- assert_hubspot_generated_casts_lossless: renders a valid always-empty
  statement when the registry has no typed rows (review follow-up from #578),
  and declares both refs via depends_on hints since they now sit inside a
  conditional block.
- YAML: every output column is now documented (38 entries added).

## Verification

- Pre-change probes over the full raw table proved the JSON extractions
  reproduce every migrated column exactly where representation could differ
  (Airbyte-typed date and decimal columns, int paths, custom boolean maps):
  zero mismatches.
- Dev-environment build green end to end (43/43 including the expanded
  lossless test), exhaustive render scan of all 349 generated expressions over
  the full table, and column name/type identity vs prod: zero differences.
- Full-view value parity, prod vs rebuilt view, all 353 columns, full-table
  xxhash64 checksums with both sides computed in one statement: 352/353 exact.
  The single difference is is_incumbent on ONE raw row whose Airbyte-flattened
  cell is NULL while the same row's authoritative properties JSON carries
  "Yes" — prod yields null, the migrated view correctly yields true. That
  flattened-cell desync is precisely the failure mode this migration
  eliminates, so the one-row improvement is accepted and documented.
  (Preview-schema parity rerun to follow once CI is green.)

## Notes

- Column ORDER is not part of this view's contract (consumers select by name;
  the mart is a select * passthrough). This PR does reorder: the four record
  columns move to the front, so the migrated property block shifts by two
  positions and created_at/updated_at move from ~120/121 to 3/4. Names, types,
  and values are unchanged (one row of one column excepted, above).
- The seed gains three columns and its config now sets full_refresh: true
  (drop-and-recreate per run, a few hundred rows), so seed schema changes
  deploy themselves in CI and prod; the dev build proved this by recreating
  the seed through the schema change cleanly.
- Airbyte connection cleanup (dropping the now-unread flattened properties_*
  columns) is deliberately out of scope; separate ticket with its own
  blast-radius check.

---

> [!NOTE]
> **Medium Risk**
> Large refactor of a heavily downstream HubSpot contacts staging view (353 columns) with broad consumer surface; parity is guarded by lossless tests and documented one-row `is_incumbent` improvement from JSON vs stale flattened cells.
> 
> **Overview**
> Moves **118 hand-written HubSpot contact columns** off `stg_airbyte_source__hubspot_api_contacts` and into the **`hubspot_contact_property_columns` seed**, so the staging model is only Airbyte record fields (`id`, `companies`, `created_at`, `updated_at`) plus a Jinja loop over the registry. Extractions now read **`properties` JSON** via `get_json_object` instead of Airbyte-flattened `properties_*` columns.
> 
> The seed is rebuilt as **349 `generated` rows** (one row per output column, including dual outputs like `is_pledged` / `pledge_status` and `*_raw` pairs). **`hubspot_contact_property_expression`** gains cast types **`int`**, **`int_via_float`**, **`decimal(38,10)`**, and seed-driven **`boolean_true_value` / `boolean_false_value`** for custom `cast_to_boolean` maps. **`cast_loss_expected`** marks columns where intentional cast nulling is unchanged from legacy SQL; **`assert_hubspot_generated_casts_lossless`** skips those rows and handles an empty typed set safely. Seed config adds **`full_refresh: true`** and schema tests/docs for the new metadata columns; staging YAML gains **38 column descriptions**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 907172aa3783c3618b72d5b90775f4255cd6c12a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sanjayr1:
## Preview verification (CI run green)

All checks against `dbt_cloud_pr_70471823431465_597` vs prod:

- **Full-view value parity**: all 353 output columns compared with full-table xxhash64 checksums (id + value, both sides computed in ONE statement per chunk so a single Delta snapshot serves both views). **352/353 exact; row counts identical.** The single difference is `is_incumbent` on one raw row whose Airbyte-flattened cell is NULL while the same row's authoritative `properties` JSON carries "Yes" — prod yields null, this PR correctly yields true. That flattened-cell desync is the failure mode the migration eliminates; accepted and documented in the PR body.
- **Schema identity**: column set/type full outer join staging PR-vs-prod = 353 compared, 0 missing, 0 extra, **0 type differences**; 120 ordinal moves, exactly the documented non-contractual reorder.
- **Row grain**: identical (same count, unique non-null ids, no duplicates, both sides).
- **High-risk column row-level audit** (email, br ids, goodparty_user_id, booleans, population pair, seats, contact_created_at): **0 row-level differences** except the accepted is_incumbent row.
- **cast_loss_expected audit**: the five exempted columns' cast loss equals the pre-existing hand-written behavior, and their prod-delta is **0 on every row** — the exemption changes visibility (loss is now a declared registry fact), not behavior. The non-conforming source values are workflow markers, literal "null" strings, multi-state enum words, and one test slug.
- **Mart inheritance**: preview `hubspot_contacts` vs preview staging, name/type full outer join = **0 differences** (the select * mart rebuilt and matches exactly).
- **Lossless test**: PASS in CI at its expanded coverage (~137 typed non-exempt columns, zero loss).

The parity harness is rerunnable against prod anytime as an audit.

sanjayr1:
## Adversarial review record (independent reviewer, two passes)

**Pass 1** (diff only): preliminary REQUEST-CHANGES, three majors + one minor. It also proposed four verification queries, which were run against the preview schema (one needed a mechanical alias fix to run at all — its CTE alias collided with the real `goodparty_user_id_raw` column).

**Pass 2** (with the query results as context): **FINAL APPROVE-WITH-NITS, "must-fix items: none before merge."** Its own summary: Q1 proves column name/type identity across all 353 columns; Q2 proves row-grain identity; Q3 proves row-level parity for the high-risk migrated columns with only the documented one-row is_incumbent improvement; Q4 shows the cast-loss exemptions preserve existing prod behavior rather than introducing new loss.

Dispositions of the pass-1 majors:

1. *"The committed lossless test doesn't prove value parity"* — the test's job is future drift, not one-time parity; parity is proven by the rerunnable checksum harness (352/353 exact + one documented row) plus the row-level audits above. A committed parity-vs-prod test cannot exist post-merge (prod becomes the migrated view).
2. *"cast_loss_expected removes guardrails from join keys"* — those columns had no loss guard before this PR (hand-written, outside the test), and Q4 shows byte-identical values to prod on every row. The exemption converts invisible loss into a declared registry fact.
3. *"Seed schema change isn't self-deploying for model-only runs"* — the real deployment paths (CI and the merge-triggered prod job) run state:modified+, which selects the modified seed and orders it before the model via the depends_on edge; the dev gate proved the full_refresh config recreates the seed through the schema change. An ad-hoc model-only run against a stale seed fails loudly, which is safer than a tolerant accessor silently applying default boolean maps.

Accepted follow-up (not in this PR): lightweight DQ monitoring of cast-loss rates on the five exempt columns (raw non-empty vs cast survivors), so growth in non-conforming values stays visible without a blocking test hard-coding a drifting threshold. Candidate to bundle with the Airbyte flattened-column cleanup ticket.

PR #604, opened 2026-07-14
[DATA-2118] Person/OfficeHolder marts + full election-api sync via Airflow atomic swap

danpelota:
## Summary

Populates the election-api **Person spine** added by [omni #696](https://github.com/thegoodparty/omni/pull/696) and migrates the **entire** election-api Postgres sync onto the Airflow build-and-swap DAG, retiring the legacy dbt python writer.

## dbt marts (`models/marts/election_api/`)

- **`m_election_api__person`** (new) — full canonical coverage: one row per `gp_person_id` referenced by a synced candidacy or officeholder, so `Candidacy.person_id` / `OfficeHolder.person_id` always resolve. The civics `people` mart is the coverage backbone (name/email/phone/state for every person); `int__ballotready_person` supplies BallotReady-granularity fields (bio, headshot, typed social links, `degrees`/`experiences` JSON).
- **`m_election_api__office_holder`** (new) — from civics `elected_official_terms` (term grain); required `person_id`, nullable `position_id` resolved via the Position mart.
- **`m_election_api__candidacy`** — emits `gp_candidate_id as person_id` (the new typed FK; the legacy `gp_candidate_id` text column is retained).
- **`m_election_api__position`** — adds `salary` (BallotReady `api_position.salary`).
- YAML docs + tests for the new marts and columns.

## Airflow — one FK-safe atomic swap for all tables

- **`election_api_utils`**: `swap_graph_into_target` / `drop_old_graph` swap the whole table graph in ONE transaction, in dependency order (Person and other parents first), with a children-first `_old` drop. FKs on each `_new` table reference the **sibling `staging."<Parent>_new"`** (not the live parent) — Postgres FKs follow the referenced table's OID through `RENAME`, so a swapped parent would otherwise dangle its children's FKs.
- **`sync_election_api`**: rewritten to a `TablePipeline` + task-group factory over all 13 tables. Every table's indexes build before any FK (barrier), so FK validation is the hard referential gate; then a single `swap_all` and `drop_all_old`.
- Removes `write__election_api_db.py` (and its test / `write.yaml` entry).

## Verification

- `dbt build` of the changed marts in dev: `person`/`office_holder`/`candidacy` relationship + uniqueness/not-null tests pass.
- Full Airflow test suite: **97 passed** (new graph-swap, self-heal, ordering, FK-sibling and topological-order guards, DagBag structural test).
- `pre-commit` clean.

## Before production cutover

- The DAG's first run needs the omni Prisma migration deployed on the target (already on dev).
- Run a one-time `dbt build --full-refresh --select m_election_api__position` to backfill `salary` onto existing incremental rows.

Also included: an earlier commit backfilling people names from the BR officeholder feed (drops the `people` null-name rate ~42% → ~9%), which feeds the new Person mart.


danpelota:
Closing in favor of the full election-api refresh

PR #606, opened 2026-07-14
[DATA-2101] Surface Proposed_District as a resolvable district type

hhkarimi:
## Why

Candidates in mid-decade-redistricted states (reported for OH-4, also affects UT and others) see voter data on the old congressional map. The new 2026 maps are already ingested from the vendor, in the `Proposed_District` column (values like `2026 PROPOSED CONG DIST 04 (EST.)`), but that column was never surfaced as a district type, so the election-api District dimension had no rows for it and nothing could resolve to it.

## What

Add `Proposed_District` to the `'all'` scope of `get_l2_district_types` (`macros/l2_district_columns.sql`). Its UNPIVOT then mints District rows in `m_election_api__district` keyed `(state, 'Proposed_District', '<year> PROPOSED CONG DIST NN (EST.)')`, with voter counts from `int__l2_district_aggregations` (same macro).

Deliberately **not** added to `scope='allocated'`. That scope drives `int__l2_block_district_map` and `int__serve_block_coverage`, the current-map substrate Serve resolves against, plus the census population allocation. Serve keeps the current map, and neither `accepted_values` test on `district_type` changes (both read `'allocated'`).

### The adoption gate

A proposed value in the vendor feed is **not** evidence the map was adopted. It can be rejected, superseded, or struck down, and the vendor lags on removing dead maps. Virginia is the standing case: its remap was struck down in May 2026 and L2 still ships a proposal that moves 49.7% of the state's voters. So "use proposed where it exists" is wrong, and the DATA-2186 TDD rejects it.

`seeds/district_map_adoption.csv` records the decision per (state, district type) as `current` / `proposed` / `needs_boundary`, each with a citation, sourced from the TDD's Appendix A.1. `m_election_api__district` mints a `Proposed_District` row only where the state is `proposed` **and** `is_verified`, so:

| Outcome | States |
| --- | --- |
| Bindable now | OH, NC, UT |
| Adopted but not yet checked, so held back | TX, CA, FL, TN, MO |
| Current map is correct | VA |
| Neither L2 column is the adopted map | LA, AL |

Non-congressional values in the same column (MI's proposed state senate, the WA/CO annexations) are dropped rather than minted as inert rows.

`district_number` is a sparse key, the same convention `l2_manual_district_assignments` uses: empty means the decision covers the whole state, a value narrows it to one district. Adoption is a state-wide legal fact so empty is the norm, and every row is state-level today. It exists because verification confidence is per-district: a district can be rolled out as it is checked against the enacted map without holding the rest of the state back, and a district left out behaves exactly as it does today rather than getting worse.

Absence of a seed row means `current`, which is conservative but not automatically right, since `current` is the *stale* map for a state that redistricted. `assert_proposed_cong_dist_seed_coverage` therefore fails if a state carrying a proposed congressional value has no explicit decision, and `assert_election_api_district_proposed_gated` fails if anything ungated reaches the dimension by any path, including the turnout union that the filter itself does not cover.

Also in this PR:

- A format guard on the proposed congressional names the district resolver parses, so a vendor label change fails loudly instead of silently. It runs against `m_election_api__district` (one row per state/type/name) rather than the voter-grain L2 view: same coverage for a fraction of the scan.
- Bumps the macro unit test's `all_list_size` assertion from 230 to 231.

## Rebased onto the merged macro

This branch previously edited `macros/get_l2_district_columns.sql`, which #812 deleted when it merged the two district-column lists into one `scope`-parameterized macro. The branch has been rebased and the change re-landed against `macros/l2_district_columns.sql`. The old "don't touch `get_l2_major_district_columns`" intent maps cleanly onto "add to `'all'`, not `'allocated'`".

## CI blast radius

Per that macro's own docstring, editing either scope's list marks the **other** scope's downstream models `state:modified`. So `state:modified` CI will select `int__l2_block_district_map` and `int__serve_block_coverage` even though their logic and their column list are unchanged. Expect a heavy run; it is not a Serve behavior change.

## Non-destructive / scope

- Adds rows, not output columns, so the incremental marts' `on_schema_change="fail"` stays safe. Current `US_Congressional_District` rows are unchanged.
- `int__l2_nationwide_uniform` now applies its `ltrim('0', ...)` padding strip to `Proposed_District`. That is a no-op on these values (they start with the year), so the string the resolver parses is unchanged. The column already flows to the product Voter table and the downstream selects name it explicitly, so the view's column reordering is inert.
- `Proposed_District` also carries non-congressional proposed values in a few states (e.g. proposed state-senate in MI, local annexation areas in CO/WA); those become inert District rows that nothing binds to.
- `m_election_api__district_top_issues` inner-joins a static match model, so proposed rows mint no top-issue rows there. See the follow-up note below.
- **Only adopted maps become bindable**, which is what makes this safe to deploy ahead of the routing. `GET /elections/districts/types` and `/districts/names` are `listDistinct` over the District table, so anything minted here is immediately selectable in the override picker. The seed gate is what keeps VA's struck-down map out of that list.

## Verified on the rebase

The `state:modified+ --full-refresh` CI run built `int__l2_district_aggregations` and `m_election_api__district` into the PR schema, and every test on them passed (`assert_proposed_cong_dist_format` included). Querying that schema directly confirms the rows are actually minted, not just format-clean:

- **11 states get proposed congressional rows**, each with the right number of districts: CA 52, TX 38, FL 28, OH 15, NC 14, VA 11, TN 9, MO 8, AL 7, LA 6, UT 4.
- **OH has all 15 CDs.** CD-4 = 506,552 registered voters, against the vendor portal's 504,195 read in July. The +0.5% shift is consistent with the 2026-07-29 OH reload landing between the two reads.
- **Non-congressional proposed values stay inert**, and are only what we expect: MI 38 proposed state senate, CO 4 annexation areas, WA 7 hospital-district potential-annex areas. Nothing binds to them.

Note for whoever writes the resolver: MI's values are `2026 PROPOSED STATE SEN DIST 01`, the same `<year> PROPOSED ... DIST <n>` shape as the congressional ones. Matching must require the `CONG DIST` token, not just `l2_district_type = 'Proposed_District'` plus a year and a number.

Also worth knowing: the format guard passes vacuously if no `Proposed_District` rows exist, so it protects the format but is not by itself evidence the rows are there.

## Known CI red: `assert_candidacy_hubspot_company_ids_populated`

Unrelated to this branch, and not a regression. Evidence:

- The **same** test passed on the previous commit of this branch, whose only difference is a `230` to `231` constant in `tests/test_l2_district_columns_macro.sql`.
- It passes against production data (`dbt test --select assert_candidacy_hubspot_company_ids_populated`, PASS).
- Nothing here touches candidacy, gp_api, or HubSpot. This branch only added a district column name to a macro.

Diagnosed against the PR schema rather than guessed. Exactly one row fails, campaign `326446`:

| Relation | `hubspot_company_ids` |
| --- | --- |
| PR schema `int__civics_candidacy_gp_api` | `["57496181344"]` |
| PR schema `candidacy` | null |
| production `candidacy` | no row for this campaign at all |

It is a recently created campaign. Production passes because the row does not exist there yet; CI's full refresh materializes it and the mart drops the company id. That also explains the timing: the first run of this branch passed at 17:53, and the company id landed in the source before the 20:47 run.

`candidacy.sql:115` coalesces `gp_api.hubspot_company_ids` with `br.hubspot_company_ids`, so for this row the gp_api join produced nothing even though `candidacy.product_campaign_id` is `326446` — the candidacy row reaches its `product_campaign_id` by a path other than the join key. That is a defect in the candidacy/HubSpot lineage (the DATA-2196 / DATA-2197 area), it will recur on every full-refresh CI run until fixed, and it is unrelated to this branch, which only adds a district column name to a macro. It needs its own ticket.

## Deploy (required, will otherwise no-op)

The marts are incremental and only rescan L2 rows where `loaded_at >= max(updated_at)`. Whether a normal incremental run picks up the proposed rows depends on the current mart watermark against the latest L2 load (OH, AZ and UT reloaded 2026-07-29). Check the watermark first; if the proposed-CD rows predate it, run once after merge:

```
dbt build --select "int__l2_district_aggregations m_election_api__district m_election_api__district_top_issues" --full-refresh
```

then the `sync_election_api` Airflow DAG upserts into election-api Postgres. (The old `write__election_api_db` dbt model this note used to name was removed in #816; the DAG replaced it, syncing `m_election_api__district` into `District` on the `(state, l2_district_type, l2_district_name)` unique key.) Propagation is unblocked: the upsert is generic and `L2DistrictType` is an unconstrained string in Postgres/Prisma and a `z.string()` in election-api Zod (no enum/allowlist).

## Follow-ups (tracked on the epic)

- **Per-state gate before any routing binds.** This PR mints bindable-looking rows for every state that has proposed values, including VA, whose remap was struck down in May 2026 while the vendor still ships the map. The gate has to land before omni routing, not after.
- **District top issues.** Because `m_election_api__district_top_issues` gates on a static match model, a campaign routed to a proposed CD would get zero top-issue rows where it has them today. Needs a rematch or an explicit fallback to the current district before the routing switch.
- Product routing (bind Win congressional campaigns to the proposed CD; Serve stays on current) lands as separate `omni` PRs.

## Addition: score adopted proposed districts in the top-issues mart

`m_election_api__district_top_issues` builds its `target_districts` CTE from `stg_model_predictions__llm_l2_br_match_20260126`, a static snapshot that predates the proposed-district ingest, and inner-joins on it. Proposed districts therefore score zero issue rows.

That matters now because omni #1257 routes the onboarding voter-issues endpoint to the proposed district for adopted states. Without this change, that endpoint returns nothing for NC, OH and UT — the three states the gate actually admits.

The fix unions the adopted proposed districts in from `m_election_api__district`. Because that model is already gated by `district_map_adoption`, the union **inherits** the gate rather than restating it: only `adopted_source='proposed' AND is_verified` districts have rows there to union in. No adoption logic is duplicated.

`assert_proposed_districts_have_top_issues` fails if any `Proposed_District` row lacks top-issue rows. It is not vacuous by construction: remove the union and `target_districts` holds no proposed rows, the inner join in `district_avg_scores` yields nothing for them, and the assertion returns every proposed district as unmatched.

### Verification

`dbt build --select m_election_api__district_top_issues` — PASS 26/26, the new assertion at zero failing rows. OH shows 15 proposed districts carrying issue rows, matching its adopted map.

Confirmed while doing this that the macro's default scope already includes `Proposed_District`, so the existing unqualified `get_l2_district_columns` call needed no change.

### One thing worth a follow-up ticket

`m_election_api__district`'s `turnout_districts` CTE, sourced from `int__model_prediction_voter_turnout`, applies no adoption filtering at all. If that model ever emits `Proposed_District` rows, they would reach the mart ungated and undermine the inherit-the-gate premise this change relies on.

It does not today — querying the mart returns proposed rows for exactly NC (14), OH (15) and UT (4), the three seed-adopted states. But nothing structurally prevents it, and a guard there would be cheap.


hhkarimi:
`delegate review`

Fixed in a96bbdf. Adopted point 1 and adjusted point 2:

**Point 1 (case) — applied.** The format check now runs `regexp_like(upper(proposed_district), ...)`, matching the `upper()` LIKE filter on the line above and the case-insensitive routing regex it guards (`/^(\d{4}) PROPOSED CONG DIST (\d+)/i`). Test passes with 0 drift rows.

**Point 2 (end-anchor) — not adopted, with reason.** Verified against all 7,567,704 OH proposed-CD rows: `'^[0-9]{4} PROPOSED CONG DIST [0-9]+$'` matches **0** of them, because every value carries a trailing ` (EST.)` suffix. Adding `$` would flag 100% of valid rows and fail the test on all current data. The missing end-anchor is intentional: this guard mirrors the routing regex, which is a deliberate prefix match tolerating the `(EST.)` suffix. Anchoring would make the guard stricter than the parser it protects. If we later want to reject genuinely malformed trailing content, that tightening belongs to the routing regex and this guard together, and must still allow the optional `(...)` suffix rather than a bare `$`.

hhkarimi:
delegate review

hhkarimi:
All three findings were right. Two of the suggested fixes were not, so I've fixed the underlying issue rather than applying the patch — detail below.

**Format guard, missing end anchor — valid, fix adjusted.** The concern is correct: without an anchor, `... DIST 5 INVALID EXTRA` satisfies the pattern and reaches district resolution unflagged.

The suggested `'^[0-9]{4} PROPOSED CONG DIST [0-9]+$'` would have broken the build. Tested against the column: it passes **7,958,045 of 101,966,221** rows. Only MI's state senate values arrive bare; every congressional value ends ` (EST.)`, so that anchor fires on all 94,008,176 of them. Anchored with the suffix spelled out as optional instead, and verified all 101,966,221 pass.

The same finding also flagged a case mismatch between the two predicates. That was already resolved when this guard was generalised to cover both district types — both sides go through `upper()` now, which is why the file is `assert_proposed_district_format` rather than the `_cong_dist_` path in the comment.

One deliberate asymmetry, now stated in the file: the parse macros are left **unanchored** while this guard is strict. A vendor formatting change therefore keeps routing working while still being reported, rather than silently mis-parsing.

**Drift guard, whitespace — valid, applied as suggested.** Correct on the mechanism and cheap to fix. Nothing trims `Proposed_District` upstream, so a trailing space on an already-classified value would miss the ignored-values join and fail the build on a false alarm rather than real drift. Both sides normalised.

**`is_verified` invariant — the strongest of the three, fix relocated.** The column's own description warned that a stale `true` would skip the human check on the next `adopted_source` edit, and nothing enforced it. That gap was real.

The suggested placement does not compile: a column-level `dbt_utils.expression_is_true` prepends the column name, so it generated `is_verified adopted_source = 'proposed' or not is_verified` and failed with `PARSE_SYNTAX_ERROR ... near '='`. Moved to the seed-level `data_tests` block, matching the eight existing usages in that file, where a cross-column expression stands alone.

Mutation-tested rather than trusted green: flipping VA to `is_verified = true` while it is still `current` makes it fail with exactly that row.

`1b9ee3f8`. Local build of the seed and both guards: 15 pass, 0 errors.

delegate review

hhkarimi:
delegate review

PR #609, opened 2026-07-15
[DATA-2069] Expose amplitude catalog + taxonomy through mart_analytics

hhkarimi:
## What

Exposes the two remaining Amplitude relations the governance loop reads through `mart_analytics`, so every reader hits the mart and access can be granted at the mart schema instead of on individual dbt relations.

- `amplitude_event_catalog` — view over `int__amplitude_event_catalog`
- `amplitude_taxonomy_event_type` — view over `stg_airbyte_source__amplitude_taxonomy_event_type`

The event stream is already exposed via the existing `amplitude_events` view, so no third model is needed.

## Why

The analytics governance loop (the event-health monitor and the event-provenance backfill) reads three relations: the event catalog, the event stream, and the Govern taxonomy. Its service principal had been granted `SELECT` on each individual relation. Per-relation grants break when a model is rebuilt or reconfigured, and they blur the line between what Terraform owns (access) and what dbt owns (models).

Routing the reads through the mart lets the service principal join the existing `mart_analytics` reader group and inherit `SELECT` on the whole schema, so nothing is granted per relation and rebuilds cannot break access. Both models are thin views (no storage duplication), matching the existing `amplitude_events` and `hubspot_contacts` mart exposures.

## Validation

`dbt build --select "amplitude_event_catalog amplitude_taxonomy_event_type"`: both views created, 3 tests passed (not_null on `event_type` for both; unique on the catalog). The taxonomy model omits a `unique` test because the upstream staging table is an Airbyte passthrough whose event_type-grain uniqueness is not guaranteed, and the backfill already selects distinct.

## Pairs with

- omni: repoints the monitor and backfill at `mart_analytics.*`.
- gp-terraform-dataplatform: adds the service principal to the `mart_analytics` reader group and drops the per-relation grants.

### DIFF
--- a/dbt/project/models/marts/analytics/amplitude_event_catalog.sql
+++ b/dbt/project/models/marts/analytics/amplitude_event_catalog.sql
@@ -1,7 +1,4 @@
 {{ config(materialized="view") }}
 
--- Exposure of int__amplitude_event_catalog into mart_analytics so governance
--- consumers read the mart (grants live at the mart schema, not per relation).
--- Thin view over an already-materialized table; no storage duplication.
 select *
 from {{ ref("int__amplitude_event_catalog") }}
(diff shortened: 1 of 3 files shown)


hhkarimi:
delegate review

PR #617, opened 2026-07-15
Cover all uv projects in Dependabot config

hhkarimi:
## What

Dependabot version-updates were only configured for `/dbt` and `/airflow/astro`, plus a root `directory: "/"` entry. Since the uv migration there is no Python manifest at the repo root, and `directory` is not recursive, so that root entry scanned nothing.

This replaces the dead root entry and adds the remaining uv projects, so every directory that ships a Python manifest (`pyproject.toml`/`uv.lock`, or `requirements.txt` for the Astro image) is tracked.

Directories now covered:
- `/dbt` (existing)
- `/airflow` (new)
- `/airflow/astro` (existing)
- `/analytics` (new)
- `/matcha` (new)
- `/people-api-loader` (new)
- `/apps/genie-tools` (new)
- `/apps/genie-slack-bot` (new)

Each entry keeps the existing settings unchanged (weekly schedule, `increase-if-necessary`, cooldown, minor/patch and major groups, 2 open-PR limit per directory). Kept as explicit per-directory blocks rather than YAML anchors or a `directories` list, since Dependabot's parser rejects unknown top-level keys and a shared PR limit would starve most directories.

## Why

The other uv projects currently get no automated dependency-update PRs. This is also related to a batch of stale security alerts that were just dismissed: they pointed at `dbt/poetry.lock` and `airflow/poetry.lock`, which were removed in the uv migration, while the current `uv.lock` files already carry the patched versions.

## Testing

- `python3 -c "import yaml; yaml.safe_load(open('.github/dependabot.yml'))"` parses cleanly (8 update entries, all `pip`).
- pre-commit `check yaml` hook passes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,7 @@
 version: 2
 updates:
-  - package-ecosystem: "pip"
+  - package-ecosystem: "uv"
     directory: "/dbt"
-    versioning-strategy: increase-if-necessary
     schedule:
       interval: "weekly"
       day: "sunday"
@@ -29,9 +28,8 @@ updates:
         update-types:
           - "major"
 
-  - package-ecosystem: "pip"
+  - package-ecosystem: "uv"
     directory: "/airflow"
-    versioning-strategy: increase-if-necessary
     schedule:
       interval: "weekly"
       day: "sunday"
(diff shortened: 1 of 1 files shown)


hhkarimi:
Good catch — fixed in c189842.

Switched the seven uv-managed directories (`/dbt`, `/airflow`, `/analytics`, `/matcha`, `/people-api-loader`, `/apps/genie-tools`, `/apps/genie-slack-bot`) to `package-ecosystem: "uv"` and dropped `versioning-strategy` from those entries. `/airflow/astro` stays on `pip` since it ships `requirements.txt` only.

This also explains the failing `dbt/pyproject.toml` update job visible in the dependency graph — it was being scanned under `pip`, which can't handle a uv project.

delegate review

PR #632, opened 2026-07-16
people-api-loader: seed the CRM name-search indexes (trigram + name b-trees) and install pg_trgm

tomer-tgp:
## Why

The CRM typeahead (omni ENG-10683) switched voter name search to trigram-backed substring matching (omni #794/#796). Those indexes were added as a people-api prisma migration, but prisma migrations do not run against loader-built clusters, and any index missing from the committed seed vanishes at the next rebuild/cutover. This carries the five name-search indexes in the loader seed and installs `pg_trgm` so `build_indexes` can re-issue the `gin_trgm_ops` entries.

Three of the five are not new to the CRM work: `Voter_last_first_id_idx` (backs the list `ORDER BY "LastName","FirstName",id`) and the two `lower()` b-trees (back short-token prefix search) came from the Dec-2025 omni migration `20251208120000` and were never in the seed. The serving clusters were running name search unindexed until 2026-07-16, when all five were built directly on the live dev and prod clusters (staged per-partition `CONCURRENTLY` + `ATTACH`, all parents `indisvalid=true`, `ANALYZE` run). This PR makes that state survive the next rebuild.

## How

The hand-added indexes live in a dedicated module, `_serving_seed_extra.py`, and merge into the composed spec via `schema_spec.indexes_for`. This is deliberate: `extract-serving-structure` regenerates `_serving_seed.py` wholesale, so any entry placed directly in the generated seed would be silently dropped on the next extraction. Keeping them in `_serving_seed_extra.py` means regeneration cannot lose them, and a name collision resolves in favor of the generated entry.

`pg_trgm` is installed in both `create_schema` and `build_indexes`. It is a trusted extension (no superuser needed on RDS). Each step guards on its own manifest and must be independently re-runnable, so `build_indexes` installs it too rather than assuming `create_schema` ran first. Seed SQL deliberately has no `CONCURRENTLY` (invalid on a partitioned parent; `build_indexes` handles per-partition builds) and no `IF NOT EXISTS` (`_ensure_idempotent` adds it).

## Expression spelling and opclass

people-api emits `lower(v."FirstName") LIKE ...` (see omni `buildVoterWhereSql.utils.ts`), and the planner only matches an expression index whose definition is identical, so the seed entries use `lower("FirstName")` / `lower("LastName")` exactly.

The two `lower()` b-trees carry the `text_pattern_ops` opclass, which is a deliberate divergence from the current people-api prisma index (plain opclass). A default-opclass b-tree only serves `lower(col) LIKE 'tok%'` prefix search in a C-locale database, and the serving cluster is `en_US.UTF-8` (confirmed `datcollate=en_US.UTF-8`). Without `text_pattern_ops` those b-trees would be dead weight and the anchored-prefix name search would fall back to the trigram GIN. The loader-built `public` serving replica is the source of truth the upcoming feature reads, and the prisma controls are slated for deprecation, so the seed is corrected here rather than mirrored from a migration that will be retired. The green OLTP side is intentionally left alone. The composite `Voter_last_first_id_idx` keeps the default opclass because it backs `ORDER BY "LastName","FirstName"` under the DB collation, which `text_pattern_ops` would not serve.

## Merge with main

This branch merges `main`, which landed the District family loader work (DATA-2100, #607). That refactor touched the same three files, so the merge resolves conflicts in `create_schema.py` (the `pg_trgm` install now sits inside main's generic per-table DDL loop), `build_indexes.py` (the one-time `pg_trgm` install sits before main's per-table index loop), and `tests/test_schema_spec.py` (both test suites kept).

## Tests

`test_create_schema` and `test_build_indexes` assert the `pg_trgm` install; `test_schema_spec` covers the extras merge, the absent-from-generated invariant, and collision suppression. A new integration test (`test_integration_pg.py`, gated behind `LOADER_INTEGRATION_PG`) drives all five committed extras through the real partitioned parent-only, per-partition child, and `ATTACH` path against a live Postgres, asserts each parent reaches `indisvalid=true` with a child per state, and uses `EXPLAIN` with people-api's real predicate to confirm the substring path lands on the trigram GIN and the anchored-prefix path lands on the `text_pattern_ops` b-trees. The full unit suite passes (302), plus ruff check, ruff format, and ty.

---

> [!NOTE]
> **Medium Risk**
> Changes loader rebuild schema/index behavior on large partitioned Voter tables; mis-matched index expressions would hurt CRM search performance, but there is no auth or data-mutation risk.
> 
> **Overview**
> Loader-built serving clusters will **keep CRM voter name-search indexes** after rebuilds by encoding them in the serving seed and ensuring **pg_trgm** is available before index creation.
> 
> **`_serving_seed.py`** gains five hand-added `Voter` indexes: btree on `lower("FirstName")` / `lower("LastName")`, GIN trigram indexes on those same expressions, and `Voter_last_first_id_idx` on `("LastName", "FirstName", "id")`. These mirror people-api Prisma migrations that do not run on loader clusters; expression spelling matches people-api’s `lower(...)` queries so the planner can use them.
> 
> **`create_schema`** now runs `CREATE EXTENSION IF NOT EXISTS pg_trgm` (alongside existing AWS extensions) so `build_indexes` can create the `gin_trgm_ops` definitions from the seed.
> 
> **`test_create_schema`** asserts the `pg_trgm` extension statement is executed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17d53d97879a8a9f7ec12c22517b95f8e9ca7e24. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tomer-tgp:
delegate review

tomer-tgp:
delegate review

tomer-tgp:
delegate review

hhkarimi:
delegate review

people-api-loader/tests/test_integration_pg.py
```diff
@@ -214,3 +234,105 @@ def test_partitioned_lifecycle(pg_conn: psycopg.Connection, monkeypatch: pytest.
...
+
+    # The point of the indexes: the planner must index-serve people-api's real name-search
+    # predicates. Forcing seqscan off makes the plan reveal whether an index is even usable for
+    # the query shape — a mismatched index expression would leave only a seq scan available.
+    with pg_conn.cursor() as cur:
+        _exec(cur, "SET enable_seqscan = off")
+        try:
+            # substring (CRM typeahead): only the trigram GIN can serve lower(col) LIKE '%tok%',
+            # so both name trigram indexes must appear (BitmapOr across the FirstName/LastName OR).
+            sub_plan = _explain(cur, _name_search_sql("%oh%"))
```
hhkarimi: `%oh%` is only 2 chars between the wildcards; pg_trgm needs ≥3 to extract a trigram, so `LIKE '%oh%'` degrades to a full GIN scan rather than real trigram matching. It still names the index (test passes) but doesn't prove the trigram path and is cost/version-fragile on a real DSN. Use a ≥3-char token (e.g. `%ohn%`, which matches John/Johnny).

```suggestion
            sub_plan = _explain(cur, _name_search_sql("%ohn%"))
```

people-api-loader/tests/test_integration_pg.py
```diff
@@ -214,3 +234,105 @@ def test_partitioned_lifecycle(pg_conn: psycopg.Connection, monkeypatch: pytest.
...
+                f"WHERE c.relname = '{idx.name}'",
+            )
+            want = len(_STATES)
+            assert attached == want, f"{idx.name}: {attached} children attached, want {want}"
+
+    # The point of the indexes: the planner must index-serve people-api's real name-search
+    # predicates. Forcing seqscan off makes the plan reveal whether an index is even usable for
+    # the query shape — a mismatched index expression would leave only a seq scan available.
+    with pg_conn.cursor() as cur:
+        _exec(cur, "SET enable_seqscan = off")
```
hhkarimi, line 317: These plan/catalog assertions are tuned to the local en_US.UTF-8 temp cluster and can be flakier against a real `LOADER_TEST_PG_DSN`: the planner may pick a different index (PG version/stats), `"Seq Scan" not in` trips if a partition is empty, and the `WHERE c.relname = ...` catalog lookups match the name in any schema. Fine for the gated local path — just a caveat if this is ever pointed at RDS.

hhkarimi:
delegate review

hhkarimi:
delegate review

people-api-loader/tests/test_integration_pg.py
```diff
@@ -214,3 +234,86 @@ def test_partitioned_lifecycle(pg_conn: psycopg.Connection, monkeypatch: pytest.
...
+        f"(lower(v.\"FirstName\") LIKE '{pattern}' {esc} OR lower(v.\"LastName\") LIKE '{pattern}' {esc})"
+    )
+
+
+def test_name_search_indexes_build_and_serve(pg_conn: psycopg.Connection) -> None:
+    """The committed name-search extras build through the real partitioned parent-only +
+    per-partition child + ATTACH path, and the planner serves people-api's lower(col) LIKE
+    predicates on public."Voter" — DB semantics the FakeConn unit tests can't reach. Substring rides
+    the trigram GIN; anchored-prefix rides the text_pattern_ops b-trees (a default-opclass b-tree
+    can't serve LIKE-prefix on this en_US.UTF-8 cluster).
```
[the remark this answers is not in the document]
hhkarimi: Good catch. Fixed in 672c61c, but via a locale-independent check rather than pinning `initdb` — CI runs this test against a `postgres:16` service through `LOADER_TEST_PG_DSN`, so the fixture's `initdb --locale` wouldn't affect the CI cluster anyway. The test now asserts the opclass straight from `pg_opclass` (`indclass[0]` -> `text_pattern_ops`), which catches a dropped opclass on any collation. The prefix EXPLAIN is relaxed to "index-served, no seq scan" (planner choice is cost-dependent); the opclass assertion is the real guard.

hhkarimi:
delegate review

PR #634, opened 2026-07-17
[DATA-2147] Add dbt/Databricks cost-audit skill with interactive dashboard

hhkarimi:
## What

A reusable Claude skill, `dbt-databricks-cost-audit`, that attributes Databricks and dbt Cloud spend to its real drivers and generates a **self-contained interactive HTML dashboard** from the `system.billing` and `system.query` system tables.

```
.claude/skills/dbt-databricks-cost-audit/
  SKILL.md
  scripts/cost_audit.py            # pull + render
  scripts/dashboard_template.html  # self-contained, theme-aware
```

Run:
```bash
python scripts/cost_audit.py --profile databricks-gp --warehouse <id> --since 2026-01-01 --out cost_dashboard.html
# rebuild the viz without re-billing queries:
python scripts/cost_audit.py --warehouse <id> --from-json cost_dashboard.json --out cost_dashboard.html
```

The dashboard slices cost **monthly and weekly** by product, by warehouse consumer, and by **dbt node type (model vs test vs snapshot)**, with a headline that computes testing cost and an estimate of the high-frequency job's cost.

## Why / what it already told us

Answering the two questions that motivated it, from real data:
- **Testing is ~half of dbt's Databricks cost and climbing** (allocated): ~$275 (May) → ~$818 (Jun) → ~$1,023 (Jul MTD), ~22% of the entire bill.
- **Eliminating the high-frequency job is not sufficient on its own** — that job (~$700–850/mo est) is smaller than total testing cost, and most testing runs in the twice-daily main job. The full test suite needs decoupling from every build.

## Notes

- The `$` in the dbt/consumer views is the SQL-warehouse bill **prorated by query execution time** (serverless bills warehouse uptime, not per query); the dashboard states this.
- Generated `cost_dashboard.{html,json}` contain internal cost figures and are **not** committed — the PR ships only the tool.
- Requires a Databricks CLI profile with `system.*` read access and a SQL warehouse id.

Ticket: DATA-2147

### DIFF
--- a/.claude/skills/dbt-databricks-cost-audit/SKILL.md
+++ b/.claude/skills/dbt-databricks-cost-audit/SKILL.md
@@ -59,7 +59,10 @@ figures are an allocation, not a metered per-query charge. State that.
 - **View-test antipattern (usual top driver):** a test on a `view`-materialized
   model re-executes the whole view query. Rank top nodes; if expensive nodes are
   `test` rows whose model is a view, materialize that model as a **table**
-  (in `dbt_project.yml` directory config, per `dbt/project/CLAUDE.md`).
+  (in `dbt_project.yml` directory config, per `dbt/project/CLAUDE.md`). If the
+  mart emits pretty/spaced column names (e.g. a reverse-ETL contract like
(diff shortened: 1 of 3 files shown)


hhkarimi:
delegate review

PR #636, opened 2026-07-17
[DATA-2143] quality_bench: 3-arm quality benchmark harness

datawithtristan:
## What

New `analytics/diagnostics/quality_bench/` package: the harness for the analytics-framework quality benchmark (design doc local in analytics/planning/, 2026-07-16). Modules: bank (manifest/key loading), grading (deterministic checks + cross-rep consistency), floor_gen (common floor from live warehouse inventory), prep_arms (3 isolated arm environments), run_matrix (parallel headless claude -p launcher with resume), judge (blind rubric judge), grade (scores.csv + pre-registered verdicts), README (runbook + verdict rules + holdout rotation), smoke question + key. 46 new tests (analytics suite 158 total).

## Notable decisions to review

- **analytics/pyproject.toml gains `databricks-sql-connector` + `databricks-sdk`**: pre-existing gap; nothing in analytics/ could run `databricks_conn.run_query` via uv. Required for floor generation and all benchmark runs.
- **Arms are git-archive exports, not worktrees**: final review found answer keys were recoverable via git metadata from worktree arms; archive exports carry no history.
- **Per-run isolated working dirs**: each of the up-to-72 matrix runs gets a fresh copy of its arm (cross-rep contamination would corrupt the consistency metric).
- **Floor inventory comments truncated** (first sentence, 150 chars): raw warehouse comments leaked treatment content (incl. the provenance CSV path) into the bare arm, violating the design's floor/treatment split.
- Generated `floor.md` is committed for batch reproducibility; regenerate deliberately before a batch.

## Verified

- End-to-end live smoke: bare + full arms both answered a live warehouse question correctly (875, exact) and graded green; the smoke caught and fixed a real floor-contract bug (YAML quoting).
- Final whole-branch review (2 Criticals, 4 Importants) fixed and re-verified: ready to merge.

## Pre-first-batch gates (post-merge, in README checklist)

1. Live judge probe (`grade.py --judge` on the smoke batch).
2. Full-arm headless probe confirming the analytics-process skill actually engages under the arm settings.

Next: DATA-2144 (question bank + gold-run keys), DATA-2145 (run matrix + verdict).

---

> [!NOTE]
> **Medium Risk**
> New offline benchmark tooling with live Databricks and headless Claude runs; no production path changes, but adds warehouse-query deps to analytics and a large generated floor artifact.
> 
> **Overview**
> Adds **`analytics/diagnostics/quality_bench/`**, a harness to compare **full** (framework + skills), **knowledge** (skills only), and **bare** (shared Databricks floor) arms on live warehouse questions, with deterministic grading, optional blind judge, and pre-registered verdict rules.
> 
> **Pipeline:** `floor_gen.py` builds a committed **`floor.md`** from warehouse table inventory (comments truncated so bare arms don’t get treatment leakage); **`prep_arms.py`** materializes arms via **`git archive`** (no `.git`/keys), strips `quality_bench` from run trees, and sets headless Claude permissions; **`run_matrix.py`** runs question × arm × rep with per-run arm copies, resume via `state.json`, and transcript capture; **`grade.py`** emits `scores.csv` / `report.md` and evaluates trustworthiness, overhead, and bloat signals.
> 
> **Supporting pieces:** manifest/key loaders (`bank.py`), YAML results-block checks + cross-rep consistency (`grading.py`), judge CLI (`judge.py`), smoke question `q00_smoke`, README runbook, **gitignore** for `results/*`, **`pyyaml`** and **Databricks** deps on the analytics project, and broad pytest coverage.
> 
> **Notable design choices in this PR:** archive-based arms (keys not recoverable from history), isolated run directories (no cross-rep contamination), and inventory truncation on the common floor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4664cb3f5f94c1d7266cfcbace7d3bf5563dad9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


datawithtristan:
delegate review

datawithtristan:
bugbot run

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,218 @@
...
+
+def rep_threshold(reps: int) -> int:
+    """Design §8: a question passes when >= 2/3 of its reps pass (2 of 3 at the
+    default rep count). Scale with the batch's actual rep count."""
+    return max(1, math.ceil(reps * 2 / 3))
+
+
+def _question_passes(df: pd.DataFrame, arm: str, qid: str, threshold: int) -> bool:
+    reps = df[(df.arm == arm) & (df.question_id == qid)]
+    return int(reps.numbers_pass.sum()) >= threshold
```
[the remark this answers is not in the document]
datawithtristan: Ruled on with Tristan: keeping the pre-registered gate as designed. Design §8 defines a rule-1 pass as within-tolerance numbers + zero severity-1 + cell consistency; adherence is a separate diagnostic layer (§7). Two reasons not to gate on sources/assumptions: (1) it rewrites a pre-registered verdict rule after results-infrastructure exists, and (2) mandatory sources are treatment-only (e.g. the provenance CSV), so gating would make the bare arm auto-fail every trap by construction and turn rule 2 (full beats bare on traps) into a tautology — the bare arm must be able to pass on numbers alone for the comparison to mean anything. The underlying concern is addressed in a701055: report.md now includes a per question x arm adherence section (sources/assumptions rep rates) so a passing number derived without consulting the mandatory sources is visible, just not verdict-gating.

datawithtristan:
delegate review

datawithtristan:
delegate review

datawithtristan:
delegate review

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,237 @@
...
+    questions = load_manifest(here / "questions" / "manifest.yaml")
+    keys = {q.id: load_key(here / "keys" / q.key_file) for q in questions}
+
+    rows, blocks = [], {}
+    judge_eligible = judge_ok = 0
+    for run_id, run_state in state["runs"].items():
+        if not run_state.get("ok"):
+            continue
+        qid = run_id.split("__")[0]
+        row = grade_run(run_state, keys[qid])
```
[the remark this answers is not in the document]
datawithtristan: Previously ruled (see finding 7bf138be on this PR): declined as marginal with Tristan's sign-off, tracked in the DATA-2143 ClickUp comment. The scenario requires removing/renaming a manifest question between run_matrix and grade.py — relevant once holdout rotation happens (DATA-2144+), and the fix will land with that work if it bites. Not blocking this merge.

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,237 @@
...
+
+    # Coverage denominator = every question with ANY recorded run, not only the
+    # ok ones. A question whose runs all failed must still count against rule 1,
+    # otherwise a silent drop would inflate the pass rate.
+    qids_in_state = {rid.split("__")[0] for rid in state["runs"]}
+    denom_questions = [q for q in questions if q.id in qids_in_state]
+    cells = {ck: grading.cell_consistency(bs, keys[ck[0]]) for ck, bs in blocks.items()}
+    # Rep count comes from the recorded run ids, so a resumed or non-default
+    # --reps batch grades against its own size, not a hardcoded 3.
+    reps_recorded = max((int(rid.split("__")[2].lstrip("r")) for rid in state["runs"]), default=0)
```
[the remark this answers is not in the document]
datawithtristan: Declined — the current behavior is the intended conservative choice, not a distortion. Deriving the threshold from ok rows only (per-cell max of the graded df) would let a question that crashed down to 1 ok rep pass rule 1 on a single observation: the trust bar would adapt downward exactly when evidence is missing, inflating the pass rate on partial data. That inversion is the same failure mode the coverage denominator deliberately guards against ('a question whose runs all failed must still count against rule 1'). The designed remedy for a partial batch is resume (state.json exists for precisely this) and grade when complete; grading early should under-certify, and report.md's coverage/missing-by-arm lines make the partial state visible.

analytics/diagnostics/quality_bench/prep_arms.py
```diff
@@ -0,0 +1,150 @@
...
+    shutil.rmtree(dest / "analytics" / "diagnostics" / "quality_bench", ignore_errors=True)
+    claude_md = dest / "CLAUDE.md"
+    existing = claude_md.read_text() if claude_md.exists() else ""
+    claude_md.write_text(existing + "\n\n" + _fill(floor_text, "analytics/lib", "analytics"))
+
+    if arm == "knowledge":
+        skills = dest / ".claude" / "skills"
+        for child in skills.iterdir() if skills.exists() else []:
+            if child.name not in KNOWLEDGE_SKILLS:
+                shutil.rmtree(child)
```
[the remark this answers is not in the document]
datawithtristan: Verified unreachable in the actual flow: arm dirs are git-archive exports (prep_arms._export), which contain tracked files only. There are zero tracked plain files directly under .claude/skills/ (git ls-files confirms), and the local .DS_Store that would trigger this is gitignored, so it never reaches an export. The test fixture's directories-only shape mirrors the archive reality rather than missing a case. Agreed it's a cheap robustness guard — noted for the DATA-2144 hardening pass if the tracked layout ever changes; not blocking this merge.

datawithtristan:
delegate review

analytics/diagnostics/quality_bench/judge.py
```diff
@@ -0,0 +1,98 @@
...
+            text=True,
+            timeout=timeout_s,
+        )
+    if proc.returncode != 0:
+        return None
+    try:
+        result = json.loads(proc.stdout).get("result", "")
+    except json.JSONDecodeError:
+        return None
+    return parse_judge_output(result)
```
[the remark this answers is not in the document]
datawithtristan: Fixed in fe4e1ac: judge_answer now catches TimeoutExpired/OSError and a null CLI result, returning None — grade.py records the run as 'judge FAILED' and keeps its deterministic grades. Covered by test_judge_answer_contains_timeout and test_judge_answer_null_result_returns_none.

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,218 @@
...
+    # bool(full_cells) keeps rule 1 non-vacuous: a batch where the full arm
+    # produced no graded cells must not pass on all([]) == True.
+    full_cells_ok = bool(full_cells) and all(c["consistent"] for c in full_cells)
+    # ceil(7/8 * N) generalizes the pre-registered ">= 7 of 8" proportionally,
+    # so a small denominator (e.g. a 1-question smoke batch) gets no free slack.
+    rule1 = bool(qids) and full_pass >= math.ceil(len(qids) * 7 / 8) and sev1_count == 0 and full_cells_ok
+
+    rule2 = None
+    if "bare" in set(df.arm):
+        rule2 = _pass_count(df, "full", trap_qids, threshold) > _pass_count(df, "bare", trap_qids, threshold)
```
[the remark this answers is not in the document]
datawithtristan: Fixed in fe4e1ac: rule2_earns_overhead is now None (inapplicable) when the batch has no trap questions, matching the bare-arm-absent case. Covered by test_rule2_inapplicable_without_trap_questions.

analytics/diagnostics/quality_bench/grade.py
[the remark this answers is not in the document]
datawithtristan: Declined as marginal with Tristan's sign-off (2026-07-20), tracked in the DATA-2143 ClickUp comment: requires removing/renaming a manifest question between run_matrix and grade.py, which only becomes realistic once holdout rotation starts (DATA-2144+). Will land with that work if it bites. Delegate re-raised this as finding c4fcc726 — same disposition.

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,237 @@
...
+    judge_eligible = judge_ok = 0
+    for run_id, run_state in state["runs"].items():
+        if not run_state.get("ok"):
+            continue
+        qid = run_id.split("__")[0]
+        row = grade_run(run_state, keys[qid])
+        blocks.setdefault((row["question_id"], row["arm"]), []).append(row.pop("_block"))
+        if args.judge:
+            judge_eligible += 1
+            answer = Path(run_state["answer_file"]).read_text()
```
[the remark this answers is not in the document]
datawithtristan: Declined as marginal with Tristan's sign-off (2026-07-20), tracked in the DATA-2143 ClickUp comment: run_matrix only ever sets ok=True together with answer_file (out.update(ok=True, ..., answer_file=...)), so the bad state shape requires hand-edited state.json. The guard in grade_run is defensive consistency, not evidence the shape occurs. On the fold-in-if-it-bites list for DATA-2144.

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,188 @@
...
+    for q in questions:
+        prompt = (questions_dir / q.prompt_file).read_text()
+        for arm, arm_dir in arm_dirs.items():
+            for rep in range(1, reps + 1):
+                runs.append(RunSpec(q.id, arm, rep, prompt, arm_dir))
+    return runs
+
+
+def transcript_path(arm_dir: Path, session_id: str) -> Path:
+    munged = re.sub(r"[/_.]", "-", str(arm_dir))
```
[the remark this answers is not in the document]
datawithtristan: Refuted empirically. The claim that Claude CLI keys ~/.claude/projects/ by replacing only '/' is wrong: the live smoke run's actual transcript dirs are '-Users-tristan--cache-gp-quality-bench-arms-bare' and '...-arms-full' — for arm path ~/.cache/gp_quality_bench/arms/bare, i.e. '.' and '_' ARE replaced with '-' (note '--cache' from '/.cache'). Zero directories under ~/.claude/projects/ contain a literal dot or underscore. The [/_.] munge matches observed CLI behavior, and transcript collection demonstrably worked end-to-end in the smoke batch (sources checks graded against real transcripts). The suggested fix would break transcript resolution.

analytics/tests/test_quality_bench_run_matrix.py
```diff
@@ -0,0 +1,184 @@
...
+    runs = run_matrix.build_runs([q()], arm_dirs, reps=2, questions_dir=tmp_path)
+    assert len(runs) == 4  # 1 question x 2 arms x 2 reps
+    assert runs[0].prompt == "What is the monthly funnel?"
+    assert runs[0].run_id == "q01__full__r1"
+
+
+def test_transcript_path_munges_cwd(tmp_path: Path):
+    p = run_matrix.transcript_path(Path("/Users/t/.cache/gp_quality_bench/arms/full"), "abc123")
+    assert p.name == "abc123.jsonl"
+    assert "-Users-t--cache-gp-quality-bench-arms-full" in str(p.parent)
```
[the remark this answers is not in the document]
datawithtristan: Follows from the munge finding (ebf5b5fc), which is refuted empirically: real ~/.claude/projects/ keys replace '.', '_', AND '/' with '-' (verified against the live smoke run's actual transcript dirs). The assertion encodes the correct observed format, so it is not tautological — the suggested assertion is the one that would encode a wrong format.

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            payload = json.loads(proc.stdout)
+        except json.JSONDecodeError:
+            out["error"] = f"unparseable CLI output: {proc.stdout[-500:]}"
+            return out
+        session_id = payload.get("session_id", "")
+        answer = payload.get("result", "")
+        answer_file = batch_dir / f"{spec.run_id}.answer.md"
+        answer_file.write_text(answer)
+        out.update(ok=True, session_id=session_id, answer_file=str(answer_file))
```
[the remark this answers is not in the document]
datawithtristan: Valid hardening, acknowledged and tracked (DATA-2143 ClickUp comment, fold-in list for DATA-2144): an exit-0 CLI response missing session_id would set ok=True with no transcript, and resume would not retry that run. Not applied now — the PR is converged and in human review; in practice the CLI's json output format always carries session_id on success, and a run graded without a transcript fails its mandatory-source checks loudly ('transcript missing') rather than silently passing.

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,237 @@
...
+    reps = df[(df.arm == arm) & (df.question_id == qid)]
+    return int(reps.numbers_pass.sum()) >= threshold
+
+
+def _pass_count(df: pd.DataFrame, arm: str, qids: list[str], threshold: int) -> int:
+    return sum(
+        _question_passes(df, arm, q, threshold)
+        for q in qids
+        if not df[(df.arm == arm) & (df.question_id == q)].empty
+    )
```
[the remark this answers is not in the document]
datawithtristan: Refuted on impact: the guard and the suggested guard-free version are numerically identical. A question skipped by the .empty guard contributes 0 to the sum; the same question included would return _question_passes = (0 >= threshold) = False, contributing 0 — threshold is always >= 1 (rep_threshold floors at 1, default 2). Rule 2 is a count comparison with no denominator, so no silent-drop inflation is possible here; full vs bare counts are unchanged either way. Agreed the guard is redundant and can go in a cleanup pass, but there is no behavioral bug.

datawithtristan:
## Bot-review disposition summary (for human review)

All automated review threads are answered and resolved. Full accounting across 6 delegate rounds + 2 Bugbot passes:

**Fixed in code (11 findings):** atomic state.json writes; zero-key answers grade correctly (check_numbers + cell_consistency); all-failed batches no longer crash grading; rule 1 non-vacuous (proportional ceil(7/8) bar, requires full-arm cells); rep pass threshold scales with --reps (ceil 2/3); required_assumptions enforced as assumptions_pass; judge subprocess failures contained; rule 2 = None without trap questions; per-cell adherence section in report.md; judge-only severity-1 test coverage; import lint fix.

**Declined by design ruling (Tristan, 2026-07-20):** gating rule-1 passes on sources_pass/assumptions_pass — rejected because mandatory sources are treatment-only, so gating would make the bare arm auto-fail traps by construction and reduce rule 2 to a tautology. Adherence is reported per question x arm instead (not verdict-gated).

**Declined as marginal, tracked for DATA-2144 (3):** stale-manifest-qid KeyError in grade.py (raised twice); defensive answer_file guard in the judge branch; session_id-absent-but-exit-0 CLI edge in run_matrix. All on the fold-in-if-they-bite list in the DATA-2143 ClickUp ticket.

**Refuted with evidence (3):** transcript-path munge regex "wrong" — disproven empirically against the live smoke run's actual ~/.claude/projects/ keys ('.', '_', '/' all munge to '-'); the paired "tautological test" finding falls with it; _pass_count .empty guard "silent drop" — numerically identical either way (skip and count-as-False both contribute 0; threshold >= 1 always).

Checks are green across the board at c161c2a (incl. dbt Cloud after the branch update pulled main's election-api race fix). 173 tests passing.

analytics/diagnostics/quality_bench/prep_arms.py
```diff
@@ -0,0 +1,150 @@
...
+    sys.path.insert(0, str(Path(__file__).parents[1]))
+    from quality_bench.bank import ARMS
+
+KNOWLEDGE_SKILLS = ["win-analytics-knowledge", "serve-analytics-knowledge"]
+
+SETTINGS_JSON = {
+    "permissions": {
+        "defaultMode": "acceptEdits",
+        "allow": [
+            "Bash(*)",
```
sanjayr1, line 37: **[P1] This is a permission boundary, not a filesystem-isolation boundary.**

Because the child process runs as the host user and receives unrestricted `Bash(*)` / `Read(*)`, removing `quality_bench` from the archive does not make the grader-side keys inaccessible. A run can still search the original checkout, inspect sibling run directories and `~/.claude`, or read unrelated inherited environment values. That defeats the stated answer-key blindness even if the model does not happen to exploit it during the smoke.

Could we run each cell behind an OS/container sandbox with an ephemeral `HOME`, an allowlisted environment, and only the arm plus explicitly required Databricks credentials mounted? The archive export and tool permissions are still useful defense-in-depth, but they should not be the isolation mechanism.

analytics/diagnostics/quality_bench/prep_arms.py
sanjayr1, line 129: **[P1] Make the arms self-contained, comparable one-factor treatments.**

The prepared knowledge arm currently has eight broken relative links into content removed with `analytics-process/` and `data-matching/`. More broadly, the full and knowledge arms receive most of the repository while bare gets a scratch tree, and full retains unrelated skills/reviewers, so a win cannot be attributed cleanly to the intended treatment. Start every arm from the same sanitized common substrate; define bare as common-only, knowledge as common plus the two product-knowledge skills and their factual dependencies, and full as that plus the process and explicitly claimed reviewers. Then run a post-preparation integrity check over links and a content manifest so each arm is both runnable and auditable.

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,188 @@
...
+    batch_dir: Path,
+    keep_run_dir: bool = False,
+    runner=subprocess.run,
+) -> dict:
+    batch_dir.mkdir(parents=True, exist_ok=True)
+    # Per-run isolation: copy the arm into runs/<run_id> and run there. A fresh
+    # cwd per run means no cross-rep leakage (one rep can't mutate the arm the
+    # next rep sees) and a fresh per-path auto-memory. uv re-creates the venv
+    # from cache on the first `uv run` in the copy, so the copy stays cheap.
+    run_dir = spec.arm_dir.parent / "runs" / spec.run_id
```
sanjayr1: **[P1] Make this path unique per batch and per attempt.**

The current path is only `question × arm × rep`. Two batches running concurrently can therefore remove/copy into the same directory, and sequential batches reuse the same Claude project key under `~/.claude/projects`. That conflicts with the “fresh per-path auto-memory” guarantee and can let prior-batch state influence a later experiment. A failed retry also reuses the same identity.

Please include the batch plus a random attempt ID in the working path, and preferably give each run an isolated Claude config/home directory as part of the sandbox boundary.

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,188 @@
...
+        action="store_true",
+        help="keep the per-run copy of each arm (runs/<run_id>) instead of deleting it after the run",
+    )
+    args = parser.parse_args()
+
+    here = Path(__file__).parent
+    batch_dir = here / "results" / args.batch
+    batch_dir.mkdir(parents=True, exist_ok=True)
+    state_file = batch_dir / "state.json"
+    state = json.loads(state_file.read_text()) if state_file.exists() else {"runs": {}}
```
sanjayr1, line 304: **[P1] Bind resume to an immutable benchmark bundle.**

The state file records outcomes but not the inputs that produced them, so a resumed batch can silently mix changed code, questions, keys, prompts, models, or data. Create a `BatchConfig` fingerprint containing the resolved source SHA; hashes of the floor, arm manifests, prompts, questions/keys, helper, and lockfile; exact responder and judge model versions; warehouse snapshot/key dates; arms/repetitions; and the schedule seed/settings. Persist it before execution and refuse to resume when it differs. This also gives every reported verdict a reproducible provenance record.

analytics/diagnostics/quality_bench/grading.py
```diff
@@ -0,0 +1,130 @@
...
+            if mean:
+                spreads[name] = (max(vals) - min(vals)) / abs(mean) * 100
+            else:
+                # Zero mean: identical reps (all zero) are perfectly consistent.
+                spreads[name] = 0.0 if max(vals) == min(vals) else float("inf")
+    forks = set(key.required_resolutions)
+    agreement = {}
+    for fork in forks:
+        seen = {_resolutions(b).get(fork) for b in parsed}
+        agreement[fork] = len(seen) == 1 and None not in seen
```
sanjayr1, line 148: **[P1] Validate correctness and completeness of the assumptions ledger.**

Agreement only shows that repetitions returned the same value; it does not show that the value is correct. A consistently wrong required resolution currently passes this check, and the assumptions fork only checks key presence, allowing an empty or unverified resolution to count. Parse a typed ledger with controlled assumption IDs, require non-empty resolutions plus verification status/source, and compare each expected resolved value against the answer before marking consistency. Gate full deterministic correctness on that result as the benchmark design describes.

analytics/diagnostics/quality_bench/judge.py
```diff
@@ -0,0 +1,104 @@
...
+    """Parse the judge's fenced YAML output. Extracts the last `judge:` block; echoed fences could shadow the real verdict."""
+    for match in reversed(YAML_FENCE.findall(text)):
+        try:
+            parsed = yaml.safe_load(match)
+        except yaml.YAMLError:
+            continue
+        if not (isinstance(parsed, dict) and "judge" in parsed and isinstance(parsed.get("judge"), dict)):
+            continue
+        j = parsed["judge"]
+        scores = j.get("scores", {})
```
sanjayr1: **[P1] Make judge parsing and provenance durable.**

The parser validates only the outer object, so values such as `scores: null` or strings in boolean fields can crash or be misinterpreted. Validate the complete nested schema, score ranges, and boolean types. Also persist the immutable raw `JudgeResult`—scores, rationale, session, exact model, rubric hash, answer hash, and key hash—and reuse it unless an explicit re-judge is requested. Because the responder and judge are currently from the same model family, either use an independent calibrated judge or document and measure that bias against a human-labeled sample.

analytics/diagnostics/quality_bench/bank.py
```diff
@@ -0,0 +1,90 @@
...
+    required_assumptions: list[str] = field(default_factory=list)
+    intent_card: str = ""
+
+
+def load_manifest(path: Path) -> list[Question]:
+    raw = yaml.safe_load(path.read_text())
+    questions = []
+    for row in raw["questions"]:
+        q = Question(**row)
+        if "__" in q.id:
```
sanjayr1: **[P1] Validate identifiers before they are used in paths and run IDs.**

Rejecting `__` avoids one parsing ambiguity, but duplicate IDs and path components are still accepted. A question ID such as `../../unrelated` makes the later `runs / run_id` path resolve outside the runs directory before `shutil.rmtree`; duplicate IDs produce colliding concurrent cells. Prompt/key paths can similarly escape their intended directories, and the loaded key ID is never matched back to the manifest ID.

Please enforce a strict identifier regex, uniqueness, basename/containment rules for referenced files, and `key.id == question.id`. It would be useful to validate the entire bank once before preparing or launching any runs.

analytics/diagnostics/quality_bench/keys/q00_smoke_key.yaml
```diff
@@ -0,0 +1,14 @@
+id: q00_smoke
+as_of: "2026-07-17"
+numbers:
+  # The floor inventory shows 874 tables (it excludes __dbt_tmp), while a live
+  # count of the dbt schema returns 875. The 5% tolerance absorbs the gap.
+  - name: dbt_table_count
+    value: 875
+    tolerance_pct: 5.0
+required_resolutions: {}
+mandatory_sources: []
```
sanjayr1, line 10: **[P2] This smoke can pass without exercising Databricks.**

The floor already contains the inventory, the 5% tolerance accepts its 874-row count against the 875 key, and there is no mandatory source. I reproduced an offline answer grading `numbers_pass=True` and `sources_pass=True` with `transcript_ok=False`. That proves answer parsing, but not the live warehouse/auth/query/transcript path the README says this smoke verifies.

Please require evidence of an executed query and make transcript presence part of smoke success, or choose a live value that is not already derivable from the common floor. Ideally the smoke command should exit nonzero if that invariant is not met.

analytics/diagnostics/quality_bench/run_matrix.py
sanjayr1, line 264: **[P2] Capture complete execution metadata through one runner adapter.**

The persisted result has duration and transcript but not input/output tokens, context loaded, tool/skill/agent/SQL calls, retries, rate-limit delays, or estimated cost. Without those fields the harness cannot substantiate its overhead/bloat claims or verify that an arm actually used its assigned treatment. Centralize invocation and parsing in a shared runner adapter and emit structured execution metadata for every run.

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,237 @@
...
+    rule1 = bool(qids) and full_pass >= math.ceil(len(qids) * 7 / 8) and sev1_count == 0 and full_cells_ok
+
+    rule2 = None
+    # No trap questions in the batch -> rule 2 is inapplicable (None), the same
+    # as when the bare arm is absent — not a failed verdict on two zero counts.
+    if "bare" in set(df.arm) and trap_qids:
+        rule2 = _pass_count(df, "full", trap_qids, threshold) > _pass_count(df, "bare", trap_qids, threshold)
+
+    rule3 = None
+    if "knowledge" in set(df.arm):
```
sanjayr1, line 130: **[P2] Treat Rule 3 as paired non-inferiority, not a one-sided win count.**

This calculation can report a bloat signal from knowledge-only rows because it does not require matched full-arm coverage. Return indeterminate unless both arms have complete paired observations for the same questions/repetitions. Then define a quality non-inferiority margin, evaluate the paired quality difference with uncertainty, and require a meaningful overhead threshold before concluding the full process adds bloat.

sanjayr1 (changes requested):
## Consolidated review — changes requested

The module decomposition and test coverage are a good foundation. My concern is that this harness is intended to support a consequential decision—whether the analytics framework improves answer quality enough to justify its overhead—while several current behaviors can produce a confident but experimentally invalid verdict.

### Merge blockers

1. **Isolation:** benchmark processes can still read the original checkout, grader keys, sibling runs, host environment, and prior Claude project state.
2. **Treatment integrity:** the arms are not self-contained or one-factor contrasts; retained skills have broken dependencies and repository/runtime access differs by arm.
3. **Reproducibility:** resume is not bound to an immutable source/model/floor/question/key/data/dependency configuration, and run identities are reused.
4. **Grading correctness:** expected scoping resolutions are not enforced, nested model output is insufficiently validated, and one-sided batches can emit comparative verdicts.
5. **Input safety:** manifest identifiers and paths reach recursive filesystem operations without adequate validation.

### Required before the first real benchmark

- define a coherent non-interactive mode for the otherwise interactive full framework;
- make the smoke prove a live warehouse query;
- capture usage, latency, tool/query activity, judge provenance, and per-check evidence;
- freeze or identify the warehouse/key snapshot;
- block and randomize arm scheduling, use paired effect estimates, and treat pruning as a non-inferiority decision;
- enforce/record holdout use and report results by product/trap strata;
- commit the pre-registration design or link an immutable reviewed version.

### Consolidation direction

Use one immutable `BatchConfig`/benchmark bundle and typed `RunResult`, `AnswerBlock`, `JudgeResult`, and `RunGrade` records. Derive every arm from the same sanitized substrate, centralize Claude invocation in one adapter, and persist append-only run/check/judge evidence. Keep the current bank, preparation, execution, deterministic-grading, and judge modules separate; those boundaries are useful.

Validation on head `c161c2a`: complete analytics suite 173 passed, focused reproductions confirmed, CI green, and `git diff --check` clean.

sanjayr1 (commented):
Superseded by the consolidated experiment-design review: https://github.com/thegoodparty/gp-data-platform/pull/636#pullrequestreview-4739661500

analytics/diagnostics/quality_bench/prep_arms.py
```diff
@@ -0,0 +1,150 @@
...
+        raise ValueError(f"unknown arm {arm!r}; expected one of {ARMS}")
+    dest = arms_root / arm
+
+    if arm == "bare":
+        if dest.exists():
+            shutil.rmtree(dest)
+        env = dest / "env"
+        env.mkdir(parents=True)
+        (dest / "CLAUDE.md").write_text(_fill(floor_text, "env", "env"))
+        shutil.copy(repo_root / "analytics" / "lib" / "databricks_conn.py", env / "databricks_conn.py")
```
sanjayr1: **[P1] The bare arm is not built from the same source/dependency provenance as the archived arms.**

`full` and `knowledge` receive files and `uv.lock` from `ref`, but `bare` ignores `ref` here, copies `databricks_conn.py` from the current working tree, and resolves the broad dependency ranges in `BARE_PYPROJECT` without a lock. A dirty checkout, stale branch, or later package release can therefore make the control arm run different helper code or dependency versions from the treatment arms.

Please derive the common runner/helper from the same resolved commit and lock for every arm, or create one locked runner environment outside the treatment content and mount/use it identically. Record its hash in the batch manifest.

sanjayr1 (commented):
Superseded by the consolidated experiment-design review: https://github.com/thegoodparty/gp-data-platform/pull/636#pullrequestreview-4739661500

analytics/diagnostics/quality_bench/floor_static.md
```diff
@@ -0,0 +1,48 @@
+# Working context (common floor)
+
+You are answering a product analytics question against GoodParty's Databricks
+warehouse. Everything you need to connect is below. Answer the question you are
+given; do not ask clarifying questions (no user is present) — resolve ambiguity
```
sanjayr1, line 5: **[P1] Define a noninteractive benchmark contract for the full process.**

The treatment says to ask one question at a time and wait for answers, while the harness supplies a one-shot prompt and expects completion without operator input. That can make the full arm fail specifically because it follows its instructions. Either define a benchmark mode with a preapproved analysis brief and explicit defaults, or implement a multi-turn driver that answers the process questions consistently. Record the chosen brief/defaults in the batch bundle so every arm and repetition sees the same problem.

analytics/diagnostics/quality_bench/floor_gen.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def build_inventory_md(run_query: RunQuery) -> str:
+    df = run_query(INVENTORY_SQL)
+    lines = ["| table | description |", "|---|---|"]
+    for _, row in df.iterrows():
+        if isinstance(row["comment"], str) and row["comment"]:
+            # Collapse whitespace runs (incl. newlines) to single spaces and escape pipes
+            comment = " ".join(str(row["comment"]).split()).replace("|", "\\|")
+            # Mechanical truncation: the floor states WHAT exists, not what it
```
sanjayr1, line 32: **[P1] Keep product semantics out of the common floor.**

The generated inventory includes the first sentence of every Markdown file plus hundreds of highly descriptive data rows. That gives the bare arm a lossy copy of product/process knowledge and makes the very large floor a competing treatment. Generate the control floor from identifiers and neutral type/size metadata only, maintain an explicit allowlist of shared operational facts, and add a leakage test for treatment-specific terms and content. This will make a bare-versus-knowledge difference interpretable.

analytics/diagnostics/quality_bench/grade.py
```diff
@@ -0,0 +1,237 @@
...
+    sev1 = grading.check_severity1(answer, key)
+    assumptions = grading.check_assumptions(block, key)
+    return {
+        "run_id": run_id,
+        "question_id": question_id,
+        "arm": arm,
+        "rep": int(rep.lstrip("r")),
+        "results_block_ok": block is not None,
+        "transcript_ok": transcript_ok,
+        "numbers_pass": all(c.passed for c in numbers),
```
sanjayr1, line 56: **[P2] Persist structured evidence for every grade.**

This reduces detailed CheckResult objects to summary booleans, and the source check searches raw transcript text, so merely mentioning a path can count without a successful tool or SQL event. Parse structured execution events, then persist each check ID with expected/actual values, tolerance, source execution evidence, and judge rationale in an append-only checks artifact. Reports remain concise, but a reviewer can reproduce and diagnose every pass or failure.

sanjayr1 (commented):
## Consolidated experiment-design notes

Three remaining inline notes cover the benchmark protocol, control-arm contamination, and audit evidence. Before the first decision-bearing batch, I would also freeze a randomized/block-paired schedule, define a meaningful Rule 2 effect and Rule 3 non-inferiority margin, stratify the question bank and reserve a holdout, and commit the preregistered design plus frozen data/key snapshot. Those steps make the result defensible rather than merely repeatable.

analytics/diagnostics/quality_bench/bank.py
datawithtristan: Fixed in 0a7b278. load_manifest now enforces a strict id regex ([a-z0-9][a-z0-9_-]{0,63}, no __), rejects duplicate ids, and rejects prompt_file/key_file values that are absolute or contain .. segments. load_key takes an expected_id and grade.py passes the question id, so a key grading the wrong question fails loudly. Covered by four new tests in test_quality_bench_bank.py, including the ../../unrelated traversal case.

analytics/diagnostics/quality_bench/grading.py
datawithtristan: Partially addressed in 0a7b278: required_assumptions forks now need a non-empty, non-null resolution to pass (a bare fork entry no longer counts), and a YAML-null resolution maps to empty rather than the string "None". Tests cover empty, null, and whitespace-only resolutions.

On gating deterministic correctness on resolution values: comparing each resolved value against the key already happens in two places (cell_consistency checks cross-rep resolution agreement against required_resolutions, and the judge scores scoping_correctness against the key). Making it a rule-1 gate is the same pre-registration question Tristan ruled on earlier in this PR: adherence is reported per question x arm but not gated, because mandatory-source and resolution guidance is treatment-only and gating would build the treatment into the verdict. Happy to revisit the reporting depth in the DATA-2144/2145 follow-ups.

analytics/diagnostics/quality_bench/judge.py
datawithtristan: Parsing hardened in 0a7b278: the judge block is now schema-validated end to end. scores must be a dict with every rubric dimension present as a real int in 0-2 (bools rejected, since bool is an int subclass), and severity1_found must be a bool. scores: null, string scores, out-of-range values, and non-bool severity1_found all return None, which grade.py already records as judge FAILED without overriding deterministic grades. New tests cover each malformed shape.

The other two asks here (persisting the raw JudgeResult with provenance hashes, and the same-family judge bias) are not in this commit. Judge provenance fits naturally with the evidence-artifact work you flagged on grade.py, and the judge-bias question is part of the pre-first-batch live judge probe already on the README checklist. Proposing both land as pre-first-batch work rather than in this PR.

analytics/diagnostics/quality_bench/run_matrix.py
datawithtristan: Fixed in 3da281b: run copies now live at runs/<batch>/<run_id>, so concurrent or sequential batches never share working paths, and each run executes with an ephemeral HOME (8eca9b9 adds the auth seeding), which removes the shared ~/.claude/projects key reuse entirely — transcripts are collected from inside the per-run HOME. Verified with a live smoke run.

analytics/diagnostics/quality_bench/prep_arms.py
datawithtristan: Partially addressed in 3da281b/8eca9b9: each run now gets an ephemeral HOME and an allowlisted environment (PATH/locale plus DATABRICKS_*/ANTHROPIC_*/CLAUDE_* only), so the operator's ~/.claude state and unrelated env vars are out of reach. Agreed this is still a permission boundary rather than a filesystem one — the container-grade version (only the arm dir and creds mounted, network allowlist) is captured as a backlog ticket with your framing: https://app.clickup.com/t/86ajmyk2q. Interim mitigation for decision-bearing batches: spot-check transcripts for out-of-arm reads.

analytics/diagnostics/quality_bench/run_matrix.py
datawithtristan: Proportionate version landed in 3da281b: state.json now records a fingerprint (sha256 of question prompts + floor, model, harness git sha) and resume refuses when prompts/floor/model changed — verified live (a --model change on resume exits with 'use a fresh --batch name'). Keys are deliberately exempt: they are grader-side and applied uniformly at grade time, so a key edit cannot mix run provenance. The harness sha is recorded but not gated so an unrelated commit does not force a full re-run. The full immutable BatchConfig bundle (arm manifests, lockfile, judge model, snapshot dates, schedule seed) is tracked as backlog: https://app.clickup.com/t/86ajmykh4.

analytics/diagnostics/quality_bench/prep_arms.py
datawithtristan: Agreed on the diagnosis, deferred by decision: additive construction from a common substrate (with a post-prep dead-link/manifest integrity check) is captured as DATA-2164 (https://app.clickup.com/t/86ajmykcc). One flag before anyone implements it: the pre-registered full arm is 'the repo as deployed', because the claim under test is the framework as actually used; redefining full as a curated bundle is cleaner attribution but tests a different claim, so that trade gets an explicit sign-off first. The knowledge-arm dead links get fixed upstream by making the two knowledge skills self-contained (copying the linked analytics-process content into the knowledge arm would violate the one-factor contrast in the other direction).

analytics/diagnostics/quality_bench/floor_gen.py
datawithtristan: Deferred to DATA-2164 (https://app.clickup.com/t/86ajmykcc) together with the arm-construction rework, including your allowlist idea plus canary-string leakage tests. Design tension we want to resolve deliberately rather than in this PR: the current first-sentence/150-char truncation was a considered midpoint, because a names-only inventory makes the bare arm fail on pure table discovery among hundreds of tables, which overstates the framework's value in the other direction.

analytics/diagnostics/quality_bench/floor_static.md
datawithtristan: Deferred to DATA-2164 (https://app.clickup.com/t/86ajmykcc). The current noninteractive contract is the floor's output contract (resolve forks yourself, record them in the assumptions ledger), and whether the full arm's process skill actually complies headless is exactly what the pre-first-batch skill-engagement probe on the README checklist verifies. The fuller solution in the ticket is a multi-turn driver that answers process questions from the grader-side intent_card consistently across arms and reps.

analytics/diagnostics/quality_bench/keys/q00_smoke_key.yaml
datawithtristan: Acknowledged; tracked in the pre-first-batch hardening ticket (https://app.clickup.com/t/86ajmykh4, evidence-artifact item: a consulted source must be backed by a real tool/SQL event, which also makes the smoke prove an executed query rather than a mentionable path). Noting for the record that today's re-run of the smoke under the new ephemeral-HOME isolation did execute a live Databricks query end to end, but you are right that nothing deterministically enforces that yet.

analytics/diagnostics/quality_bench/run_matrix.py
datawithtristan: Deferred to https://app.clickup.com/t/86ajmykh4 (runner-adapter item: tokens in/out, cost, latency, tool/skill/SQL call counts, retries). Agreed it is required before rule 3 overhead claims mean anything; it does not block merging the harness.

analytics/diagnostics/quality_bench/grade.py
datawithtristan: Deferred to https://app.clickup.com/t/86ajmykh4 (protocol item: paired coverage requirement, a defined non-inferiority margin, and a meaningful overhead threshold before a bloat conclusion). Until then rule 3 stays a coarse pre-registered signal that feeds prune candidates into the existing calibration loop, not an automatic prune decision.

analytics/diagnostics/quality_bench/grade.py
datawithtristan: Deferred to https://app.clickup.com/t/86ajmykh4 (append-only per-check evidence artifact with expected/actual/tolerance and source-execution evidence, plus judge provenance). The detailed CheckResult objects already exist in grading.py; persisting them per run is additive and planned before the first decision-bearing batch.

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,309 @@
...
+def run_env(home: Path, source_home: Path | None = None) -> dict[str, str]:
+    """Allowlisted environment with an ephemeral HOME. The run cannot see the
+    operator's ~/.claude (session history, memory, other projects) or any env
+    var outside the allowlist; its transcript lands inside `home`, where
+    launch_run collects it. CLI credentials are seeded in when file-based
+    (macOS keychain auth works regardless of HOME)."""
+    source_home = source_home or Path.home()
+    env = {"HOME": str(home)}
+    for k, v in os.environ.items():
+        if k in ENV_PASS or k.startswith(ENV_PASS_PREFIXES):
```
sanjayr1, line 144: **[P1] Force Claude config and temporary state into the per-run environment.**

The CLAUDE_ prefix passthrough includes CLAUDE_CONFIG_DIR. If it is set on the host, the child inherits that path unchanged; Claude Code documents that this variable overrides the default config directory and holds settings, credentials, session history, and plugins. In that case the run loads operator state, writes its transcript outside this HOME, and transcript_path looks in the wrong place. Replace the broad prefix with an exact allowlist of authentication/provider variables, explicitly set CLAUDE_CONFIG_DIR and CLAUDE_CODE_TMPDIR to per-run locations, and add a test proving a host path cannot override either. Reference: https://code.claude.com/docs/en/env-vars

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,309 @@
...
+            env[k] = v
+    # uv keys its cache off HOME by default; point the run at the real cache so
+    # it does not re-download every dependency into the ephemeral HOME.
+    cache = os.environ.get("UV_CACHE_DIR") or _uv_cache_dir()
+    if cache:
+        env["UV_CACHE_DIR"] = cache
+    (home / ".claude").mkdir(parents=True, exist_ok=True)
```
sanjayr1, line 155: **[P1] Preserve the documented Databricks U2M auth path inside the ephemeral HOME.**

The repository helper falls back to the CLI/SDK profile in ~/.databrickscfg, with the U2M cache under ~/.databricks, but this code moves HOME without seeding either. Under the standard repository setup—profile-based OAuth and no DATABRICKS_* secrets—I ran Config() using the returned environment and reproduced “default auth: cannot configure default credentials.” A smoke performed with environment credentials does not cover that path. Seed only the selected profile and token cache into the ephemeral HOME, or use an explicit credential broker/config path, and add a profile-based probe. Avoid solving this by blindly forwarding a possibly stale DATABRICKS_TOKEN.

analytics/diagnostics/quality_bench/run_matrix.py
```diff
@@ -0,0 +1,309 @@
...
+        src = transcript_path(home, run_dir, session_id)
+        if src.exists():
+            dst = batch_dir / f"{spec.run_id}.transcript.jsonl"
+            shutil.copy(src, dst)
+            out["transcript_file"] = str(dst)
+        return out
+    finally:
+        if keep_run_dir:
+            out["run_dir"] = str(run_dir)
+            out["home_dir"] = str(home)
```
sanjayr1, line 275: **[P1] Do not retain copied OAuth credentials with debug run directories.**

When keep_run_dir is true, the finally block retains the entire ephemeral HOME and returns its path. That HOME contains the copied .claude/.credentials.json, including the keychain export on macOS; I reproduced the credential file remaining after a successful kept run. The option help only promises to retain the run directory. Always remove authentication material regardless of the debug setting; if CLI state is needed, retain a scrubbed copy or require a separate explicitly unsafe opt-in with a prominent warning. Add a test asserting that no credential survives.

sanjayr1 (changes requested):
## Recalibrated review — 80/20 scaffold

The latest head, 02950b9, addresses the identifier/path checks, nested judge validation, Claude config pinning, happy-path credential scrubbing, resolution visibility, and committed no-decision documentation. I reran the full analytics suite: 190 tests pass; all GitHub checks are green.

Given the explicitly stated 80/20 scope, I no longer treat container-grade isolation, perfect one-factor arms, full evidence metadata, or stricter resolution canonicalization as blockers to merging this scaffold. They remain hard gates before a decision-bearing batch.

Two high-leverage recommendations remain:

1. **P1 security:** put run directory/HOME setup and run_env inside an outer try/finally, and make credential cleanup fail closed. A setup exception can currently occur after a secret is copied but before the existing finally is entered.
2. **P1 functional/scope:** support one Databricks authentication path end to end and preflight it, or narrow the documented support and fail fast. Copying the legacy token cache does not preserve the current keychain-backed U2M profile path.

Resolution matching is P2 while report-only. Full-process/container isolation is nonblocking for this scaffold but P1 before anyone treats the generated signals as trustworthy experimental conclusions.

analytics/diagnostics/quality_bench/grading.py
sanjayr1: I re-ran this against 8eca9b9, and the statement that cell_consistency compares the resolved value to required_resolutions is not what the current code does. Line 125 only checks that every repetition returned the same non-null value; it never compares that value with key.required_resolutions[fork]. The judge is optional, and its scoping score is not used by Rule 1. I reproduced three correct-number reps all resolving denominator as "wrong": assumptions_pass=True, cell consistent=True, and rule1_trustworthy=True. The new non-empty check is useful, but the original correctness gap remains. Please compare the observed value to the expected key value (separately from cross-rep agreement) and cover the consistently-wrong case with a regression test.

analytics/diagnostics/quality_bench/prep_arms.py
sanjayr1: Concrete implementation suggestion for DATA-2164: separate host usernames are not required if every cell has an enforced mount/process/network boundary. My preferred design is one disposable OCI container per RunSpec:

1. Build and pin one benchmark image containing the exact Claude CLI, Python, uv lock, and dependencies. Run it as a non-root UID with a read-only root filesystem, dropped capabilities, no-new-privileges, resource limits, and tmpfs-backed HOME and /tmp. Do not expose the Docker socket or a shared writable uv cache.
2. Mount only that cell’s prepared arm at /workspace. Never mount the source checkout, quality_bench/keys, the parent arms/runs directories, the operator HOME, or another cell. Capture the final answer/transcript through stdout or a unique driver-owned output channel, and use a random container identity per attempt.
3. Inject only explicit per-run secrets into tmpfs. Prefer a dedicated read-only Databricks service principal limited to the required warehouse/catalog over personal U2M credentials, plus a scoped Claude credential. Destroy the container and tmpfs on every exit path, including debug runs.
4. Put the container behind a deny-by-default egress proxy/firewall that allows only the required Anthropic/auth endpoints and the exact Databricks workspace host. No general internet egress.
5. Fail the cell closed if the container, sandbox, mount policy, or network policy cannot be established.

A lighter transitional option is to wrap the entire claude process with Anthropic’s sandbox runtime and configure deny-by-default read/write/network access, allowing only the run directory, ephemeral config directory, and required domains. The built-in Bash sandbox alone is not sufficient here because Read/Edit tools, MCP servers, and hooks remain in the host process; Anthropic explicitly recommends a full-process runtime, container, or VM for unattended sessions. Disable all unsandboxed fallback. References: https://code.claude.com/docs/en/sandbox-environments and https://code.claude.com/docs/en/devcontainer

Whichever implementation is chosen, add adversarial acceptance tests: place canaries in the original answer-key directory, real HOME, and a sibling run; prove the cell cannot read them; prove an unlisted network destination is blocked; prove the allowed Databricks query still works; prove unrelated host environment variables are absent; and prove no credential/config directory survives cleanup. Any failed isolation assertion should invalidate the batch, not merely produce a transcript warning.

Transient Unix accounts with 0700 directories could also separate local runs, but they require privileged account lifecycle management and still need separate network and secret controls. Containers or the full-process sandbox are a cleaner fit for this harness.

analytics/diagnostics/quality_bench/run_matrix.py
datawithtristan: Fixed in 02950b9: the CLAUDE_ prefix passthrough is gone (exact allowlist: CLAUDE_CODE_OAUTH_TOKEN plus DATABRICKS_*/ANTHROPIC_* for provider auth), and CLAUDE_CONFIG_DIR / CLAUDE_CODE_TMPDIR are explicitly pinned to per-run locations, with the CLI config seeded at both the HOME-derived and CLAUDE_CONFIG_DIR paths. Test asserts a host CLAUDE_CONFIG_DIR cannot override the pin. Re-verified with a live smoke under the pinned config dir (run ok, transcript collected from the per-run location).

analytics/diagnostics/quality_bench/run_matrix.py
datawithtristan: Fixed in 02950b9: run_env now seeds ~/.databrickscfg and the ~/.databricks U2M token cache into the ephemeral HOME when present, keeping the repo helper's documented profile fallback working; nothing forwards a stale DATABRICKS_TOKEN (env passthrough remains whatever DATABRICKS_* the operator already exports). Both seeded files are scrubbed on every exit path (see the keep_run_dir thread). Test covers the seeding; a profile-based live probe is a fair ask for the pre-first-batch checklist since my environment is env-var based.

analytics/diagnostics/quality_bench/run_matrix.py
datawithtristan: Fixed in 02950b9: scrub_auth() removes .claude/.credentials.json, .databrickscfg, and .databricks/ on every exit path, including --keep-run-dirs retention, so debug keeps transcripts and session state but never auth material. Tests assert nothing survives both at the scrub_auth level and through the launch_run keep path; also verified live — the kept HOME from a real smoke contains projects/sessions but no credential files.

analytics/diagnostics/quality_bench/grading.py
datawithtristan: You are right, and my previous reply misstated the code: cell_consistency checks cross-rep agreement only and never compares to key.required_resolutions — thanks for the repro. 02950b9 adds check_resolutions: each resolved fork is compared to the key value (normalized containment, since models phrase resolutions freely) and reported as a resolutions_match column in scores.csv and in the adherence section, with your agreeing-but-wrong case as a regression test. It is reported rather than gated, consistent with the earlier pre-registration ruling on adherence; the deterministic guard for a consistently wrong resolution stays the numbers, which makes tolerance-tight-enough-to-separate-forks an explicit key-review criterion for DATA-2144 (now documented in KEY_SCHEMA.md). The README also gained the no-decision gates section you asked for in the review body.

analytics/diagnostics/quality_bench/prep_arms.py
datawithtristan: Thank you for the concrete design — copied verbatim into the isolation ticket (https://app.clickup.com/t/86ajmyk2q) including the sandbox-runtime transitional option and the canary acceptance tests, so whoever picks it up starts from your spec.

analytics/diagnostics/quality_bench/run_matrix.py
sanjayr1: The file seeding is a useful improvement, but I ran the actual profile-auth path on 02950b9 and it is still broken with the repository-standard U2M setup. Inside the returned environment, `databricks auth profiles` reports the copied profile as Valid=NO, and `databricks auth token --profile dbc-3d8ca484-79f3` fails because current CLI versions no longer use the copied legacy token-cache file; the same command succeeds under the operator HOME via keychain-backed storage. The unit test only proves files were copied, not that Config/authentication works. Please make an actual Config or suppressed-output token probe the regression test. For unattended runs, the cleaner fix is a dedicated read-only M2M service principal; a transitional local fix is for the parent to obtain/refresh a short-lived token and inject only explicit host/token/profile values. Also replace the broad DATABRICKS_ prefix: it currently forwards host DATABRICKS_CONFIG_FILE paths and any stale DATABRICKS_TOKEN unchanged, contrary to the intended pinning.

analytics/diagnostics/quality_bench/run_matrix.py
sanjayr1: The normal keep path is now scrubbed, but this is not yet every exit path. `env = run_env(home)` executes before the outer `try/finally`; if setup copies a credential and then raises while writing config or copying the Databricks cache, `scrub_auth` is never called. I reproduced that by having run_env write .claude/.credentials.json and then raise: launch_run propagates the exception and leaves the credential behind. Move arm/HOME setup and run_env inside an outer try/finally, initialize the result before setup, and make cleanup fail closed while attempting every credential removal even if one unlink fails. A regression test should inject a failure after the first secret is written and assert that neither credentials nor the temporary HOME survive.

analytics/diagnostics/quality_bench/grading.py
sanjayr1: The new resolutions_match visibility is valuable, but the current containment comparison is too permissive to be a correctness check. With expected `cumulative_registered`, both `not cumulative registered` and the underspecified `registered` pass because one normalized string contains the other. At the same time, cell_consistency still compares raw strings, so `cumulative_registered` and `Cumulative (registered)` each match the key but are marked inconsistent with each other. Prefer a controlled resolution_id/enum in the output contract and exact comparison. If free text must remain, compare canonical token sequences exactly and use the same canonical representation for cross-rep agreement. Since this field remains report-only, the README/no-decision status is still essential: Rule 1 can still be true for the consistently wrong case.

analytics/diagnostics/quality_bench/prep_arms.py
sanjayr1: The new no-decision section is a strong improvement, but I would remove the Gate 2 alternative that lets written acceptance of transcript spot-checks substitute for container/full-process isolation. Transcript review is detection and sampling, not an enforced blindness boundary; accepting it in writing does not make a causal arm comparison decision-grade. Keep container/full-process isolation as the hard gate. If the team knowingly runs before it, label the batch and generated report exploratory_only and suppress the trustworthy/earns-overhead/bloat verdict labels. The same distinction applies to Gate 3: listing arm confounds is honest for exploratory results, but it cannot close a one-factor-treatment gate for a decision-bearing conclusion.

sanjayr1:
## 80/20 priority calibration

Given the intended scope, I think this is close as an exploratory benchmark scaffold. I would focus the remaining effort on two high-return items:

- **P1 security:** move all per-run setup, including run_env, under the cleanup try/finally. If setup copies a Claude or Databricks credential and then raises, the current finally is never entered. Cleanup should attempt every removal and delete the temporary HOME on any scrub failure. Detailed repro: https://github.com/thegoodparty/gp-data-platform/pull/636#discussion_r3625120156
- **P1 functional/scope:** make one Databricks auth method work end to end and add a real preflight, or explicitly document the supported method and fail fast. Current keychain-backed U2M auth is not preserved by copying the legacy token-cache file. Prefer a dedicated read-only M2M identity for unattended batches. Detailed repro: https://github.com/thegoodparty/gp-data-platform/pull/636#discussion_r3625119994

I would not block this scaffold on containerization, arm reconstruction, full metadata/evidence artifacts, or resolution enums. Those are follow-ups before the first decision-bearing batch; until then, reports should remain clearly exploratory and should not be used to approve pruning or declare the framework trustworthy. Resolution canonicalization is P2 while it remains report-only.

This keeps the 80/20 boundary practical: fix the small security failure path, guarantee one supported auth path, merge the harness, and finish scientific-validity controls before acting on its conclusions.

sanjayr1 (approved):
LGTM

PR #645, opened 2026-07-21
[DATA-2161] Load ACS 2020-2024 summary file and stage block-group estimates

sanjayr1:
## Summary

PR 1 of 2 for the ACS demographic layer: ingestion and staging only.

- Manifest-verified loader (`dbt/scripts/load_acs_summary_file.py`) for the locked ACS 2020-2024
  5-year summary-file inventory (9 shipping tables + 3 staged-not-shipped) plus the Connecticut
  block-to-planning-region crosswalk; raw all-string tables in `sandbox` via a UC volume.
  Bootstrap mode records the manifest and never touches the warehouse; a load run verifies
  sha256 + row counts against the committed manifest, preflights existing targets
  (fail-if-present unless `--allow-replace`, which records prior Delta versions as RESTORE
  targets), then publishes one table at a time with an immediate count check; any
  post-submission contact loss is reported as an unknown outcome, never a plain failure.
- New `census_acs` dbt source with provenance and the crosswalk's one-row-per-block contract
  tests.
- One staging model, `stg_census_acs__geo_estimates`: summary levels 150/040/050/160/010 +
  950/960/970, GEO_ID parsed (complete-geography rows only), the six jam codes mapped
  (controlled margins become 0, everything else null; a guard test proves the raw files publish
  no literal zero margin), semantic estimate+MOE pairs with plain root-sum-of-squares bracket
  combinations, materialized as a table (deliberate staging exception, documented in-model).
- Tests: unique key, geoid-length-by-level, sentinel range walls, published national +
  spot-state anchors (cross-checked against the Census API before pinning), exact table-shell
  column inventory, retained-key alignment, a staged-versus-raw recomputation oracle covering
  every estimate and margin pair, manifest row counts, the no-raw-zero-margin guard, and the
  Connecticut crosswalk content contract (raw-format walls before any normalization,
  bidirectional decennial bind, pure-relabel suffix anchor, coverage gate).

Deliberately NOT here (PR 2): the block-group district weights model, the allocation
intermediate, and the `district_census_stats` columns.

## Reviewer notes: two empirical source facts, encoded as structural test exceptions

1. **Aggregate income omits rows instead of jamming.** The B19025 file expresses "no published
   value" for some small geographies by omitting the row entirely, where every other loaded
   table carries a jam-code sentinel; the Census API returns null for the same places. The
   retained-key alignment test therefore keeps eleven strict arms and allows b19025 to be a
   subset that adds no keys; staged aggregate income is null on those rows with ordinary
   suppression semantics (documented in the yml). Downstream ripple: mean income must pair this
   numerator with its households denominator over the same rows (pairwise-complete).
2. **The Connecticut crosswalk omits unpopulated water-only blocks.** The state-republished
   block-level equivalency mapping excludes coastal-water blocks that carry zero decennial
   population; the corresponding ACS water block groups carry zero population and zero
   households. The decennial-side bind applies to populated blocks and the coverage gate fails
   any uncovered block group unless it provably carries no ACS mass (self-verifying from staged
   values; nulls count as populated; failure rows return geoid, population, and households).
   Nothing a weighting step could ever draw mass from is exempt.

Also worth a reviewer's eye: the loader prefers IPv4 for its downloads (documented,
unit-tested) because the Census bulk host publishes AAAA records whose route is unreachable
from some networks, leaving an unspecified-family lookup with nothing usable.

Full investigation evidence (row counts, geoids, API cross-checks) lives in the ticket working
notes; the committed artifacts deliberately carry no incidental counts.

## Verification

Manifest generated and committed BEFORE any warehouse write; the load verified every file
against it and count-verified each of the 13 tables as it was replaced; a verify-only re-run
confirmed reproducibility. Full dbt pytest suite (78 tests), the source-contract suite, a
targeted remote build, `inspect_data` review, and a single enumerated suite naming the model,
the source selector, and all nine singular tests all ran green BEFORE this PR was opened
(PASS=20/20). CI preview checks (row counts by summary level, anchor values, spot block groups
including the poverty-margin root-sum-of-squares recomputation) recorded in the ticket working
notes.

---

> [!NOTE]
> **Medium Risk**
> Large new data pipeline with nuanced Census semantics (jam codes, controlled margins, CT geography); mistakes would skew downstream district demographics, but scope is sandbox staging with heavy tests and no mart changes yet.
> 
> **Overview**
> Adds the **ingestion and staging path** for ACS 2020–2024 5-year summary-file demographics (PR 1 of 2; district allocation is out of scope).
> 
> A new **manifest-verified loader** (`load_acs_summary_file.py` + `acs5y2024_manifest.json`) downloads the locked table inventory and Connecticut block crosswalk, checks sha256/row counts before any warehouse write, and publishes raw all-string `sandbox` tables one at a time with count verification and documented Delta rollback behavior.
> 
> **dbt** gains a `census_acs` source, shared **`census_acs_*` macros** (jam-code → null/controlled-zero margins, bracket sums, retained complete-geography rows), and **`stg_census_acs__geo_estimates`** as a materialized table: one row per `(summary_level, geoid)` for block group, state, county, place, national, and school-district levels, with semantic count/MOE pairs (no precomputed shares).
> 
> **Contract tests** pin loader/staging quality: column inventory, manifest row counts, key alignment across tables (with a documented **B19025 row-omission** exception), national/CA anchor values, staged-vs-raw oracle, no literal zero margins in raw data, and Connecticut crosswalk format, decennial bind, suffix identity, and block-group coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e420a4b693a3b6b2767a9a259fe39d925717b0e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sanjayr1:
delegate review

sanjayr1:
On the jam-code asymmetry finding (delegate-finding f066a0d7): declining this one, with evidence. Per the table-based Summary File handbook's annotation table (now committed verbatim in the census_acs_values.sql header), -333333333 and -222222222 are MOE-only sentinels; the estimate-side sentinels are exactly the three the estimate macro maps. Two empirical checks on the loaded data back this up: -666666666 appears in zero margin columns across all 12 tables (its ~97k occurrences are all estimate-side in b19013/b25077/b19025, paired with the documented -333333333 MOE code), and the >= 0 range walls on every estimate column would have failed the build if -333333333 or -222222222 ever appeared estimate-side — they didn't. Posture note: if an undocumented sentinel ever lands on either side in a future vintage, we want the loud range-wall failure, not a silent null. That is now also enforced at source by the new margins-integrity test, which rejects any negative margin value outside the five documented MOE sentinels.

sanjayr1:
delegate review

dbt/scripts/load_acs_summary_file.py
```diff
@@ -0,0 +1,647 @@
...
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument("--year", type=int, default=2024, help="final year of the 5-year vintage")
+    parser.add_argument("--profile", default="dbc-3d8ca484-79f3")
+    parser.add_argument("--warehouse-id", default="18583d8b081c6486")
+    parser.add_argument("--catalog", default="goodparty_data_catalog")
+    parser.add_argument("--schema", default="sandbox")
+    parser.add_argument("--volume", default="census_acs_raw")
+    parser.add_argument("--manifest", type=Path, default=Path(__file__).parent / "acs5y2024_manifest.json")
```
sanjayr1, line 490: **Vintage/manifest mismatch:** `--year` presents this as future-vintage support, but the default manifest is permanently `acs5y2024_manifest.json`. `_run` later merges existing file entries without checking `vintage`, and verification permits extra manifest entries. Following the documented flow with `--year 2025` would therefore create a mixed 2024/2025 manifest and publish tables the dbt source/model do not declare.

Please either make this loader explicitly 2024-only (the simpler option), or derive a year-specific default manifest and reject any merge whose top-level vintage does not match. A future vintage already needs reviewed changes to model/source names and anchors, so removing false genericity is reasonable.

dbt/scripts/load_acs_summary_file.py
```diff
@@ -0,0 +1,647 @@
...
+    during the create itself is an UNKNOWN outcome, never a plain failure."""
+    replace_versions = replace_versions or {}
+    verified: list[str] = []
+    order = [targets[f.filename] for f in facts]
+    for i, f in enumerate(facts):
+        table = targets[f.filename]
+        remaining = tuple(order[i + 1 :])
+        # phase 1, pre-mutation: upload; version re-check for replacements
+        try:
+            upload(f)
```
sanjayr1, line 429: **This concurrency safeguard is more complex than the guarantee it provides.** The shared volume file is overwritten before the Delta-version recheck, and another writer can still race between `DESCRIBE HISTORY` and `CREATE OR REPLACE`; the check/replace pair is not atomic.

For these one-time, operator-owned sandbox tables, can we simplify to: plain `CREATE` by default, explicit `--allow-replace`, record the prior version for rollback, publish/count one table at a time, and stop on the first error? That would let us remove the version-drift branch and its tests while preserving the useful operational safety. If this machinery stays, its comments should avoid claiming concurrent-writer protection.

dbt/project/models/staging/census_source/stg_census_acs__geo_estimates.sql
```diff
@@ -0,0 +1,505 @@
...
+            {{ census_acs_moe("b19001_m017") }} as households_income_200k_plus_moe
+        from {{ source("census_acs", "acs5y2024_b19001") }}
+        where {{ census_acs_retained_geo_rows() }}
+    ),
+
+    -- Every table publishes the same geography set, but the spine is the union
+    -- of all twelve key sets so a single-file publication gap can never drop a
+    -- geography row silently; the retained-key alignment test pins exact
+    -- equality of the sets.
+    all_keys as (
```
sanjayr1: **Use one canonical key spine instead of unioning all twelve key sets.** The retained-key contract already says the eleven strict tables have identical retained keys and B19025 is a subset, so `b01001` can be the spine and every other table can left-join to it.

This removes the union/distinct work and makes the intended grain explicit. Please update the key test to compare every strict source directly to the B01001 keys and assert B19025 has no keys outside that canonical set. The current `b19025_keys left anti join staged_keys` arm is vacuous because the staged union necessarily contains every B19025 key.

dbt/project/tests/assert_census_acs_2024_staging_matches_raw.sql
```diff
@@ -0,0 +1,412 @@
...
+-- recomputation from the raw tables at two fixed rows: the national row and
+-- state 06. Estimates compare null-safe and exact; margins compare within
+-- 1e-6 (two floating-point paths). One-sided nulls are mismatches. The
+-- fixture-count guard at the bottom makes vacuous passes impossible: if
+-- EITHER fixture row goes missing anywhere upstream (staged, or any of the
+-- twelve raw tables feeding the inner joins), the compared set shrinks and
+-- this test fails loudly instead of comparing nothing. To see which column
+-- mismatched on a failure, run the compiled compared select without the
+-- final where.
+{% set clean_estimate = (
```
sanjayr1: **This 412-line oracle is effectively a second implementation of the model, but it exercises only two rows.** It creates substantial lockstep maintenance and can still pass correlated mistakes when the same mapping is updated incorrectly in both places.

Please replace it with a compact fixture test for the value macros (all sentinels, controlled-zero MOE, null propagation, and one RSS combination), keep the published national/California anchors, and add concise all-row semantic invariants. High-value invariants include population/race universe equality, race components summing to their universe, household universes agreeing, income brackets summing to their universe, and numerator bounds for education, employment, poverty, tenure, and broadband.

dbt/project/tests/assert_census_acs_2024_raw_margin_integrity.sql
```diff
@@ -0,0 +1,50 @@
...
+-- never publish a literal zero margin -- so a raw 0 fails here. (2) Sentinel
+-- authority guard: the staging macros map exactly the five documented margin
+-- sentinels, so ANY other negative margin value (an undocumented sentinel --
+-- for example an estimate-side code leaking into a margin column) fails here
+-- at the source, before root-sum-of-squares could ever square it into a
+-- plausible-looking number. Deliberate fail-loud posture: no speculative
+-- mappings for codes the authority does not document. If this ever fails, the
+-- mapping decision goes back to review before anything consumes staged
+-- margins.
+{% set expected_e_columns = {
```
sanjayr1: **Please narrow this to the margin columns the staging model actually consumes.** Scanning every margin column in every table repeatedly protects unused cells in a static, hash-locked source and duplicates the table-shell inventory.

The useful property is that an unrecognized negative margin cannot be squared by RSS into a plausible positive value. We can preserve that by checking only consumed margin inputs, or by making the MOE transformation fail immediately on an unrecognized negative value. Either approach keeps the fail-loud behavior with a much smaller contract.

dbt/project/tests/assert_census_acs_2024_source_columns.sql
```diff
@@ -0,0 +1,80 @@
+{{ config(severity="error") }}
+
+-- Exact raw-table column contract, generated from the published 2024 5-year
+-- table shells: geo_id plus interleaved estimate/margin pairs 001..N per
+-- table, every column string-typed (the raw layer is all-string by
+-- convention; casting happens in staging). Flags missing columns, unexpected
+-- columns, and non-string types by NAME, so a restructured source file or a
+-- wrong load can never masquerade as suppression downstream.
+{% set expected_e_columns = {
```
sanjayr1: **This exact shell-wide column inventory appears redundant with the locked file hashes and the staging references.** A missing consumed column already fails the model, unexpected unused columns do not affect it, and the loader itself controls the all-string read.

Please remove this test or reduce it to the columns actually required by staging. If the exact published shell is valuable as investigation evidence, it can live in the manifest/ticket rather than as a recurring warehouse contract.

dbt/project/tests/assert_census_acs_ct_suffix_identity.sql
```diff
@@ -0,0 +1,13 @@
+{{ config(severity="error") }}
+
+-- Independent pure-relabel anchor for the Connecticut mapping: the 2022
```
sanjayr1: **Consolidate the three Connecticut singular tests without weakening the checks.** The suffix identity, crosswalk content contract, and populated block-group coverage all operate on the same crosswalk/staged relations.

Please fold suffix identity and the reverse populated-coverage condition into `assert_census_acs_ct_crosswalk_contract.sql` as additional named violation CTEs, then remove the two standalone files. Keeping the violation labels will preserve diagnostics while reducing repeated setup and test-selection bookkeeping.

dbt/tests/test_load_acs_summary_file.py
```diff
@@ -0,0 +1,356 @@
...
+    decide_action,
+    preflight_existing_targets,
+    publish_tables,
+    sha256_of,
+    stream_download,
+    verify_against_manifest,
+)
+
+
+def test_acs_table_inventory_is_locked():
```
sanjayr1: **Trim tests that only restate constants or exercise the standard library.** This exact inventory assertion changes whenever `ACS_TABLES` changes but does not validate behavior; the manifest/source contract already exposes inventory drift. The JSON round-trip and standalone `hashlib` equivalence tests are similarly low-value.

Please keep the boundary tests that protect actual loader behavior: row counting with/without a trailing newline, BOM stripping, manifest mismatch blocking publication, SQL read options, happy-path publication/count verification, stop-on-failure, and unknown post-submission outcome handling.

sanjayr1 (commented):
Complexity/consolidation review for the current 2024 ingestion and staging scope.

Please keep all 12 ACS tables, including B19001, B19013, and B25077; those staged-for-later tables are intentional. The inline comments focus on simplifying the loader and tests without removing that future-facing data contract.

Highest-value changes:
1. Make the loader honestly 2024-specific or make manifest selection vintage-safe.
2. Use one canonical geography spine.
3. Replace duplicated implementation-oracle coverage with compact fixtures and semantic invariants.
4. Remove or narrow safeguards that re-prove the locked static source rather than plausible application failures.
5. Consolidate the Connecticut contract tests without weakening their populated-coverage guarantees.

sanjayr1:
delegate review

sanjayr1:
On the chained-AND null-semantics finding (delegate-finding b937b923): declining, with the engine's own answer. Under SQL three-valued logic FALSE dominates AND: a negative value in any non-null column forces the whole expression to FALSE, NOT(FALSE) = TRUE, and expression_is_true returns the row. NULL only survives the chain when no term is FALSE, i.e. when every non-null column already satisfies its bound — which is the intended suppression semantics (null means the Census suppressed the value, not an invalid sentinel). Verified on the warehouse: select not(cast(null as bigint) >= 0 and -5 >= 0) returns true (row flagged), and the all-valid-with-null case returns null (row passes). The coalesce(col >= 0, true) rewrite is therefore semantically identical to the current expression on every input, and adds 42 wrapper terms to a test this PR just simplified. The range walls catch every negative non-null value on every row, including rows carrying structural nulls from b19025 omissions or suppressed composites.

sanjayr1:
delegate review

sanjayr1:
delegate review

PR #646, opened 2026-07-21
people-api-loader: add PostGIS point column + GiST index on Voter

danpelota:
## What

After the loader finishes loading, add a geospatial column and index to the partitioned `Voter` table.

- New `geom GEOMETRY(POINT, 4326)` column on `Voter`, a STORED generated column derived from `Residence_Addresses_Longitude`/`Latitude`. STORED means the single `ALTER` populates it, no separate `UPDATE` pass.
- New GiST index `Voter_geom_idx` on `geom`, added to `_serving_seed_extra.EXTRA_INDEXES` so it builds through the existing per-partition parent-only + child-attach machinery and lands in the manifest.

## Why GiST

Standard PostGIS point index: fast bulk build and fast point/bbox lookups. BRIN builds smaller/faster but only gives quick lookups on spatially-sorted data, which isn't guaranteed here. The index is built once and never rebuilt.

## Notes

- Idempotent: `IF NOT EXISTS` on the extension, column, and index keeps the step re-runnable.
- `geom` is loader-added and not in the Prisma serving schema (Prisma ignores unknown columns). Exposing it to the app would be a separate Prisma change.

---

> [!NOTE]
> **Medium Risk**
> Adds PostGIS DDL and a partition-wide ALTER on the large Voter table during index build, which can extend loader runtime and must succeed before GiST indexes build; changes are idempotent and confined to the loader pipeline, not app auth or serving Prisma schema.
> 
> **Overview**
> Adds **geospatial serving support** on the partitioned `Voter` table after load, without changing the Prisma schema.
> 
> During **`build_indexes`**, the step now enables **PostGIS**, applies session settings on that connection, and adds a **STORED generated** `geom` column (`GEOMETRY(POINT, 4326)`) from `Residence_Addresses_Longitude` / `Latitude` before the normal index loop runs. **`Voter_geom_idx`** is registered in **`EXTRA_INDEXES`** as a **GiST** index on `geom`, so it builds with the existing per-partition parent/child index machinery.
> 
> New tests assert PostGIS + `ALTER TABLE` DDL during `run()` and that the extra index definition uses GiST on `geom`. Extension, column, and index creation remain **idempotent** via `IF NOT EXISTS`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 021c93daacded383ee695ff337dce94562ad4602. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danpelota:
delegate review

danpelota:
delegate review

hhkarimi (dismissed):

danpelota:
delegate review

PR #658, opened 2026-07-22
Loader: provisioned serving + astro cache-bust to redeploy merged loader

hhkarimi:
Combines two changes needed before the next dev-cluster rebuild:

## 1. Serve on a provisioned instance class (not Serverless v2)
The `resize` step (step 6) now flips the served writer to a provisioned instance class instead of Serverless v2:
- `serve_instance_class` config, env `LOADER_SERVE_INSTANCE_CLASS`, **prod default `db.r6g.4xlarge`**; astro-dev sets `db.t4g.medium`.
- New `flip_writer_to_provisioned` in `core/aws.py` (instance-level `modify_db_instance` + `wait_instance_class_applied`; no cluster-level ServerlessV2 config).
- `ResizeManifest` drops `min_acu`/`max_acu`, keeps `final_instance_class`.
- The on-failure cost guard (`scale_down`) is unchanged and still flips to cheap Serverless v2 to hold a failed cluster — its ACU fields were renamed to `scale_down_min_acu`/`scale_down_max_acu` (env `LOADER_SCALE_DOWN_*`) to keep the two concerns distinct.
- Loader DAG doc updated to match.

Load/index phases are untouched (still the big `db.r8g` classes), so no load-perf impact.

## 2. Astro cache-bust bump
Bumps the `requirements.txt` cache-bust token so `astro deploy` reinstalls the loader `@main` — picking up the merged enum (#657), the `hf_most_important_policy_item` column, the green→public compat views, and the provisioned-serving change above.

Tests: full loader suite 310 passed + integration suite 4 passed; ruff/format/ty clean.

### DIFF
--- a/airflow/astro/docs/people_api_loader.md
+++ b/airflow/astro/docs/people_api_loader.md
@@ -13,8 +13,8 @@ data-loading revamp). Latest work is on the active feature branch (currently
 
 The loader is a Python CLI (`people-api-loader/`, `loader <step> --date <ds_nodash>`) that follows a
 train-deployment model: every refresh provisions a brand new Aurora cluster, loads and indexes it,
-resizes it to Serverless v2, and (at cutover) swaps it in as the serving database. The existing
-serving cluster is never mutated in place.
+resizes it to its serving instance class, and (at cutover) swaps it in as the serving database. The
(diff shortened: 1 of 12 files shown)


hhkarimi:
delegate review

PR #659, opened 2026-07-22
Loader: read inspect_prod baseline from the connection string's schema

hhkarimi:
## Why
The dev rebuild's `load_people_api` run failed at step 0 `inspect_prod` with `invalid URI query parameter: "schema"`. Root cause: the dev baseline SSM connection string (`people-db-connection-string-dev`) was changed on 2026-07-21 to a Prisma-style URL pointing at swain-db's `green` schema (`...?schema=green`). libpq/psycopg has no `schema` query param, and `inspect_prod` also hardcoded the `public` schema — so the loader couldn't read the intended `green` baseline.

`inspect_prod` reads the baseline with read-only `SELECT count(*)`; swain-db is never mutated.

## What
- **db.py:** new `_conninfo_from_ssm` strips a `?schema=<name>` param out of the URL and translates it to the libpq-native `options=-c search_path=<name>`. Connection strings without `?schema=` are byte-for-byte unchanged (no `options` added). The schema value is validated as a bare identifier before ever building the options string (no injection). Both `connect_prod` and `connect_new` route through it.
- **inspect_prod.py:** count queries drop the hardcoded `public.` (resolve via `search_path`); `_has_column` uses `current_schema()` instead of `table_schema='public'`.
- **serving_structure.py:** `extract_indexes` (advisory index-drift check) uses `current_schema()`; the PK/FK extractors (not on the baseline path) are untouched. cli.py's `extract-serving-structure` is unchanged in normal use (default search_path → `public`).
- **requirements.txt:** cache-bust bump so `astro deploy` reinstalls the loader.

## Tests
23 tests across `test_connections.py` (new `_conninfo_from_ssm` cases incl. injection rejection) and `test_inspect_prod.py` (unqualified SQL + `current_schema()`). Full suite 318 passed / 4 skipped; integration suite ran 4/4 against local pg; ruff/format/ty clean.

### DIFF
--- a/airflow/astro/requirements.txt
+++ b/airflow/astro/requirements.txt
@@ -15,7 +15,7 @@ sshtunnel>=0.4.0,<1.0.0
 # post-merge. (During review this tracked the DATA-2100 feature branch so astro-dev ran the PR's loader;
 # repinned to @main for merge — it resolves to this PR's loader once merged.) See PR #607.
 # cache-bust: bump this token whenever the loader source or this pin changes so the image reinstalls
-# (the install layer is keyed on this file; an unchanged file = a stale loader). bust=2026-07-22.1
+# (the install layer is keyed on this file; an unchanged file = a stale loader). bust=2026-07-22.2
 people-api-loader @ git+https://github.com/thegoodparty/gp-data-platform.git@main#subdirectory=people-api-loader
 # The voter-data gate runs in dbt Cloud (DbtCloudRunJobOperator) rather than locally — dbt cannot
 # run on the image's Python 3.14, and the full monorepo project needs ~24 parse-time env vars. This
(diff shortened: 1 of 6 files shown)


hhkarimi:
Closing in favor of fixing the root cause at the source. The break was a hand-edited SSM value (`people-db-connection-string-dev` got a Prisma-style `?schema=green` on 2026-07-21); we reverted that param to no schema (reads swain-db `public`, the proven config that built the 7-16 and 7-20 clusters). With the baseline on `public` by decision, the schema-honoring code here is for a path we won't use (YAGNI).

For the record, both delegate-reviewer findings were valid: (1) the options/search_path collision would silently drop a URL-supplied `options=`, and (2) changing only `extract_indexes` to `current_schema()` left `extract_primary_keys`/`extract_foreign_keys` on hardcoded `public`, an inconsistency `extract-serving-structure` would hit. If we ever deliberately move the baseline to `green`, we'll rebuild this completely and consistently.

PR #665, opened 2026-07-22
[DATA-2175] Backfill zip coverage for LLM-unmatched override positions

danpelota:
## Summary

The `override_zip_to_br_office` injection in `int__zip_code_to_br_office` only ran for override positions **absent** from the LLM match snapshot. Positions that are **present but unmatched** (`is_matched = false`) carry a `NOT_MATCHED` l2_district, so the match-driven path places them nowhere, and the override path excluded them because they have a snapshot row. The result: the position gets a district in `m_election_api__position` but no zip rows, so the office is invisible to the officepicker (empty `positions/search?zip=` and empty `getZipCodesByBrPositionId`).

This surfaced as a missing office: KY District Court Judge, 51st District (Henderson County), br_database_id 86432/86433.

## Change

- Scope the backfill exclusion to genuinely matched positions (`is_matched`), so unmatched overrides flow through `override_zip_to_br_office`.
- Widen `assert_override_positions_have_zip_coverage` to match. The test previously used the same `br_database_id not in (snapshot)` filter and shared the exact blind spot, so it passed despite the gap.

No collision risk: every `is_matched = false` override position has `l2_district_type = 'NOT_MATCHED'`, so it produces zero rows in the match-driven path and cannot duplicate the unique key.

## Verification

Built `int__zip_code_to_br_office` and reran its tests (18 pass, including the widened coverage test). Zip rows now generated:

| br_database_id | position | zips |
| --- | --- | --- |
| 86432 | KY District Judge, District 51, Div 1 | 10 (Henderson) |
| 86433 | KY District Judge, District 51, Div 2 | 10 (Henderson) |
| 86451 | KY District Judge, District 8, Div 1 | 10 (Warren) |
| 413282 | KY District Judge, District 22, Div 6 | 24 (Fayette) |

59 of 65 KY override positions were affected; likely other states too.

## Test plan
- [ ] CI dbt build + tests green
- [ ] After marts sync, `getZipCodesByBrPositionId` returns Henderson zips for 86432/86433
- [ ] Office appears when impersonating a Henderson County voter file

---

> [!NOTE]
> **Medium Risk**
> Changes which BallotReady positions get zip rows in a core election/officepicker path; scope is narrow and matched rows are explicitly excluded to avoid unique-key collisions.
> 
> **Overview**
> **`override_zip_to_br_office`** in `int__zip_code_to_br_office` now treats override positions as already handled by the LLM path only when they have a **genuinely matched** snapshot row (`is_matched`), not merely any snapshot row. Override seeds for positions that appear in the LLM table but are **unmatched** (`NOT_MATCHED` districts) are included again, so they get zip→office rows from the override seed instead of zero coverage.
> 
> **`assert_override_positions_have_zip_coverage`** uses the same `is_matched` filter and renames the CTE to `llm_unplaceable_overrides`, so the test catches the same gap the model fix addresses. Comments and the override `llm_reason` string are updated to describe both absent-from-snapshot and unmatched-in-snapshot cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73037ffc081d275ab23680ddef603b5e2616d140. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Deploy note

This change corrects 54 override positions the LLM matched to the wrong district (e.g. Sioux Falls ward seats matched to the whole city). Their stale LLM-district rows must be purged with a one-time `dbt build --full-refresh --select int__zip_code_to_br_office` on deploy; an incremental merge cannot delete them.

### DIFF
--- a/dbt/project/models/intermediate/l2/int__zip_code_to_br_office.sql
+++ b/dbt/project/models/intermediate/l2/int__zip_code_to_br_office.sql
@@ -102,11 +102,13 @@ with
     ),
 
     -- Same zip->office linkage as above, but built from the override seed for
-    -- positions absent from the LLM snapshot (the match-driven path keys off
-    -- the snapshot, so they would otherwise get zero zip rows). Scoped to
-    -- snapshot-absent overrides, so existing overrides are untouched and the
-    -- unique key cannot collide. Reads the unfiltered zip->district source so a
-    -- new override backfills without a full refresh.
+    -- positions the LLM path can't place: those absent from the snapshot and
+    -- those present but unmatched (is_matched = false, which carry a
+    -- 'NOT_MATCHED' l2_district and so join to no zip district). Both would
+    -- otherwise get zero zip rows. Scoped to exclude only genuinely matched
+    -- positions, so LLM-matched rows are untouched and the unique key cannot
+    -- collide. Reads the unfiltered zip->district source so a new override
+    -- backfills without a full refresh.
     override_zip_to_br_office as (
         select
             tbl_zip.zip_code,
@@ -124,7 +126,7 @@ with
             tbl_override.l2_district_name,
             tbl_override.l2_district_type,
             true as is_matched,
-            'l2_br_match_overrides seed (no LLM match row)' as llm_reason,
+            'l2_br_match_overrides seed (no LLM match)' as llm_reason,
             null as confidence,
             null as embeddings,
             null as top_embedding_score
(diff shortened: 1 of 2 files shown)


danpelota:
delegate review

danpelota:
Fixed. Added the `is not null` guard on the `covered` subquery per the repo NOT IN convention.

delegate review

PR #670, opened 2026-07-23
Loader: run validate before resize + env-configurable build_indexes parallelism

hhkarimi:
Two bundled loader improvements (both future-build; no effect until deployed).

## 1. Run `validate` before `resize`
`resize` flips the writer down to the small serving instance (dev `db.t4g.medium`), and `validate` then runs heavy per-state `count`/`GROUP BY` over ~227M rows on 2 vCPU — very slow (observed ~60 min, I/O-bound, no post-load VACUUM so no index-only scans). Reordering to `build_indexes >> validate >> resize` runs validate on the `r8g.48xlarge` index instance that's already scaled up — fast, parallel — then `resize` is the clean final step.

`validate` only checks row counts + schema/index structure, none of which change across `resize`, so correctness is identical either order.

**Safety fix:** `scale_down_on_failure` (trigger_rule=one_failed) now includes **`validate`** in its upstreams. With validate running before resize on the big `r8g.48xlarge` instance, a validate failure must flip the writer to `db.serverless` — otherwise it strands the expensive instance (exactly the cost bleed we hit manually).

## 2. `LOADER_INDEX_PARALLELISM` env knob
`build_indexes` runs 128 concurrent builders over one shared bastion tunnel (128 channels), which is fragile against the bastion sshd's session limits. This adds a `LOADER_INDEX_PARALLELISM` env override (default **128 = current behavior when unset**); the CLI `--parallelism` flag still overrides. Lets us dial concurrency down per-deployment (e.g., dev) without a code change if the bastion is constrained.

## 3. Minor
- Light post-resize `SELECT 1` connectivity smoke check (resize is now the final step) — confirms the resized/rebooted writer is queryable; not the heavy counts.
- Docs (`people_api_loader.md`): step order, scale_down note, `LOADER_INDEX_PARALLELISM`.
- Cache-bust bump.

## Tests
people-api-loader: 316 passed / 4 skipped + integration 4 passed; ruff/format/ty clean. airflow: 82 passed / 2 skipped — updated the pre-existing DAG-order tests (`test_load_people_api_sequence`, `test_load_people_api_scale_down_on_failure`) to the new graph, asserting validate-before-resize and validate in the scale_down upstreams.

### DIFF
--- a/airflow/astro/dags/load_people_api.py
+++ b/airflow/astro/dags/load_people_api.py
@@ -117,27 +117,35 @@ def load_people_api():
     create_schema = _step("create_schema", "create-schema")
     copy = _step("copy", "copy", extra_args=f"--parallelism {_COPY_PARALLELISM}")
     build_indexes = _step("build_indexes", "build-indexes")
-    resize = _step("resize", "resize")
     validate = _step("validate", "validate")
+    resize = _step("resize", "resize")
     # Cost guard: a failed/aborted run after provision strands whatever writer instance class was
-    # in effect (up to db.r8g.48xlarge post build_indexes) because `resize` never runs. Fires
-    # (trigger_rule=one_failed) if any resource-creating task fails, flipping the stranded writer to
-    # db.serverless WITHOUT resize's serve lockdown (no param-group swap, backup retention,
-    # deletion protection, or reboot) so the cluster + loaded data stay around for a resumed retry
-    # or forensics. Skipped when the run succeeds (resize already made the writer serverless).
+    # in effect (up to db.r8g.48xlarge post build_indexes, the same box validate now runs its heavy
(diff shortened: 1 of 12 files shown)


hhkarimi:
delegate review

Addressed the blocker: `build_indexes` parallelism < 1 (bad `LOADER_INDEX_PARALLELISM` or an explicit `--parallelism 0`) is now clamped to at least 1, so it can't spawn zero builders, silently skip every index, and still write a "complete" manifest. Guard added at two layers — `LoaderConfig.from_env()` (env path) and `build_indexes.run()` (covers the CLI `--parallelism` and programmatic paths the finding noted were unguarded) — plus a parametrized config test for 0 and negative values.

PR #676, opened 2026-07-23
Loader: VACUUM (ANALYZE) after build_indexes instead of plain ANALYZE

hhkarimi:
## Why
`build_indexes` runs a plain `ANALYZE public."<table>"` per table. On a freshly bulk-loaded table the **visibility map** is empty, so `count(*)` and index-only scans fall back to full heap scans — which is why `validate` crawls on the small serving instance and why serving-side broad-filter counts are slow.

`ANALYZE` only refreshes planner stats; **`VACUUM` sets the visibility map** (enabling index-only scans). `VACUUM (ANALYZE)` does both in one pass, so it fully subsumes the current `ANALYZE`.

## What
- `build_indexes`: `_analyze` → `_vacuum_analyze`, running `VACUUM (ANALYZE) public."<table>"` per table, in the same spot (after that table's indexes, on the `r8g.48xlarge` index instance, before `resize`) so the one-time full scan is paid on the fast box.
- `VACUUM` requires autocommit (can't run in a transaction block); `connect_new` defaults `autocommit=True`, which this relies on — documented in the helper's docstring and now **DB-verified** by an integration test that drives `_vacuum_analyze` against real Postgres under autocommit.
- Docs (`people_api_loader.md`) updated; cache-bust bump.

## Tradeoff
`VACUUM (ANALYZE)` full-scans each table (vs `ANALYZE`'s sample), adding wall-clock to `build_indexes` — a one-time build cost on the big instance instead of paying full heap scans on every serving-side count forever.

## Tests
people-api-loader: 318 passed / 5 skipped + integration 5 passed (incl. the new `VACUUM (ANALYZE)`-against-real-PG test); ruff/format/ty clean.

### DIFF
--- a/airflow/astro/docs/people_api_loader.md
+++ b/airflow/astro/docs/people_api_loader.md
@@ -55,7 +55,7 @@ structure — none of which change across `resize` — so correctness is identic
 | `provision` | Create a fresh Aurora cluster + writer on the **copy-phase** instance class, empty cluster parameter groups, connection string to SSM. Reuses the shared S3 gateway VPC endpoint. |
 | `create_schema` | Apply CREATE TABLE statements from the committed snapshot for all four tables: `Voter` and `DistrictVoter` (partitioned by State with per-state LIST children) and `District` and `DistrictStats` (flat). |
 | `copy` | Parallel server-side `aws_s3.table_import_from_s3`: for partitioned tables (Voter, DistrictVoter), per-state per-file copy with rows routed to partitions by `"State"`; for flat tables (District, DistrictStats), whole-table copy. Idempotent per state/table (count / DELETE + reload). |
-| `build_indexes` | Scale the writer up to the **index-phase** class, then add primary keys and indexes for all four tables: for partitioned tables (Voter, DistrictVoter) build per-partition indexes and attach to parent; for flat tables (District, DistrictStats) build parent-level indexes. Then `ANALYZE` all tables. Concurrent-builder count defaults to `LOADER_INDEX_PARALLELISM` (else 128). See below. |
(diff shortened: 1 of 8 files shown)


hhkarimi:
delegate review

Fixed: added `DROP TABLE IF EXISTS public."VacTgt" CASCADE` before the CREATE, matching the pattern the other integration tests use — so a second run against a persistent `LOADER_TEST_PG_DSN` DB won't collide. Verified by running the test twice consecutively (both pass).

hhkarimi:
Added a second fix to this PR, from the first refresh's `validate` results: `schema_diff_clean:Voter` was failing because `hf_most_important_policy_item` (the mart column this whole refresh adds) is absent from the swain prod baseline and wasn't in the schema-diff allowlist. Added it to `LOADER_ADDED_COLUMNS["Voter"]` (broadened the comment to cover both loader-created columns like `geom` and newly-projected mart columns ahead of the baseline), with a matching `_check_schema_diff` test. So the next build's schema-diff goes green; the VACUUM (ANALYZE) change in this PR addresses the slow-count side (which is the likely cause of the other failing check, `sample_queries_pass`).

hhkarimi:
delegate review

hhkarimi:
delegate review

Fixed the tautology. The VACUUM test now spies on `connect_new` and asserts `_vacuum_analyze` never passes `autocommit=False` (captured kwargs), on top of executing VACUUM (ANALYZE) against the real autocommit `pg_conn`. And to cover the gap you flagged (no test pinned `connect_new`'s default), added `test_connect_new_defaults_to_autocommit` asserting its `autocommit` default is `True` via signature introspection — so a silent flip to `False` fails a unit test. 320 pass, lint/ty clean.

PR #680, opened 2026-07-23
[DATA-2144] quality_bench: gold-run answer key q06 (serve-win overlap)

datawithtristan:
## What
Adds the graded answer key for gold-run holdout **q06** (Serve/Win overlap) to `analytics/diagnostics/quality_bench/keys/`. One file; no runtime code touched (keys are grader-side only and stripped from run worktrees).

## Numbers (as of 2026-06-30, verified by direct SQL)
| metric | value | tol |
|---|---|---|
| serve_users_with_win_candidacy | 1833 | 0.5% |
| pct_serve_users_with_win | 99.7 | 0.4% |
| pct_win_users_who_are_serve | 3.08 | 5.0% |

Population: Serve = `users_serve_base.is_serve_user` + `eo_activated_at < 2026-07-01`; Win = distinct `user_id` in `users_win_candidacy` (`is_latest_version AND NOT is_demo`) + `user_created_at < 2026-07-01`; internal `@goodparty.org` excluded on both sides; person-grain intersection on product `user_id`.

## Key design
- **Count is the discriminator.** The 0.5% tolerance fails the internal-filter fork (1858), the stale doc figure (1860), and skip-both (1867); both percentages are near-invariant across forks, so those forks are guarded by `required_resolutions` + judge, not the numbers.
- **Doc-figure trap.** The serve-knowledge doc states "1,860 of 1,869 ... verified 2026-07-14"; this run computed the overlap directly and reconciled (current-state 1,867/1,876, +7/+7 over the 9-day gap) rather than accepting it.
- severity1 tripwires cover inversion and low-serve-overlap; deliberately none on the doc figure or the independent-recruitment over-claim, since those share vocabulary with required correct prose (the judge grades them).
- Reviewed by both pipeline reviewers; one narrative fix applied (the residual-recency claim was corrected: 1 of 5 unattached Serve users is in the newest month, N too small to detect skew).

---

> [!NOTE]
> **Low Risk**
> Benchmark grading metadata only; no production paths, auth, or data pipelines are modified.
> 
> **Overview**
> Adds **`q06_serve_overlap_key.yaml`** for the quality_bench gold-run holdout on Serve/Win user overlap as of June 30, 2026. Grader-only configuration; no application or harness code changes.
> 
> The key pins three metrics (**1833** overlap count at **0.5%** tolerance, **99.7%** / **3.08%** percentages with wider bands) and documents why the **count** is the main discriminator against common forks (skipping internal filters, stale doc **~1,860**, wrong denominators). It locks **`required_resolutions`**, **`mandatory_sources`** (`users_serve_base`, `users_win_candidacy`, as-of anchors), **`severity1_patterns`** for gross metric inversions, and an **`intent_card`**—with explicit notes on decay/re-baseline and why some traps are left to the judge instead of regex.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2308104b4f5b2b5dc95e2259a9d9c9f878f5b31a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/analytics/diagnostics/quality_bench/keys/q06_serve_overlap_key.yaml
+++ b/analytics/diagnostics/quality_bench/keys/q06_serve_overlap_key.yaml
@@ -1,0 +1,102 @@
+id: q06_serve_overlap
+as_of: "2026-07-23"
+numbers:
+  # DISCRIMINATOR = the count. The two percentages are near-invariant across every
+  # legitimate fork (pct_serve 99.5-99.7; pct_win 3.0-3.1), so they confirm shape but
+  # cannot guard scoping — the count does. Near-miss ladder (verified 2026-07-23,
+  # ladder2.py): MAIN (as-of + internal-excluded) = 1833; skip internal filter = 1858
+  # (+1.36%); skip as-of cut = 1840 (+0.38%); skip both = 1867 (+1.85%); stale doc
+  # figure serve-analytics-knowledge/sources.md = 1860 (+1.47%); conflating the Serve
+  # population (1838) for the overlap = +0.27%.
+  # TOLERANCE 0.5% (+-9 users, 1824-1842): fails the internal-filter fork (1858), the
+  # stale doc figure (1860), and skip-both (1867) — the answers this question exists to
+  # catch. Deliberately does NOT numerically catch the as-of-skip (1840) or the
+  # population-conflation (1838); both are guarded by required_resolutions + judge, not
+  # the number (they are <0.4% away and a tolerance that caught them would fail on
+  # ordinary mart drift).
+  # DECAY NOTE: as-of-bounded but built on current-state flags (is_serve_user, is_demo,
+  # is_latest_version) applied to historical timestamps, so a correct re-run can only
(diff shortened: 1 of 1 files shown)


datawithtristan:
Both blockers addressed in 027032d:

1. **pct_serve ceiling false-pass** — tightened `tolerance_pct` 0.4 → 0.3. Confirmed against `grading.py` (`check_numbers` uses a raw relative diff, no percentage clamp), so 0.4% (upper 100.10) would have passed a run reporting exactly 100.0 and rounding away the 5-user residual the intent_card calls out. At 0.3% the window is [99.40, 100.00): 100.0 fails (diff 0.30%) while every legit fork value (99.52–99.73) still passes.

2. **as-of source check too weak** — anchored `win_asof_anchor` to the July-1 boundary token adjacent to the column, per the suggestion. Extended the same fix to `serve_asof_anchor`, which had the identical weakness (`check_sources` is a full-transcript IGNORECASE search; both timestamp columns appear in essentially every transcript regardless of whether an as-of cut was applied). Verified the new patterns fire on a real filter expression and stay silent on a bare column mention.

delegate review

PR #698, opened 2026-07-27
Drive loader environment from a single ENVIRONMENT var

hhkarimi:
## Why

The loader used two vars for environment — `LOADER_ENV` (SSM connection-string param names + dated cluster ids) and `LOADER_TAG_ENVIRONMENT` (the `Environment` tag on created resources) — both defaulting to `dev`. Neither is set on the Astro deployments, so the **prod** deployment silently ran as `dev`: `inspect_prod` read `people-db-connection-string-dev` and was denied (the prod role only grants the `-prod` param), and `provision` would later be denied too (the prod create policy requires `RequestTag/Environment=prod`).

Read the single `ENVIRONMENT` var instead. It's already set per Astro deployment (`dev` on astro-dev, `prod` on astro-prod) and matches how the DAGs otherwise differentiate environments (per-deployment Variables/Connections). One value now keys the SSM param names, the dated cluster identifiers, and the `Environment` tag, so they can't drift.

No new Astro variables needed — deploying this makes the prod loader resolve to `prod` on its own.

Follow-up after merge: deploy to astro (dev + prod), then re-run the prod `load_people_api` DAG.


hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

PR #700, opened 2026-07-28
Bump loader cache-bust to reinstall the ENVIRONMENT loader

hhkarimi:
Follow-up to #698. The astro image installs the loader from `@main` and caches that pip layer on `requirements.txt`, so the ENVIRONMENT change on main won't reach the image until this file changes. Bump the `bust=` token (and drop the stale review-time note on the pin).


hhkarimi:
delegate review

PR #703, opened 2026-07-28
Add a final database-wide ANALYZE step (fresh per-partition planner stats)

hhkarimi:
## Why

`build_indexes` runs `VACUUM (ANALYZE)` on each partitioned parent (`Voter`, `DistrictVoter`). The VACUUM recurses to the leaf partitions and sets their visibility maps (confirmed: partitions are 100% all-visible, index-only scans work), but the parent-level **ANALYZE only refreshes the parent's inheritance statistics** — it does not populate each leaf partition's own per-column stats. Verified on the freshly-built prod cluster: every `Voter_XX` / `DistrictVoter_XX` partition had `last_analyze IS NULL AND last_autoanalyze IS NULL` after a full successful run, so the planner was serving partition scans on default selectivity.

## What

- New final step `analyze` (CLI `loader analyze`) that runs a **bare, database-wide `ANALYZE`** — one statement that processes every table *and* every leaf partition the connecting role owns, which the per-parent ANALYZE never reached. Idempotent + manifest-guarded like the other steps.
- Wired as the last step: `… build_indexes >> validate >> resize >> analyze`, and added to the `scale_down_on_failure` cost-guard upstreams.
- Fixed the `build_indexes` comment that claimed the parent `VACUUM (ANALYZE)` "recurses to every partition automatically" — true for the VACUUM/visibility-map, false for the ANALYZE/stats.

Manually running `ANALYZE;` on the prod cluster dropped the large-district (Dallas, 676k rows) count to ~1.35s warm with a correct nested-loop/index plan and 0 heap fetches.

### DIFF
--- a/airflow/astro/dags/load_people_api.py
+++ b/airflow/astro/dags/load_people_api.py
@@ -119,6 +119,12 @@ def load_people_api():
     build_indexes = _step("build_indexes", "build-indexes")
     validate = _step("validate", "validate")
     resize = _step("resize", "resize")
+    # Final step: a database-wide ANALYZE. build_indexes' per-parent VACUUM (ANALYZE) sets the
+    # leaf partitions' visibility maps but leaves their per-partition planner stats empty
+    # (parent-level ANALYZE only does the parent's inheritance stats), so the planner serves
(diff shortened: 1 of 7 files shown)


hhkarimi:
delegate review

PR #704, opened 2026-07-28
Automated serving cutover: promote step (overwrite-latest + build-{date} label)

hhkarimi:
## Why

We are collapsing the per-build dated connection-string parameters into one serving parameter per environment, `people-db-connection-string-{env}`, that each refresh overwrites. That gives one stable entry for people-api to read.

`promote` copies a validated run's connection string into that serving parameter, creating a new version. people-api reads the parameter's **latest** version and revalidates every ~5 minutes (hot-swapping its Prisma client), so the overwrite itself is the cutover — no restart, and no `live` pointer or people-api change needed (settled in the team thread).

The new version also gets a best-effort `build-{date}` label: an operator-visible anchor mapping a version to its cluster. SSM caps labels at 10 **per version**, and we apply exactly one per version, so this extends indefinitely. Labeling is best-effort — the overwrite already did the cutover, so a rejected label (`InvalidLabels`) or a raised error is logged and swallowed, never failing or reverting promote. The `build-` prefix is required (SSM labels can't start with a digit).

## Behavior

- **Automated + gated**: `promote` is the final task of `load_people_api` (`… resize >> analyze >> promote`), so a successful refresh promotes itself. It refuses unless the run's `validate` manifest is complete with all checks passed. It is intentionally **not** a `scale_down_on_failure` upstream (resize already made the cluster serving-ready, so a promote failure must not scale it down).
- **Rollback** is a manual operator (AWS admin group) action: write a prior connection string back as a new version; people-api picks it up within its revalidate interval. Only works while that prior cluster still exists, so don't teardown the previous cluster until the new one is confirmed healthy.
- The version-preserving `overwrite_ssm_parameter` is used (not `put_ssm_parameter`, which delete+recreates to enforce tags and would reset versions).
- The old manual-only `promote_people_api` DAG is removed as superseded.

## Companion / follow-ups (out of this diff)

- `ssm:LabelParameterVersion` on the serving parameter — merged in gp-terraform-dataplatform #49 (still needs the gated apply).
- A loader cache-bust bump to deploy the promote step to the Astro image after merge.


hhkarimi:
delegate review

The prior blocker (shell-injection in `promote_people_api.py:45` via `dag_run.conf`) is resolved by removing that manual DAG entirely: promote is now the automated final step of `load_people_api`, invoked as `loader promote --date {{ ds_nodash }}` (Airflow's own date string, not operator input), matching every other step.

hhkarimi:
delegate review

Merged `main` to resolve conflicts. The only real conflict was in `aws.py`: `put_ssm_parameter` on main now delete+recreates an existing parameter to enforce an exact tag set, which resets version numbering and drops all version labels. Using that on the serving parameter would wipe the `build-`/`live` label history every cutover, so I kept main's `put_ssm_parameter` for the dated params and added a version-preserving `overwrite_ssm_parameter` (plain Overwrite, no tag churn) that `promote` now uses.

hhkarimi:
delegate review

Per team discussion: dropped the `live` label. gp-api stays on the serving param's latest version, so the overwrite is the cutover and no gp-api change is needed; rollback is an operator writing a prior connection string back as a new version. promote now applies only `build-{date}` (one label per version, safe under SSM's per-version 10-label cap), and labeling is best-effort — `label_ssm_parameter_version` returns `InvalidLabels` and promote swallows both a rejected label and a raised error with a warning, since the overwrite already did the cutover.

hhkarimi:
delegate review

Pulled latest main (includes the elections/civics fixes DATA-2214/DATA-2226 and the general-mart deprecation) to clear the unrelated int__civics slim-CI failures, and reworded one test comment to drop a residual `live` mention. No functional change since the last approval.

PR #707, opened 2026-07-28
Match serving column types to the Prisma contract + type-diff guardrail

hhkarimi:
## Why

Swain's functional tests found many columns in the freshly-loaded cluster whose types are not compatible with the app — 45 columns across three tables where the new DB type differs from the prod / Prisma contract:

- **Voter (40 → TEXT):** 26 election flags (mart `boolean`), 8 ZIP / DPBC / ethnic / sequence codes (mart `integer`), 6 lat/long + voting-performance scores (mart `double precision`). Prisma types all `String`.
- **District:** `created_at`/`updated_at` (`timestamptz` → `timestamp`), and `state` (`text` → `USState` enum).
- **DistrictStats:** `district_id` (`text` → `uuid`), `updated_at` (`timestamptz` → `timestamp`).

## Root cause

`emit_ddl` types every column from the Databricks mart (`to_pg_type(col.spark_type)`) unless `schema_spec` overrides it. Only a handful were overridden, so the rest silently took the mart's types. The serving contract is Prisma, and the two diverge. And nothing compared built types against the contract — `schema_diff_clean` checks names only — so it surfaced in functional tests.

## Fix (three parts)

1. **Correct the types** via `schema_spec` overrides + regenerated `target_schema.sql`. Only the flagged columns change; legitimately numeric ones (`Age_Int`, `FIPS`, confidence codes) are untouched.
2. **Resolve District.state** by appending `US` to the `USState` enum (`create_schema._USSTATE_LABELS`, after the 51 state labels so existing ordinals are preserved) so District's country-scope row (`type=Country`, `state=US`) fits, and typing `District.state` as the enum to match prod. `US` is an enum label only, not a partition key. **Requires a coordinated people-api change: the Prisma `USState` enum must also include `US`**, or the app fails reading that row.
3. **Guardrail:** a new `validate` check, `schema_types_match`, compares every shared column's type (`udt_name`) on the new cluster against prod per table and fails the run on drift — the type analogue of `schema_diff_clean`. Intended divergences live in `schema_spec.ACCEPTED_TYPE_DIVERGENCES` (now empty).

## Not covered here: value fidelity (leading zeros)

Correcting the column type does not restore values the mart already coerced — a ZIP the upstream typed as `integer` has already lost its leading zeros. The coercion is in the PySpark model `int__l2_nationwide_uniform.py` (the mart passes these through untyped), so preserving the original strings is a separate dbt change in the L2 intermediate lineage, followed by a reload. Tracked separately for the data team.


hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

PR #712, opened 2026-07-28
Read raw L2 source as strings; match prod values for the L2 code columns

hhkarimi:
## The bug

Zero-padded codes lose their leading zeros. An MA ZIP `01854` is stored as `1854`, in both the new cluster **and current prod** (prod green.Voter returns `2191` for a voter that should be `02191`).

## Root cause

`load__l2_s3_to_databricks.py` reads every L2 file with `inferSchema=True`. Spark infers a type per column, so any numeric-looking column is parsed as int/double at ingestion and the leading zero is gone the moment the file is read (confirmed: `dbt_source.l2_s3_ma_uniform.Residence_Addresses_Zip` is `int 1854`). The staging passthrough and the int model's string-cast can't recover a zero that never landed.

## Fix

Read the raw files as all-string (`inferSchema=False`) so the source keeps the exact vendor text, and cast downstream only where a non-string type is needed. This also matches the repo convention that casting belongs downstream, not at the raw read.

## How this maps to the full mismatch list

This PR is the **value-fidelity** half; the loader PR (#707) is the **type** half. Together they cover the full list:

| Group | Type fixed by | Value fixed by |
|---|---|---|
| 8 integer codes (ZIP / ZipPlus4 / DPBC / ethnic / sequence) | #707 → text | **this PR** (flip → string passthrough, leading zeros preserved) |
| lat/long (2) | #707 → text | **this PR** (flip → exact source string, no float reformatting) |
| voting-performance (4) | #707 → text | **this PR** (mart int→string cast → `42`, not `42.0`) |
| election flags (26) | #707 → text | already correct — the int model casts to boolean, which unloads as `true`/`false`, exactly what prod stores |
| District `state`→enum, `created_at`/`updated_at`→timestamp; DistrictStats `district_id`→uuid, `updated_at`→timestamp | #707 | n/a (not value bugs) |

Two Voter columns (`FIPS`, `CalculatedRegDate`) were relying on inference as a passthrough and would regress to text under the flip; this PR casts them explicitly in the mart so they stay int/date. Every other non-text Voter column was already explicitly cast or generated.

## Blast radius — read before merge

- **This re-types the entire L2 source to string**, affecting every L2 consumer, not just people-api. The Voter contract is handled here; the District family is covered by #707's `schema_types_match` guardrail. **All other L2 consumers (haystaq, turnout inference, zip-to-district, deid_voters, the write paths) must be validated with a full `dbt build` before merge** — anything doing numeric ops on a now-string column needs its own cast.
- **Takes effect only after a full L2 re-ingest + mart rebuild + a loader run.** The zeros come back on a fresh read of the `.tab` files, so the existing prod-20260728 cluster needs a rebuild (an in-place ALTER cannot recover them).

No ticket linked yet — please attach the ClickUp ticket.

### DIFF
--- a/dbt/project/models/load/load__l2_haystaq_s3_to_databricks.py
+++ b/dbt/project/models/load/load__l2_haystaq_s3_to_databricks.py
@@ -123,10 +123,14 @@ def model(dbt, session: SparkSession) -> DataFrame:
             table_name = _extract_table_name(source_file_type, state_id)
 
             s3_path = f"s3a://{s3_bucket}/{file.s3_state_prefix}{source_file_name}"
+            # Read every column as string (inferSchema=False), matching the main L2 loader. LALVOTERID
+            # is the join key to the uniform table in int__l2_nationwide_uniform_w_haystaq; if this
+            # side inferred it as a number and dropped a leading zero, the join would silently miss
+            # any zero-padded id. Keeping both sides string keeps the keys identical.
             data_df = session.read.options(delimiter="\t").csv(
                 path=s3_path,
                 header=True,
-                inferSchema=True,
+                inferSchema=False,
             )
             data_df = data_df.withColumn("loaded_at", current_timestamp())
 
--- a/dbt/project/models/load/load__l2_s3_to_databricks.py
+++ b/dbt/project/models/load/load__l2_s3_to_databricks.py
@@ -180,10 +180,15 @@ def model(dbt, session: SparkSession) -> DataFrame:
             # set up reader and add loaded_at column
             delimiter = "\t" if source_file_name.endswith(".tab") else ","
             s3_path = f"s3a://{s3_bucket}/{file.s3_state_prefix}{source_file_name}"
+            # Read every column as string (inferSchema=False). The raw L2 files carry zero-padded
+            # codes (ZIPs, ZipPlus4, DPBC, sequence codes) that Spark's type inference silently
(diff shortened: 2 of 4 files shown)


hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
Not applying the lat/long `try_cast(... as double)` suggestion (finding 48e03259) — it would undo this PR's purpose.

`Residence_Addresses_Latitude`/`Longitude` are intentionally strings here. Prod's serving contract stores them as `text` (verified directly against `green.Voter`), and PR #707 types the serving columns TEXT to match. This PR's whole point is to preserve the exact source string instead of a reformatted double. Casting to `double` in the mart would make COPY load the double's text form into the TEXT serving column, reintroducing the float-reformatting fidelity loss we're removing.

The `try_cast("double")` in `write__l2_databricks_to_gp_api.py` is not a precedent for this: that legacy write path targets numeric per-state `Voter{state}` columns, so it needs a numeric cast; the serving mart targets a TEXT contract, so it must not. A downstream analytics/ML consumer that wants a numeric lat/long can cast at read time; the serving mart's job is to match the serving contract, which is text.

Leaving this as a deliberate divergence for human review rather than looping.

hhkarimi:
delegate review

PR #716, opened 2026-07-29
Make election_api Person slug globally unique

danpelota:
## Summary

The public person profile URL is moving to `/people/<slug>` with no trailing UUID (see the paired election-api change adding a unique index on `Person.slug`). That makes the slug the authoritative lookup key, so it must be globally unique. Today the mart emits bare `first-last`, which collides heavily (e.g. one name maps to 82 people), so the unique-index migration cannot apply.

This makes every person slug carry an 8-hex suffix from the person id, enforces uniqueness with a dbt test, and generalizes the `slugify` macro so the whole slug is built in one call.

## Changes

- `m_election_api__person.slug` is now `slugify(first_name, last_name, left(gp_person_id, 8))` — an 8-hex suffix on every row (not only on collision), so a slug only changes if the person's id re-mints, never because a later same-name person appears. Added a `unique` test on `slug`.
- `slugify` is now variadic: it joins any number of column/string expressions with `-` and runs the existing cleaning pipeline, which drops empty/null fields and trims stray hyphens. This removes the need for the manual leading-hyphen trim (the 13 of 507k rows whose names slugify to empty now fall back cleanly to just the hex).
- Dropped `keep_hyphens` (its two branches were identical, so it was a no-op) and the orphaned `generate_candidate_slug` macro (no callers).

## Notes

- The suffix rides `gp_person_id`, minted by the earliest-member rule, so slugs are stable across ETL runs and safe under full refresh. The `unique` test is the real guarantee; if 8 hex ever collides, widen the suffix.
- Unblocks the election-api unique-index migration, which needs Person re-materialized on dev first.

## Testing

`dbt build` passes for all five `slugify`-consuming models (`m_election_api__person`, `m_election_api__candidacy`, `m_election_api__race`, `int__enhanced_place`, `int__geo_id_attributes`), including `unique_m_election_api__person_slug` against ~507k rows.

### DIFF
--- a/dbt/project/macros/slugify.sql
+++ b/dbt/project/macros/slugify.sql
@@ -1,113 +1,35 @@
-{% macro slugify(column_name, keep_hyphens=true) %}
+{% macro slugify(column_name) %}
     {#-
-        Matches npm (slugify package)[https://www.npmjs.com/package/slugify] behavior with lower=true and defaults:
-        - replacement: '-'
-        - lower: true
-        - strict: false
-        - trim: true
-
-        Args:
-            column_name: The column or string expression to slugify
-            keep_hyphens: Boolean to keep existing hyphens (default: true)
+        Slugify one or more column/string expressions into a single lowercase,
+        hyphen-joined slug. Extra positional args are joined with '-'; empty or
+        null fields drop out, so there are no leading, trailing, or doubled
+        hyphens. Matches npm (slugify)[https://www.npmjs.com/package/slugify]
+        with lower=true, replacement='-', strict=false, trim=true.
 
         Example:
             {{ slugify('title') }}
-            {{ slugify('title', keep_hyphens=false) }}
+            {{ slugify('first_name', 'last_name', "left(id, 8)") }}
     -#}
-    {% if keep_hyphens %}
-        trim(
-            both '-/'
-            from  -- trim leading/trailing hyphens and forward slashes
+    {%- set cols = [column_name] + (varargs | list) -%}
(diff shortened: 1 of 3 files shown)


danpelota:
delegate review

danpelota:
delegate review

dbt/project/macros/slugify.sql
```diff
@@ -1,113 +1,35 @@
...
-                        '-'  -- collapse multiple hyphens
-                    ),
-                    '/{2,}',
-                    '-'  -- collapse multiple forward slashes (can use '/' instead)
-                )
-        )
-    {% endif %}
-{% endmacro %}
-
-{% macro generate_candidate_slug(first_name_col, last_name_col, office_name_col=none) %}
```
danpelota, line 60: This was orphaned

oakinh:
Heads up: this DB-globally-unique-slug approach looks **superseded** by the app-side design that shipped in the person-profiles app + election-api.

That design keeps `Person.slug` intentionally NON-unique (`first-last`) and resolves `/people/<base>-<id8>` where `<id8>` = first 8 hex of the person UUID, via an indexed id-range scan in `getPersonBySlug`. Collisions are structurally impossible without a generation-time uniqueness check, so the `@unique` slug + dbt `unique` test here (and the paired `Person_slug_key` migration) aren't required for the profile URLs.

Recommend closing unless the globally-unique slug is still wanted for another consumer. Flagging to the author to confirm before we close — not closing unilaterally.

PR #717, opened 2026-07-29
Guard L2 SFTP sync logs against full-refresh

hhkarimi:
## Why

`load__l2_sftp_to_s3` and `load__l2_haystaq_sftp_to_s3` look like transforms but are **stateful append sync logs**. Their UDF skips any state whose file already exists in S3 (`return EMPTY_LOAD_DETAILS`), so a `--full-refresh` drops the accumulated multi-state history and rebuilds with rows only for the states that happen to have a brand-new vendor file that day. Every other state disappears from the log, and since `load__l2_s3_to_databricks` derives its `state_list` from that log, those states can never be re-ingested again — their `dbt_source.l2_s3_{state}_uniform` tables go stale and orphaned.

This actually happened in prod: a full-refresh collapsed the log to 16 states, leaving the other 35 stuck on the pre-`inferSchema=False` load (leading zeros stripped from ZIP/DPBC/sequence codes). It was not recoverable from Delta time travel (older than the 168h retention).

Pinning `full_refresh=False` on both models makes `--full-refresh` — including the on-merge `state:modified+` build — a no-op for them; they always run incrementally and only ever append. This is the safe default for a sync log.

## Note on merging

The prod sync log has already been repaired out-of-band (its 35 missing states re-pointed at their existing S3 files), so it now holds all 51 states again. Merging this PR triggers the on-merge `state:modified+ --full-refresh`, which will re-ingest `load__l2_s3_to_databricks` from the restored 51-state log and rebuild the people-api marts with correct leading zeros.

That on-merge run does a heavy full-refresh of the L2 lineage and has previously hit `clusterByAuto` concurrent-write conflicts, so it should be merged in a quiet window with the scheduled L2 dbt jobs paused.

### DIFF
--- a/dbt/project/models/load/load__l2_haystaq_s3_to_databricks.py
+++ b/dbt/project/models/load/load__l2_haystaq_s3_to_databricks.py
@@ -42,6 +42,11 @@ def model(dbt, session: SparkSession) -> DataFrame:
         incremental_strategy="append",
         unique_key="id",
         on_schema_change="fail",
+        # Append load-history log. A full refresh isn't data-losing here (it rebuilds from the
+        # guarded upstream S3 log) but re-reads every state and is prone to clustering conflicts.
+        # Pin full_refresh off so heavy re-reads stay a deliberate op, not an on-merge side effect;
+        # the incremental history gate then stays authoritative.
+        full_refresh=False,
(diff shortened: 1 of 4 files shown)


hhkarimi:
delegate review

PR #719, opened 2026-07-29
[DATA-2196] Fix HubSpot contact-company association and surface company id on candidacy

hhkarimi:
## What this does

Fixes two related HubSpot company-linkage bugs (DATA-2196), reported by the data team.

**Bug 1 — `int__hubspot_contacts_w_companies.company_id`.** The model reconstructed the contact-to-company link from engagement (activity) records via a regex, so a contact only got a `company_id` if it shared a logged call/email/note with a company. It now sources `company_id` from the contact's direct HubSpot companies association (an explode of `int__hubspot_contacts.companies`), which is already synced. The existing `ranked_matches` window still collapses to one row per contact, so grain is unchanged.

- Coverage of directly-associated contacts: **50.4% → 100%** (about 32k contacts recovered).
- Reproduce test `assert_hubspot_contacts_w_companies_direct_assoc_has_company`: 28,988 → 0 rows.

**Bug 2 — `candidacy.hubspot_company_ids` for post-2026 candidacies.** The mart read this column only from the BallotReady source, which hard-codes it null, so current candidacies never carried a company id even though `hubspot_contact_id` was populated. gp_api candidacies carry the HubSpot company id on the campaign, so `int__civics_candidacy_gp_api` now emits it as a JSON-array string and `candidacy.sql` coalesces it ahead of the BallotReady source.

- Verified against production: **11,798 → 0** campaign-linked candidacies missing the company id.
- Reproduce test `assert_candidacy_hubspot_company_ids_populated`: 11,517 → 0 rows.

## Also included

- `birth_date` in the int model is now wrapped in `try_cast(... as date)`. The wider join surfaces campaign `dob` values in non-ISO formats; since `birth_date` is date-typed, the coalesce hard-cast errored the build. This mirrors the file's existing `try_cast` pattern for the election-date columns and is type-preserving (date to date), so only malformed strings change behavior (they fall back instead of crashing the build).
- Doc correction in `m_civics.yaml` for the `hubspot_company_ids` provenance and format.

## Follow-up (separate ticket: DATA-2197)

For gp_api candidacies, `hubspot_contact_id` actually holds the HubSpot **company** id, not a contact id (verified: `campaigns.hubspot_id` matches 56,240 HubSpot companies and 0 contacts). This PR deliberately leaves that column untouched and only populates `hubspot_company_ids`, because correcting the mislabel (nulling `hubspot_contact_id` for these rows) also affects `prospects.sql` and the `candidacy_techspeed` HubSpot reverse-ETL. That correction is tracked separately in DATA-2197.

## Testing and CI note

Verified via per-model `dbt build` plus the two reproduce tests, and an independent production-data check of the mart fix.

Heads up for review: `int__civics_candidacy_gp_api` materializes 0 rows in a partial local build, due to a pre-existing production-deferral staleness in its upstream inner join to `int__civics_candidate_gp_api` (a model this branch does not touch). In that state a few downstream `relationships` tests (keyed on `gp_candidacy_id` / `gp_person_id` / `gp_candidate_id` / `gp_election_id`, none on `hubspot_company_ids`) fail or warn. These are not regressions from this branch and resolve when the full DAG is built in production.

## Data validation (post-deploy, production)

Confirmed live after the merged build rebuilt `candidacy`. "Before" = production pre-merge; "after" = current production.

| table | metric | before | after |
|---|---|---|---|
| `candidacy` | rows with `hubspot_company_ids` | 30,974 | 42,801 |
| `candidacy` | campaign-linked rows missing company id | 11,798 | 0 |
| `candidacy_scored` | rows with `hubspot_company_ids` | 30,974 | 42,801 |
| `candidacy_scored` | campaign-linked rows missing company id | 11,798 | 0 |

All 11,798 post-2026 campaign-linked candidacies now carry `hubspot_company_ids` in both `candidacy` and `candidacy_scored`.

## Tickets

- DATA-2196 (this fix)
- DATA-2197 (follow-up: correct the `hubspot_contact_id` mislabel)


hhkarimi:
delegate review

PR #724, opened 2026-07-30
Fix FIPS and audit-timestamp serving types to match prod

hhkarimi:
## Why

The prod `load_people_api` validate step flagged four serving-type mismatches against the prod contract (`schema_types_match:Voter` and `:DistrictVoter`). This fixes all four so the next build passes cleanly:

- **FIPS** — the voter mart did `cast(Voters_FIPS as int)`. `Voters_FIPS` is a uniform 3-digit code (`001`–`840`), so int both **stripped the leading zero** and shipped it as `int4` where prod is `text`. Now passed through as the raw string, and added to the int model's `IDENTIFIER_STRING_COLUMNS` alongside the ZIP/DPBC codes so it stays string through the union.
- **Audit timestamps** — Voter `created_at`/`updated_at`/`Voter_Status_UpdatedAt` and DistrictVoter `created_at`/`updated_at` were emitted as `timestamptz`, but prod is `timestamp` without time zone (as District/DistrictStats already are). DistrictVoter was explicitly `TIMESTAMPTZ` on a wrong assumption; corrected.
- **Voter.State** — served as the `USState` enum (matching District and swain-db, per the 2026-07-21 decision so the app needs no code change) while the current prod baseline is still `text`. This is an intended forward migration, so it's added to `ACCEPTED_TYPE_DIVERGENCES` rather than left to fail the guardrail.

Everything else in validate passed (row counts match the unload for all 51 states, schema-diff clean, indexes valid, sample queries pass). The Maine voter-count delta is a real L2 purge (accepted, not a code issue).

## Rollout

Makes the **next** cluster build correct without any manual patching. The existing `prod-20260730` cluster is being fixed in place (FIPS is 3-digit so the stripped zeros are recoverable with a zero-pad; the timestamp/enum items are type-only) to avoid a full rebuild.

### DIFF
--- a/dbt/project/models/intermediate/l2/int__l2_nationwide_uniform.py
+++ b/dbt/project/models/intermediate/l2/int__l2_nationwide_uniform.py
@@ -91,6 +91,7 @@ PERFORMANCE_PERCENTAGE_COLUMNS = [
 # leading zeros are preserved (e.g. a New Jersey ZIP 08731 is not stored as 8731).
 IDENTIFIER_STRING_COLUMNS = [
     "Residence_Addresses_Zip",
+    "Voters_FIPS",
     "Residence_Addresses_ZipPlus4",
     "Mailing_Addresses_Zip",
     "Mailing_Addresses_ZipPlus4",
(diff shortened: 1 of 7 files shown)


hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

PR #734, opened 2026-08-02
Force-null all columns in copy so mart NULLs land as NULL

hhkarimi:
## Why

On a freshly-built serving cluster, **78 green.Voter text columns store `''` where the contract wants NULL** — e.g. `General_2026` and `Mailing_HHGender_Description` are `''` in all 218M rows, `NameSuffix` in 210M. The mart is correct: those columns are **NULL** in `m_people_api__voter` (the NULL counts match the cluster's `''` counts exactly). The loader's `copy` step is what reintroduces `''`.

Root cause: `#730`'s `source_nulled` CTE NULLs every string empty in the mart, so no column carries a genuine `''` to preserve. But Spark writes those mart NULLs as a **quoted-empty `""`** CSV field, and PG's `NULL ''` only nulls an **unquoted** empty. `_force_null_columns` was force-nulling only *typed* columns (to avoid cast errors), leaving text columns' quoted-empties to import as `''`. So the copy design was out of sync with the mart's `#730` behavior.

## What

`_force_null_columns` now returns **every** column (in DDL order), so `FORCE_NULL` converts every quoted-empty back to NULL uniformly — text and typed alike. This is safe precisely because the mart source-nulls all string empties: no column has a genuine `''` that should survive, and force-nulling a column with real data only affects quoted-empty fields (real values are untouched; NOT-NULL key columns never carry `''`).

Updated the `test_copy_s3.py` unit tests to lock the new behavior. `test_format_contract.py` (base option pairing) is unchanged and still passes, and the local Postgres integration round-trip passes.

**Takes effect on the next loader run** — the current clusters keep their `''` until a rebuild.

## Note on the env-tag ask

The connection-string param's `Environment` tag is **already** optional and does not need a change: `provision.py` strips it when `--no-ssm-env-tag` is set (the sole `put_ssm_parameter` caller), and `put_ssm_parameter` delete-and-recreates an existing param to enforce the exact tag set (so a dropped `Environment` is truly removed, not left by an upsert). `test_cluster_lifecycle_moto.py` locks this. The cluster's tag stays mandatory because the RDS create-IAM policies condition on `RequestTag/Environment`.

### DIFF
--- a/people-api-loader/src/loader/people_api/steps/copy_s3.py
+++ b/people-api-loader/src/loader/people_api/steps/copy_s3.py
@@ -73,32 +73,23 @@ _SESSION_SQL: tuple[str, ...] = (
 # clause (see `_import_options`) is appended to this base at runtime.
 _IMPORT_OPTIONS = "(FORMAT csv, DELIMITER E'\\t', NULL '', QUOTE '\"', ESCAPE '\"', ENCODING 'UTF8')"
 
-# Target types that can legitimately hold an empty string. For these, an empty CSV field stays as
-# '' (the null-vs-empty-string distinction is deliberate — see unload_sql). Every OTHER type
-# (INTEGER, BOOLEAN, DATE, TIMESTAMPTZ, DOUBLE PRECISION, UUID, ...) cannot parse '', so its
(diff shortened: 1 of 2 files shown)


hhkarimi:
delegate review

hhkarimi:
delegate review

PR #759, opened 2026-08-04
[DATA-2190] Seed candidacy/party worklist from upcoming API race rosters

hhkarimi:
## Why

A candidate's competitive landscape (the election-api "Know Your Opponent" roster) was showing the primary field instead of the November general ballot. Root cause: the candidacy and party ingestion models seeded their BallotReady fetch worklist only from the S3 candidacies feed, which omits many upcoming general-stage rosters that the BallotReady API race object already carries. Those candidacies were never fetched and never reached election-api, so when the general roster came back empty the product fell back to the primary field.

This is not isolated to one race: roughly 3,972 of 5,201 upcoming 2026 general races have candidacies in BallotReady's API but zero rows in the S3 feed.

## What

- Add `int__ballotready_upcoming_candidacy_ids`: candidacy ids from the API race roster (`race.candidacies`), scoped to upcoming elections to keep the fetch bounded.
- Union it, deduplicated, into the fetch worklists of `int__ballotready_candidacy` and `int__ballotready_party`. `int__ballotready_person` inherits the fix, since its worklist derives from the candidacy model. `m_election_api__candidacy` needs no change; it left-joins candidacy/person/party and only filters on name-not-null, so the union flows straight through to the election-api roster.

## Tradeoffs

- The incremental refetch of a changed upcoming roster keys off `race_updated_at >= max(candidacy.updated_at)` (two different timestamps) plus a not-yet-stored backfill. Worst case it refetches slightly more, never less.
- The new relationships test (upcoming roster ids resolve to an ingested candidacy) is `warn` severity on purpose, so incremental lag between a roster appearing and the next candidacy run, and API-dropped or withdrawn candidacies, surface for monitoring rather than fail the build.

## Validation

Verified end to end on a scoped run against the real models: building the chain for a single office's general roster produced the full 16-candidate roster in `m_election_api__candidacy`, with names and party populated, where it previously had zero. The full production backfill runs incrementally through the scheduled job.

Ticket: https://goodparty.clickup.com/t/86ajrfpq9

### DIFF
--- a/dbt/project/models/intermediate/ballotready_api/int__ballotready_candidacy.py
+++ b/dbt/project/models/intermediate/ballotready_api/int__ballotready_candidacy.py
@@ -3,6 +3,7 @@ import random
 import time
 from base64 import b64encode
 from collections.abc import Callable
+from datetime import datetime, timedelta
 from typing import Any
 
 import pandas as pd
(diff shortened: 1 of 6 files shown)


hhkarimi:
@claude review

hhkarimi:
delegate review

hhkarimi:
delegate review

hhkarimi:
delegate review

PR #764, opened 2026-08-04
[DATA-2133] Serve Users

datawithtristan:
## What

Encodes the Serve product user population as `serve_users`, label **Serve Users**. **Value 1,846.**

Definition owned by `users_serve_base.is_serve_user`, which resolves to holding a Serve organization or being a poll user, and is currently 1:1 with `eo_activated_at` being set (a data-state fact that may diverge).

## This is an account count, not an activity measure

Stated explicitly in the description because it is the thing most likely to be misread. There are three distinct Serve population concepts and this is only the first:

| Concept | Definition | Value | Status |
|---|---|---|---|
| **Serve Users** (this PR) | Has a Serve org or is a poll user | **1,846** | here |
| Activated Serve Users | Lifetime gate: ever sent an SMS poll AND ever pledged | 1,026 | separate PR, renames the currently-ratified metric |
| Active Serve Users 30d | Trailing-30-day activity | ~102 | blocked on a dbt model rebuild |

The description also records that the first two are **not cleanly nested**: one user carries the activated flag without the Serve-user flag, so neither is a strict subset of the other and a reviewer should not assume 1,846 minus 1,026 is a clean segment.

## Internal accounts are excluded in the measure

33 of the flagged Serve users are `@goodparty.org` staff. `users_serve_base` keeps all users by design so funnel denominators stay correct, so the exclusion has to happen somewhere. Putting it in the measure rather than leaving it to the consumer means:

- every read is correct by default, instead of depending on whoever writes the query remembering
- 1,846 matches the figure already circulated in reporting, rather than introducing a second, larger number

Note for consistency: the existing ratified `active_serve_users` does **not** currently exclude internal (12 of its 1,038 are staff). That is corrected in the companion PR, which restates it to 1,026. Until both land, the two Serve metrics are on different bases.

## Verification

- `dbt parse` clean (only the 2 pre-existing OSI warnings on the civics metrics)
- catalog-freshness `--check` reports the region up to date
- analytics suite: 395 passed
- pre-commit clean
- value cross-checked against prod: 1,879 flagged Serve users, 33 internal, 1,846 net

## Ratification is deliberately not in this diff

`config.meta` carries `owner` and `detail_doc` but no `ratified:` date, since ratifying is the metric owner's decision. To consummate, add one line under this metric's `config.meta`:

```yaml
        ratified: 2026-08-04
```

then rerun `uv run python -m semantic_catalog.cli --write` from `analytics/` so the projection matches, or catalog-freshness will fail.

---

> [!NOTE]
> **Medium Risk**
> New headline population metric with built-in internal-user filtering; misinterpretation vs active/activated Serve metrics could skew reporting until companion PRs align bases.
> 
> **Overview**
> Adds the **`serve_users`** (**Serve Users**) semantic metric as the governed **account population** for Serve (`users_serve_base.is_serve_user`), distinct from behavioral **active** counts.
> 
> A new **`serve_user_count`** measure applies the **@goodparty.org exclusion in the layer** (mart still keeps all users for funnel denominators). Metric copy stresses this is **not** an activity metric and that **`serve_users` vs activated populations are not cleanly nested**.
> 
> The Serve analytics **canonical metrics** projection and semantic-catalog routing test are updated so **`serve_users`** appears alongside **`active_serve_users`**; ratification is left **pending** in yml/meta.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5e9ecf58318b28cd8a11b4c047f6a62372f2eed. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


datawithtristan:
delegate review

datawithtristan:
delegate review

PR #767, opened 2026-08-05
[DATA-2195] Add manual district assignment pattern; fix Gainesville commission districts

danpelota:
## Problem

A user reported that Gainesville City Commission Districts 3 and 4 were missing from the office picker. Confirmed against the [city clerk's election page](https://www.gainesvillefl.gov/Government-Pages/Government/City-Clerk/Upcoming-City-Election): Districts 2, 3, and 4 are on the Aug 18, 2026 ballot and are single-member district seats, not at-large. Mayor and the two at-large seats are citywide.

There's two problems stemming from the fact that L2 carries no council-district geography for Gainesville, so the LLM Gold matcher could only map the district seats to the citywide `GAINESVILLE CITY (EST.)` district:

- **District 3** matched citywide at confidence 85, below the 90 non-state threshold in `m_election_api__position` (which would be an incorrect match)
- **District 4** came back `NOT_MATCHED` entirely.

Both fell through to a null `district_id` and were filtered out of `m_election_api__zip_to_position`. District 2 rendered only because it scored exactly 90 (still a false positive), and District 1 has the same defect pending its next election.

`City_Council_Commissioner_District` is null for all 73,912 Gainesville voters, and so is every other municipal column (`City_Ward`, `City_School_District`, `Township`, `Village`, `Town_Council`, `Borough`, `Municipal_Advisory_Council_District`). L2's FL council-district coverage is limited to Jacksonville, Miami, Miami Gardens, North Miami, and St. Petersburg. The only sub-city geography L2 carries for Gainesville is `Precinct`.

## Approach

The quick fix would have been a `l2_br_match_overrides` row pointing each seat at the citywide district. That makes the seat selectable but offers District 3 to every voter in Gainesville, which is what District 2 does today and is wrong for a single-member seat. **We should move away from this pattern, since it's objectively incorrect**.

Instead this adds a general pattern for assigning districts L2 does not supply, then applies it to Gainesville.

**`l2_manual_district_assignments`** seeds voters to a district on a sparse match tuple — `state`, `county`, `city`, `precinct` — where a null column does not constrain.

**The assignments are layered on inside `int__l2_nationwide_uniform` itself.** The raw union of the 51 state uniform files is renamed to `int__l2_nationwide_uniform_raw_districts`, and `int__l2_nationwide_uniform` becomes a thin view over it that coalesces the seeded assignment into the named district column. Every existing consumer keeps its `ref` and picks up the assignments for free; nothing downstream is repointed. `coalesce` keeps L2 authoritative wherever it does supply a value, and a view means seed edits need no voter-file rebuild.

The position side needs no new mechanism. Once the districts exist in `m_election_api__district`, the existing `l2_br_match_overrides` seed maps each BR position to them, and its existing test already asserts the target district exists. All four Gainesville district seats are retargeted off the citywide district.

## Data

Precinct assignments are transcribed from the county's [published district boundaries](https://www.votealachua.gov/264/District-Boundaries). Precinct 25 is split between Districts 1 and 4 and is assigned to District 1. Split precincts 28/33/48 are assigned whole, so a few hundred voters land in a neighboring district. 419 voters sit in 12 precincts the county does not list at all and keep a null district.

Precincts 42 and 43, which the county lists as split into District 3, contribute nobody: L2 puts all 5,089 of their voters outside the city, so the `city` constraint filters them out. The rows stay in the seed as a record of the county's boundary, and the constraint is doing the right thing — without it those unincorporated voters would be offered a city seat. Precinct 48 is the same shape but has 180 in-city voters, so it does contribute. Every other seeded precinct straddles the city line to some degree, which is why the assignment is constrained on city rather than precinct alone.

## Verification

Districts land on exactly the Gainesville population — 73,493 assigned, 419 null, zero outside the city, with Jacksonville's 14 districts and the rest of L2's own data untouched.

| District | Voters |
|---|---|
| District 1 (est.) | 18,294 |
| District 2 (est.) | 23,515 |
| District 3 (est.) | 13,866 |
| District 4 (est.) | 17,818 |
| (null) | 419 |

The view carries the same 807 columns as the table it reads, with no name differences, so the people-api export contract is unchanged.

The office picker now resolves real district geography instead of citywide:

| Seat | Zips | Voters |
|---|---|---|
| Mayor (at-large) | 14 | 73,912 |
| District 2 | 5 | 23,515 |
| District 3 | 3 | 13,866 |
| District 4 | 8 | 17,818 |

District 1's next election is 2028-08-22, outside the two-year window, so it is correctly absent from the picker today and will resolve when it enters.

`assert_manual_district_assignment_precincts_exist` guards the transcription itself: every seeded precinct must exist in the voter file for its state and county. That catches a transposed number, a wrong county, and the format trap — L2 zero-pads precincts, so a hand-entered `7` would never match `07` and would otherwise assign nobody silently. Verified against synthetic rows: `7`, `99`, and a wrong-county `07` all fail; the real `07` and the inert split precinct `42` both pass. It matches on state and county only, deliberately, since requiring the full tuple would fail on the legitimate split precincts above.

`dbt build` passes across the touched models, including the invariants that would catch a bad crosswalk: `assert_zip_to_br_office_one_district_type_per_br`, `assert_zip_to_br_office_voters_in_zip_invariant`, `assert_override_seed_resolves_to_position_district`, and `assert_override_positions_have_zip_coverage`.

One pre-existing failure is unrelated to this branch: `test_int_haystaq_loaded_at_matches_source` is red because prod `int__l2_nationwide_uniform` is at 2026-08-04 while `int__l2_nationwide_uniform_w_haystaq` lags at 2026-08-01. It fails the same way on `main`.

### One-time rename, before the first run on this code

Because the raw model changes name, each environment needs the physical table moved to the new name first:

```sql
alter table goodparty_data_catalog.dbt.int__l2_nationwide_uniform
rename to int__l2_nationwide_uniform_raw_districts;
```

Between the rename and the dbt run, `int__l2_nationwide_uniform` does not exist, so run them back to back. Anything outside dbt that reads that name (BI, ad-hoc queries) keeps working afterwards, now against a view.

### One-time full refresh, after the seeds load

Editing the assignment seed does not bump voter `loaded_at`, so the `loaded_at`-incremental consumers do not recompute on their own. Prod needs a one-time `--full-refresh` of `int__l2_district_aggregations`, `m_election_api__district`, `int__zip_code_to_l2_district`, and `int__zip_code_to_br_office`, in that order. `assert_manual_district_assignments_resolve` fails until they get one, so a forgotten refresh surfaces as a failing test rather than a silently unselectable seat.

Two things to expect from that refresh:

- `int__zip_code_to_br_office` needs the full refresh specifically because this PR *retargets* existing override rows; an incremental merge cannot delete the old zip→office rows.
- `int__l2_district_aggregations` never deletes on incremental runs, so full-refreshing it drops stale district rows that some serve-cohort districts were binding to. That is the documented caveat in `assert_serve_district_resolution_bind_l2_voters_warn`, and it may raise that warn-level count. Those rows are the existing re-match queue.


danpelota:
Good catch on the sparse-join overlap. Valid: `unique_combination_of_columns` only catches identical match tuples, but two *distinct* rows of the same district type can still both match one voter — a county-wide row (city/precinct null) and a precinct row inside that county. The per-type left join would then fan the voter file out and inflate every downstream count and the people-api export.

No live duplication today, since all 31 current rows specify county+city+precinct and every precinct is distinct. But county-level rows are a planned use of this seed, so the guard matters.

Added `assert_manual_district_assignments_do_not_overlap`: a self-join flagging any two rows of the same (state, district_type) whose sparse tuples do not conflict on any column, excluding rows identical on the full tuple since the uniqueness test covers those. It runs against the seed rather than the voter file, so it is cheap. Verified in both directions — passes on the current seed, and flags the overlap when a county-wide ALACHUA row is added.

danpelota:
delegate review

danpelota:
delegate review

hhkarimi (approved):

PR #769, opened 2026-08-07
[DATA-2226] Fix understated turnout and win numbers for Colorado's at-large commission seats

hhkarimi:
## Problem

A candidate for Larimer County Commission District 1 (2026 General) saw turnout and win numbers roughly a third of what they should be. CS worked around the voter file with the admin district override, but turnout and the win estimate did not follow, because that override only affects the voter file and not the Position to District mapping that drives turnout.

Investigating it showed the same mismapping across all 198 Colorado county commission seats, so this PR fixes Colorado rather than just the reported county.

## Root cause

Colorado commissioners are nominated by residency district but elected by the voters of the whole county. The LLM match snapshot placed each seat on its own `County_Commissioner_District` L2 geography at confidence 100, which is a correct string match but the wrong electorate.

The election-api chain is Race to Position to District to ProjectedTurnout, so the served turnout was the district's share:

| County | Was (avg district) | Now (county) | Factor |
|---|---|---|---|
| Jefferson | 102,196 | 306,639 | 3.0x |
| Douglas | 68,796 | 206,507 | 3.0x |
| Adams | 38,618 | 193,273 | 5.0x |
| Larimer | 62,938 | 188,924 | 3.0x |
| Mesa | 26,651 | 80,019 | 3.0x |
| Pueblo | 25,117 | 75,382 | 3.0x |
| Pitkin | 1,901 | 9,511 | 5.0x |

The uniform 3x and 5x is the expected signature of districts that partition the county.

`Race.win_number` has no live source for this mart (null for every row it emits), so the consumer derives the win estimate from projected turnout. Both numbers were understated by the same factor.

## Why county-wide is correct

Under [C.R.S. 30-10-306(1)](https://colorado.public.law/statutes/crs_30-10-306), "One county commissioner must be elected from each of such districts **by the voters of the whole county**." District-only election is available only to counties over 70,000 that have increased the board to five members ([C.R.S. 30-10-306.7](https://colorado.public.law/statutes/crs_30-10-306.7)). For a three-member board there is no exception available, so the statute settles 56 of the 61 counties.

The five-member boards were each verified individually, and two large three-member counties were spot-checked against their own published statements ([Larimer](https://www.larimer.gov/bocc), [Jefferson](https://www.jeffco.us/2075/Meet-the-Commissioners): "County Commissioner seats are voted on at-large by all eligible Jefferson County voters").

## Deliberately left unchanged

Not every Colorado county elects county-wide, so this is not a blanket mapping. 16 positions across 4 counties keep their district geography:

- **Arapahoe**: district-only. 2024 District 1 drew 65,086 votes against a county-wide electorate near 275,000.
- **El Paso**: district-only since the 1976 vote. 2024 District 2 drew 54,877 against roughly 325,000.
- **Weld**: home rule, three district seats plus two at-large. Only the three district seats are in our data, and their district mapping is correct.
- **Morgan**: position 323088 is a park and recreation seat the match had placed on a commission district. Its electorate is the park district, not the county, so this PR does not assert one for it.

Adams was the one that could have gone wrong quietly. Its own site says residents "select their district representative," which reads as district-only, but its 2024 District 5 race drew 211,790 votes, essentially the whole county. It is at-large with district residency, so it is included.

## Verification

After reloading the seed and rebuilding the position mart:

```
| l2_district_type             | n_positions | n_counties | null_turnout |
| County                       |         182 |         58 |            0 |
| County_Commissioner_District |          16 |          4 |            0 |
```

Every remapped position resolves to a county district that has 2026 and 2028 projected turnout, so nothing lands on a null.

`dbt test --select "m_election_api__position l2_br_match_overrides"` gives PASS=27, including `assert_override_seed_resolves_to_position_district` and the win-number-estimate coverage check.

Two tests fail in the dev target, both identical on unmodified main (verified by stashing and re-running): the seed's district-existence check (4 pre-existing Gainesville FL rows) and `assert_override_positions_have_zip_coverage` (175 rows, all MI/KY/GA/TX/FL/SD). Neither includes a Colorado row.

Originally reported races: `2050828` (2026-11-03 General) and `2050833` (2026-06-30 primary) on Larimer District 1.

## Deploying this

Merging alone does not move production. In order:

1. **Confirm the production job loads seeds.** `dbt build` includes seeds, `dbt run` does not. If the scheduled job is run-only, trigger `dbt seed --select l2_br_match_overrides`.
2. **Rebuild `m_election_api__position`.** No full refresh needed: the incremental filter includes `or tbl_override.br_database_id is not null`, so override rows re-emit every run. Confirmed in dev, where a plain incremental run after reloading the seed moved all 182 district_ids.
3. **`--full-refresh int__zip_code_to_br_office`.** Per the seed's own docs, its merge `unique_key` includes district type and name, so repointed rows leave stale zip-to-office rows behind. Redundant rather than wrong here, since each old district's zips are a subset of its county's, but worth clearing.
4. **Let the election-api writer run.** No special action, but see the note below.
5. **Verify**, then close the loop with CS on the HubSpot ticket. Race `2050828` should resolve position 97531 to County / LARIMER at 188,924.

`Projected_Turnout` needs nothing. The `sync_election_api` DAG does a daily full table swap and all 58 counties already have 2026 and 2028 county rows.

### Note on how these rows reach the API

Worth recording, because it nearly went the other way. `write__election_api_db.py` filters rows by `updated_at > max(updated_at)` in the live Postgres table, and the position mart does **not** bump `updated_at` when an override applies (the race mart does exactly that for filing-address overrides). All 182 affected positions carry an `updated_at` of 2026-06-24 against a mart max of 2026-07-07, so on that logic every one would be filtered out and the fix would look deployed while production kept serving the old numbers.

It works because `Position` is not in the writer's incremental filter list (only `Candidacy, Issue, Place, Race, Stance, District`), so it is written in full every run, and `POSITION_UPSERT_QUERY` carries `district_id = EXCLUDED.district_id` in its conflict update.

That is load-bearing: if `Position` is ever added to that filter list, seed-driven override changes will silently stop propagating. The durable fix would be to bump `updated_at` on override rows in the position mart, mirroring the race mart.

## Open questions and follow-ups

- **Persisted win numbers.** Unclear whether a campaign stores a computed win number or reads it live. If it persists, affected campaigns may need a recompute rather than just picking up the corrected API response. This is a gp-api question and should be answered before telling the customer it is fixed.
- **The product-side bug remains.** The admin district override moves the voter file and nothing else. That is the general complaint in the ticket, "a persistent issue for any candidate that requires an override," and is an app-side ticket, not data.
- **Weld County Council.** Its two district seats are mapped to Weld commission districts. The council is 3 district plus 2 at-large, but the charter text on whether district members are elected county-wide was unreachable (sources 403'd). Left unchanged rather than guessed at.
- **Other states.** The same at-large pattern very likely exists elsewhere and would be a state-by-state audit.


hhkarimi:
delegate review

PR #771, opened 2026-08-07
[DATA-2224] Point Colorado House District 23 at the padded L2 district

danpelota:
Closes [DATA-2224](https://goodparty.clickup.com/t/90132012119/DATA-2224).

Colorado House of Representatives District 23 (BR position 49196) was missing from the candidate office picker, and selecting it via the admin district override returned an empty voter file.

## Root cause

L2 zero-pads legislative district numbers to three characters, so the district is `State_House_District` / `023`. The static LLM L2-to-BR match snapshot (`llm_l2_br_match_20260126`) emitted the unpadded `23` at confidence 100.

`m_election_api__district` holds both names: `023` from the L2 voter file (64,445 registered voters) and a phantom `23` from the turnout-projection feed (no L2 voters). Both symptoms follow from the position resolving to the latter.

- Picker: `int__zip_code_to_br_office` joins zip-to-district on the same name, so the position got zero zip rows and never surfaced in `m_election_api__zip_to_position`.
- Voter file: the district the position pointed at has 0 rows in `m_people_api__districtvoter`.

## Fix

One row in the existing `l2_br_match_overrides` seed repointing 49196 to `023`. That seed drives both the position's `district_id` and the `override_zip_to_br_office` injection, so a single row fixes the picker and the voter file.

Verified in dev after reseeding and rebuilding `int__zip_code_to_br_office`, `m_election_api__position`, and `m_election_api__zip_to_position`:

- position 49196 now resolves to district `023`, 64,445 registered voters
- 11 zip rows in `m_election_api__zip_to_position` (80002, 80033, 80212, 80214, 80215, 80226, 80227, 80228, 80232, 80401, 80403), 10 of them above the app's 0.005 `pct_districtzip_to_zip` floor
- all 53 tests on the seed and the three models pass, including `assert_override_seed_resolves_to_position_district` and `assert_override_positions_have_zip_coverage`, both of which fail on the seed alone until the downstream models are rebuilt

No `--full-refresh` needed. The old district produced no zip-to-office rows, so there is nothing stale for the incremental merge to leave behind.

## The general problem, for a follow-up ticket

This is not Colorado-specific. Any state where L2 pads and the district number is below the pad width has the same mismatch. Positions with no zip coverage today: 3,566 State House, 285 Congressional, 278 State Senate. The signature is clean, for example Texas has 150 house districts and only 100-150 have zip coverage; Pennsylvania has 202 and only 100-202; the states whose L2 district names are non-numeric (MA, MD, MN, ND, NH, SD, VT) are fully covered.

The general fix is to normalize numeric district names on both sides of the join, something like `coalesce(cast(try_cast(name as int) as string), lower(name))`, so `023` and `23` collapse and non-numeric names such as `1A` or `Chittenden-1` still match exactly.

It is not a one-line change, which is why it is not in this PR:

- `m_election_api__district` would carry both variants as separate rows, so normalizing the position join alone fans each position out to two districts. The mart needs to dedupe first, preferring the L2-backed row.
- District ids are salted hashes of `(state, l2_district_type, l2_district_name)`, so deduping churns published district ids and roughly 4,100 position `district_id` values, which propagates to the election-api `District` and `Position` tables and to `m_people_api__districtvoter`.
- It would add ~4,100 offices to the picker at once, which is a product-visible change worth landing deliberately rather than as a bug fix.

Seeding one row per affected office is not a viable substitute at that scale, so the follow-up should do the normalization plus the id migration.

---

> [!NOTE]
> **Low Risk**
> Single seed override for one position; same pattern as existing rows, with no SQL or schema changes.
> 
> **Overview**
> Adds one row to **`l2_br_match_overrides`** so Ballotpedia position **49196** (Colorado House District 23) maps to L2’s zero-padded district **`023`** instead of the unpadded **`23`** from the static LLM match.
> 
> That repointing fixes the office picker (zip-to-position coverage) and the voter file (district with ~64k L2 voters vs. a turnout-only phantom district). No model or join logic changes—only seed data, consistent with other curated L2/BR corrections in the same file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1ea2c09eaf75001ba85846b96d5250ae2ba45ac. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danpelota:
delegate review

danpelota:
delegate review

PR #774, opened 2026-08-07
[DATA-2246] Strip L2's zero padding from district names

danpelota:
Closes [DATA-2246](https://goodparty.clickup.com/t/86ajxnbkg). Stacked on #771; base retargets to `main` once that merges.

L2 pads district numbers to a fixed width (`State_House_District` = `023`). The BR match snapshot and the legacy turnout feeds mostly use the bare number, and a district id is a salted hash of `(state, district_type, district_name)`, so the two spellings mint two districts: one holding the voters, one holding the office. Every name-keyed join between them silently missed. #771 patched one office with an override seed row; this fixes the cause.

## What changed

Strip the padding wherever an L2 district name enters the warehouse. All three sides were internally inconsistent, so normalizing only one would have broken the others in the opposite direction:

- `int__l2_nationwide_uniform` — one edit reaches the voter file, district aggregations, the zip-to-district map, the LightGBM turnout inference, and the election-api district mart
- `stg_model_predictions__llm_l2_br_match_20260126` — 365 of the snapshot's own rows are padded
- the five legacy turnout staging feeds — 4,418 padded rows between them

`ltrim('0', col)` rather than a regex, since it reads as what it does: a character-set trim that stops at the first non-zero. `023` to `23`, `001A` to `1A`, `01 (EST.)` to `1 (EST.)`, text and nulls untouched.

Applied to every district column from `get_l2_district_columns()`, not just the six that pad today, so there is no list to keep in sync when L2 pads a new one. Verified a no-op on the rest: no other district column holds a leading-zero value, and no two L2 district names in a state collapse onto each other once stripped.

The Colorado House District 23 override row from #771 is deleted. The match resolves on its own now.

## Measured in dev after a full rebuild

| | before | after |
|---|---|---|
| State House positions with zip coverage | 1,357 | 4,922 of 4,923 |
| Legislative offices with null ICP `voter_count` | 4,504 | 0 |
| Positions bound to a district with no voters | 10,573 | 6,080 |

CO HD23 resolves to district `23`, 64,445 registered voters, 11 picker zips, populated ICP voter count.

Two consumers beyond the office picker were broken and are fixed here:

- **`int__icp_offices`** joins the match name to the aggregation name for `voter_count`, which decides `icp_office_win` / `icp_office_serve`. 4,504 legislative and congressional offices had a null voter count and therefore null ICP flags. **This will move ICP-driven segmentation counts** — worth a heads-up to whoever owns those numbers.
- **`m_election_api__district_top_issues`** *inner* joins the same tuple, so a mismatch drops the district's issue rows entirely rather than nulling a column.

The 6,080 remaining positions on an empty district are an older, unrelated gap: the LLM named a district L2 carries no geography for. Out of scope.

## Referential integrity

The reason this ran undetected is that dbt's `relationships` test is single-column, and the district id is a hash of a three-column tuple. A naming drift never dangles an id — it nulls a left join, drops rows from an inner join, or mints a second district. Nothing fails.

`l2_district_tuple_exists` is a composite `relationships` test for that tuple, applied where it crosses a boundary:

- match snapshot to `int__l2_district_aggregations`, scoped to the four district types L2 covers in every state. Earliest catch, at the source. Would have failed on 4,504 rows.
- `int__l2_district_aggregations` to `m_election_api__district`. A district that never reaches the mart is never minted an id, so `m_people_api__districtvoter` drops its voters through a silent `.filter(isNotNull())`.

Plus singular tests for failures the tuple check cannot see: ICP `voter_count` nulling on an L2-backed district, legislative positions on a voterless district, legislative positions with no zip coverage, and an all-zeros name that `ltrim` would empty.

### Null-share ceilings on ICP office sizing

The exact-tuple tests are scoped, so they miss a drift that moves both sides together. Two proportional ceilings catch null growth from any cause, over the population where a null is not expected: matched, non-judicial, non-appointed offices whose district type L2 carries in that state.

| | runs at | with the padding drift | ceiling |
|---|---|---|---|
| Overall null share | 2.3% | 5.3% | 3.5% |
| Per district type, min 100 offices | max 29% | 100% for State House and Congressional | 35% |

The per-type ceiling is the sensitive one, since a name break concentrates in a single type rather than spreading across the ~255k-office denominator. Legitimate per-type shares top out near 29% (`School_District` 16%, `Multi_township_Assessor` 29%), so a type at 100% is unambiguous. Both were verified to trip against a simulation of the pre-fix state.

## Cutover

District ids change, so this is a full-refresh migration, not an incremental run.

Rebuild order matters in one place: `int__voter_turnout_lgbm_inference` must be refreshed **before** `int__model_prediction_voter_turnout`, or the district mart keeps minting padded turnout-only districts from the stale inference output.

Full-refresh `int__l2_district_aggregations`, `int__zip_code_to_l2_district`, `int__voter_turnout_lgbm_inference`, `int__model_prediction_voter_turnout`, `m_election_api__district`, `int__zip_code_to_br_office`, `int__icp_offices`, `m_election_api__position`, `m_election_api__zip_to_position`, `m_election_api__projected_turnout`, `m_election_api__district_top_issues`, and the people-api district models.

Both writers are upsert-by-id and will never delete the old generation, so the cutover must replace rather than upsert: rebuild `District` in election-api and `District` / `DistrictVoter` / `DistrictStats` in people-api, and re-point `Position.district_id` in the same transaction. In dev the `relationships` tests on `m_people_api__districtvoter` (844,791 rows) and `m_people_api__districtstats` (299) fail against the not-yet-rebuilt prod tables, which is exactly this requirement showing up as a test failure.

Watch the `sync_election_api` ratio gates on the cutover run. `ZipToPosition` grows by roughly 3,600 offices worth of zip rows, tripping the informational `>3x` branch. `Projected_Turnout` shrinks as duplicate district-year rows collapse; confirm it stays above the 0.5x floor or the swap refuses.

---

> [!NOTE]
> **High Risk**
> Changes core district identity hashing and L2 voter-file grain, which drives election-api/people-api districts, zip-to-office coverage, ICP segmentation, and projected turnout—requiring a full-refresh cutover with coordinated DB swaps.
> 
> **Overview**
> **Normalizes L2 district names** by stripping leading zeros (`ltrim('0', …)`) wherever they enter the warehouse, so padded values like `023` align with BallotReady match and turnout feeds and **stop minting duplicate district ids** from the same `(state, district_type, district_name)` tuple.
> 
> New macros `strip_l2_district_zero_padding` and `strip_l2_district_zero_padding_projection` apply this to **all** columns from `get_l2_district_columns()` at **`int__l2_nationwide_uniform`** (after manual assignments), plus **`district_name` / `l2_district_name`** on the LLM L2–BR match staging model and five legacy turnout staging models. The Colorado House District 23 **override seed row is removed** because the match works without it.
> 
> **Referential integrity:** adds generic test **`l2_district_tuple_exists`** (composite relationships on state + district type + name) on L2 aggregations → election-api district mart and on matched legislative/congressional rows in the match snapshot → aggregations. Adds singular tests for ICP `voter_count` binding, legislative positions on empty districts, zip coverage, and all-zero names after strip.
> 
> **Cutover note:** district ids change; downstream requires a **full refresh** and coordinated API table replacement, not incremental upserts alone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 993c3a4fbf96da8ad62f91e3c0f8303832bb5411. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dbt/project/macros/strip_l2_district_zero_padding.sql
```diff
@@ -0,0 +1,23 @@
+{% macro strip_l2_district_zero_padding(column_expr) %}
+    {#- L2 pads district numbers ('023'); the BR match snapshot and turnout feeds
+        do not, so name-keyed joins miss and the salted id mints two districts. -#}
+    ltrim('0', {{ column_expr }})
+{%- endmacro %}
```
danpelota: I don't think this needs to be a whole new macro. It's a call to a single SQL function, so let's just inline the `ltrim` call where we need it.

dbt/project/macros/strip_l2_district_zero_padding.sql
```diff
@@ -0,0 +1,23 @@
...
+        {%- set name = col | trim -%}
+        {%- set _ = out.append(
+            strip_l2_district_zero_padding(
+                relation_alias ~ ".`" ~ name ~ "`"
+            )
+            | trim ~ " as `" ~ name ~ "`"
+        ) -%}
+    {%- endfor -%}
+    {{ out | join(",\n    ") }}
+{% endmacro %}
```
danpelota: Let's just include this as a simple Jinja loop directly in the downstream model.

danpelota:
delegate review

hhkarimi (approved):

PR #784, opened 2026-08-10
[DATA-2258] Null the NOT_MATCHED sentinel on the L2-to-BR match snapshot

hhkarimi:
Ticket: https://app.clickup.com/t/86ajygzb8 (DATA-2258)

## The trap

The LLM L2-to-BallotReady match snapshot encodes "no match" as a literal string, not null:

```
is_matched = false,  l2_district_type = 'NOT_MATCHED',  l2_district_name = 'NOT_MATCHED'
```

12,272 rows. Exactly one spelling, no variants, no empty strings, no nulls.

The strings pass not-null checks, look like data in a preview, and join to nothing. So any filter using `l2_district_name is not null` to mean "this position has a district" quietly counts ~12k unmatched positions as matched.

This is not theoretical. While decomposing district-population coverage I attributed 1,700 offices to "no L2 match" when the real number was 13,820. The sentinel rows sailed through a not-null filter and hid 12,120 offices. Wrong by a factor of eight, in a number being used to reason about whether the Serve ICP could move to census constituents.

## The fix

Null both district columns for unmatched rows, so `is_matched` is the only predicate for match state and a null district means what it looks like.

Then pin the contract in **both** directions: the district columns are null exactly when `is_matched` is false. The current snapshot satisfies that with zero exceptions across 287,622 rows, so the test fixes the contract rather than tolerating drift. If a future snapshot reintroduces the sentinel, or ships a matched row with no district, it fails at staging instead of silently corrupting counts downstream.

## Behavior-neutral, verified per consumer

| consumer | why unaffected |
|---|---|
| `int__serve_district_resolution` | already `where is_matched` |
| `int__zip_code_to_br_office` | already `where ... and is_matched` |
| `m_election_api__district_top_issues` | already `where m.is_matched` |
| `int__icp_offices` | joins the district tuple; sentinel matched nothing, null matches nothing |
| `m_election_api__position` | requires a district to resolve, so these rows already fall to its `unmatched_br_positions` branch with a null `district_id` |
| `candidates_outreach` | the only reader of the string, changed to `is_matched` |

On that last one, the old and new predicates are provably equivalent on the current data:

| | rows |
|---|---|
| total | 287,622 |
| `l2_district_type != 'NOT_MATCHED'` keeps | 275,350 |
| `is_matched` keeps | 275,350 |
| **rows where they disagree** | **0** |

It is left as `is_matched` rather than the string because after this change the string no longer exists, and a filter comparing against a value that cannot occur is its own trap.

## Test results

`dbt build` on the changed models: the new contract test passes.

One pre-existing failure appears in my dev run, `l2_district_tuple_exists` with 298 rows, and it is **not** from this change: its `row_condition` gates on `is_matched`, and this change only touches rows where `is_matched` is false. Traced to a stale shadow copy of `int__l2_district_aggregations` in my personal dev schema, loaded 2026-07-10 and still carrying 290 zero-padded district names, which shadows production's clean 2026-08-07 build. Running the same comparison against production returns **0** rows for both the production and modified snapshots, so CI should be clean.

### DIFF
--- a/dbt/project/models/marts/election_api/m_election_api__position.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__position.sql
@@ -34,11 +34,18 @@ with
             and coalesce(tbl_override.l2_district_name, tbl_match.l2_district_name)
             = tbl_district.l2_district_name
         where
-            tbl_match.l2_district_name not in (
-                'County Committee Female Member',
-                'County Committee Male Member',
-                'President of the United States',
-                'Vice President of the United States'
+            -- null-safe: an unmatched row carries a null district name, which is not
+            -- one of these non-office names. Bare NOT IN evaluates to NULL there and
+            -- would drop the row, silently discarding any override that exists to
+            -- give that very position a district.
+            (
+                tbl_match.l2_district_name is null
+                or tbl_match.l2_district_name not in (
+                    'County Committee Female Member',
+                    'County Committee Male Member',
+                    'President of the United States',
+                    'Vice President of the United States'
+                )
             )
             and tbl_district.id is not null
             and (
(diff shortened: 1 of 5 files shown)


hhkarimi:
delegate review

PR #785, opened 2026-08-10
Deploy the promote step (cache-bust) + remove legacy load_people_api_old DAG

hhkarimi:
Two loader-cutover cleanups:

1. **Cache-bust** `bust=2026-08-04.1` -> `bust=2026-08-10.1` so the Astro image reinstalls the loader from `@main` and picks up the `promote` step merged in #704 (without a bump the image keeps a loader with no `promote` subcommand and the DAG's new final task fails).

2. **Remove `load_people_api_old`** — the legacy `@daily` District/DistrictStats sync. The new `load_people_api` loader now builds District, DistrictStats, and DistrictVoter from the same dbt marts (`m_people_api__district` / `__districtstats`), including the DistrictStats `buckets` camelCase remap, satisfying the DATA-2100 condition in the old DAG's own retire note. Shared include helpers (`postgres_utils`, `databricks_utils`) are kept — still used by `l2_expired_voters` and `sync_election_api`.

Note: District/DistrictStats freshness moves from daily to loader-rebuild cadence; acceptable given the cutover to a loader-built serving cluster.


hhkarimi:
delegate review

Added removal of the legacy `load_people_api_old` DAG (daily District/DistrictStats sync) to this PR — the new loader now covers the District family (District/DistrictStats/DistrictVoter) from the same marts, incl. the DistrictStats buckets camelCase remap. Shared include helpers kept (used by other DAGs).

PR #788, opened 2026-08-10
[DATA-2264] Whole-person district_population on positions, plus [DATA-2242] normalized voter_count fallback

hhkarimi:
Tickets: [DATA-2264](https://app.clickup.com/t/86ajyrfez) and [DATA-2242](https://app.clickup.com/t/86ajxb7jm)

Two changes. The first is value-neutral. **The second changes ICP flags on 3,724 positions and can grow the People Served cohort, so it needs metric-owner sign-off.**

## 1. `positions.district_population` as a whole-person integer

`double` to `bigint`. A mart column that reads as a count of people shouldn't carry a fractional tail. The allocation stays fractional upstream on purpose — that's what keeps allocations summing exactly to block population, which the rollups and binding tests depend on — so `int__district_population` keeps the exact value and the mart rounds.

Verified: **0** of 286,283 rows differ from the previous float rounded to nearest, and no other column changed.

Not applied to `district_census_stats` (a reference table `people_served` does arithmetic on, with a `1e-6` binding test) or `people_served` (the North Star).

## 2. Normalized-name fallback for `voter_count`

The match snapshot emits district names carrying L2's `(EST.)` and whitespace drift, which the exact-name join to `int__l2_district_aggregations` misses. 6,117 matched positions were left unsized, and an unsized position gets NULL ICP gates rather than false. `School_District` (3,867) and `City` (513) dominate.

Fix resolves on the exact name first and falls back to the normalized name **only where the exact join finds nothing**, so no already-sized position can change:

| safety check | result |
|---|---|
| Positions with an existing `voter_count` whose value changed | **0** of 286,283 |
| `int__icp_offices` uniqueness | passes (fallback deduped, join stays 1:1) |

Sizes 3,732 of the 6,117.

### What this moves

All changes are NULL leaving NULL, never a true/false flip:

| flag | NULL → true | NULL → false |
|---|---|---|
| `icp_office_win` | **3,313** | 411 |
| `icp_office_serve` | **361** | 169 |
| `icp_win_supersize` | 138 | 3,586 |

The 361 new Serve-ICP offices become eligible for `in_people_served_cohort`, so the North Star cohort can grow, and `icp_serve` syncs outward to HubSpot. These are corrections — the positions should always have been sized — but they're real value changes on governed surfaces, so please ratify rather than merging on a green test suite.

### Two notes for reviewers

Both lookups alias `voter_count`. A bare column on either would shadow the coalesced select alias the gates read, silently gating on the exact name alone. Databricks raised `AMBIGUOUS_REFERENCE` rather than resolving it quietly, which is the only reason it didn't ship as a silent bug.

`max()` not `sum()` on the fallback: 7 keys nationwide carry two raw spellings of one district (e.g. `SYCAMORE TWP` 13,442 vs `SYCAMORE TWP (EST.)` 935), so summing would double count. `max()` is a safe default here because the fallback can't touch an already-sized position, but what `(EST.)` means in L2 is unresolved.

The remaining 2,385 unsized positions are drift in the frozen snapshot. Accepted rather than fixed, since the crosswalk is being retired.

## Tests

`dbt build --select "int__icp_offices positions"`: **36 pass, 0 fail**, including uniqueness on `int__icp_offices` and the per-type coverage canary from #786.

### DIFF
--- a/dbt/project/models/intermediate/icp/int__icp.yaml
+++ b/dbt/project/models/intermediate/icp/int__icp.yaml
@@ -49,7 +49,13 @@ models:
                 min_value: 0
               row_condition: "district_population is not null"
       - name: voter_count
-        description: "Number of voters in the district"
+        description: >
+          L2 registered voters in the matched district, and the input to all
+          three ICP gates. Resolved on the exact district name first, falling
+          back to the normalized name (case, whitespace, trailing "(EST.)") when
+          the exact name finds nothing, since the match snapshot carries naming
+          drift the aggregations table does not. NULL only where neither
+          spelling resolves, which leaves the ICP gates NULL rather than false.
         tests:
           - dbt_expectations.expect_column_values_to_be_between:
               arguments:
--- a/dbt/project/models/intermediate/icp/int__icp_offices.sql
+++ b/dbt/project/models/intermediate/icp/int__icp_offices.sql
@@ -10,8 +10,7 @@ with
     l2_match as (
         -- the substrate normalizes district names (case, whitespace, trailing
         -- "(EST.)"); the match snapshot does not, so carry a normalized copy for
-        -- the population join. The voter join stays on the raw name, which is
-        -- what int__l2_district_aggregations keys on.
+        -- the population join and the voter-count fallback below.
         select
             *,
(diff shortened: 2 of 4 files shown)


hhkarimi:
delegate review

PR #790, opened 2026-08-11
docs: fold cutover & rollback runbook into people_api_loader.md (DATA-1914)

hhkarimi:
Adds a **Cutover & rollback runbook** section to the loader ops doc `airflow/astro/docs/people_api_loader.md` — the DATA-1914 deliverable and the last open item in the DATA-1640 epic — rather than a standalone file, since that doc is already the canonical loader reference and a separate file duplicated its serving-schema / step-sequence / references content.

The runbook documents the **shipped** overwrite-latest model (gp-api reads the serving param's latest version, revalidates ~5 min; rollback = an admin writing a prior connection string back as a new version), cross-referencing the existing sections instead of repeating them.

While in the doc, reconciled stale bits: added `analyze`/`promote` to the step sequence + table, pointed the Teardown note at the `teardown_people_api_cluster` DAG, corrected the dev serve class to `db.serverless`, and dropped the completed 'Before merge' / DATA-1855 checklist.

### DIFF
--- a/airflow/astro/docs/people_api_loader.md
+++ b/airflow/astro/docs/people_api_loader.md
@@ -5,9 +5,8 @@ on a fresh Aurora Postgres cluster from the L2 voter marts in Databricks. The lo
 four tables: `Voter` and `DistrictVoter` (partitioned by State), and `District` and
 `DistrictStats` (flat).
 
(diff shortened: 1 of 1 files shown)


hhkarimi:
delegate review

PR #792, opened 2026-08-11
Add data-ticket-triage skill

danpelota:
Adds the `data-ticket-triage` skill, which has been living untracked in a local checkout. Nothing else in the repo changes.

## What it does

Covers a ClickUp data ticket end to end: read the ticket, trace the symptom through the dbt DAG to the model where the record stops existing, post the root cause before implementing, ship the smallest fix that reuses an existing repair pattern, verify against prod from dev, open the PR, and iterate with the delegate reviewer.

The parts that carry the most weight in practice:

- **Find the boundary, not the source.** Work backwards from the surface the user complained about to the last model that has the record and the first that does not. Everything upstream of that boundary is a distraction.
- **Quantify isolation before choosing a fix.** One row nationwide is a data-entry gap and takes a seeded override. Thousands clustered on one dimension is a modeling bug and takes a code change. The skill makes this an explicit step rather than a judgment call.
- **Check the consuming application when the mart looks fine.** Points at the `omni` packages and flags the two things that most often change a diagnosis: the service reading a different table than the ticket names, and env-var thresholds or date windows the mart knows nothing about.
- **Read-only boundaries.** The product monorepo is never edited from a data-platform ticket.

## references/repair-patterns.md

Catalogs the override seeds already in the project, what each one keys on, what it repairs, and where it is applied, plus the four-rung chain an office travels to become selectable in the picker. The point is that a missing-record ticket usually lands on a rung that already has a repair path, so we stop growing new machinery for problems that have an established shape. It also records the known `place_id` gap on BallotReady `X`-prefixed special districts, so nobody forces the wrong seed onto one.

## Notes

Both files are prose, so CI is limited to the generic pre-commit hooks. Those pass.

The reference doc opens by telling the reader to verify any model or column named in it against the live DAG before relying on it, since those drift. That caveat is doing real work and should stay.

---

> [!NOTE]
> **Low Risk**
> Documentation-only addition with no changes to pipelines, marts, or application code.
> 
> **Overview**
> Adds a new **data-ticket-triage** Claude skill under `.claude/skills/` plus a **repair-patterns** reference doc. No application or dbt code changes.
> 
> The skill defines a seven-step workflow for ClickUp `DATA-*` tickets: read the ticket, trace backward through the dbt DAG to the model boundary where data disappears, quantify how widespread the defect is, check `omni` app code when marts look fine (read-only), pick the smallest fix (override seed before model logic), post root cause on ClickUp before coding, verify in dev with a surgical dev-vs-prod diff, then open a PR and loop on **delegate-reviewer**.
> 
> It also documents dbt Cloud CLI constraints (single session, no inline `limit`, stale dev schemas) and branch/PR naming from `dbt/project/CLAUDE.md`.
> 
> The reference catalogs existing override seeds (`l2_manual_district_assignments`, `l2_br_match_overrides`, `br_position_place_overrides`, etc.), the four-rung **office picker** diagnostic chain, and the known BallotReady `X`-district `place_id` gap where `br_position_place_overrides` is the wrong tool.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7409acd8a76c6e9407e8fd388cab1e21db08d285. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/skills/data-ticket-triage/SKILL.md
+++ b/.claude/skills/data-ticket-triage/SKILL.md
@@ -50,7 +50,7 @@ One row nationwide is a data-entry gap and takes a seeded override. Thousands cl
 
 ## 3. Read the consuming application when the mart looks fine
 
-If the mart has the data and the product still does not show it, the filter is in the app. The product monorepo is checked out at `~/projects/gp-data-parent/omni/packages/`:
+If the mart has the data and the product still does not show it, the filter is in the app. The product monorepo is the `omni` repository: https://github.com/thegoodparty/omni
 
 | Package | What it tells you |
 |---|---|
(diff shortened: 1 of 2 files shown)


danpelota:
delegate review

PR #795, opened 2026-08-11
[DATA-2260] Scope West Volusia Hospital Authority positions to their own district

sanjayr1:
## Problem

A pro candidate's voter file covered all of Volusia County FL (427,636 registered voters) instead of the West Volusia Hospital Authority district (~165K). All six WVHA Board positions in `m_election_api__position` resolve to the County / VOLUSIA district.

Root cause: L2 supplies no hospital-district geography anywhere in Florida — all five hospital district columns are null across all 15.0M FL voter rows — so the LLM matcher's candidate list had nothing sub-county to offer and it picked the county as "the most appropriate geographic parent" at confidence 90-95, above the non-state gate of 90. A confidence gate cannot catch a confidently chosen superset.

ClickUp: DATA-2260 (root-cause note posted there).

## Approach

Reuses the manual district assignment pattern introduced for the Gainesville commission districts (#767):

- **57 `l2_manual_district_assignments` rows** keyed (FL, VOLUSIA, precinct), transcribed from the Volusia SOE March 2026 "Active Voters by District-Precinct" report. City is left null deliberately: the authority includes unincorporated communities (Osteen, Cassadaga, Seville, DeLeon Springs, ...) whose L2 city is blank, and the authority's boundary follows precinct lines.
- **`int__l2_nationwide_uniform`** gains a second coalesce block applying seeded `Hospital_District` values. Previously the view only applied `City_Council_Commissioner_District`; a seeded type without a coalesce block is a silent no-op (the resolve test catches it, per the new header comment).
- **6 `l2_br_match_overrides` rows** point every WVHA Board position (172907, 388411, 388412, 291543, 291544, 291545) at the new district, overriding the county-level LLM match.
- **Complete the people-api chain's hospital enumeration** (second commit): `m_people_api__voter` and the `m_people_api__districtvoter` unpivot list both carried four hospital variants but omitted plain `Hospital_District`, so voters' values would never have reached the people DB's DistrictVoter table — and since the voter file scopes by DistrictVoter, the file would have gone from whole-county to empty. Two added lines complete the enumeration. The loader's unload projects columns from the serving DDL, so the serving Voter table is unchanged; the retired `write__people_api_db` (`+enabled: false`) is untouched.

## Data

The authority spans 61 precincts totaling 164,860 active voters (matches the ~165K reported). 57 precincts are assigned whole. The 4 boundary-straddling precincts are excluded as mostly outside the authority:

| Precinct | In WVHA (SOE actives) | Countywide (L2) |
|---|---|---|
| 504 | 238 | 1,395 |
| 505 | 176 | 910 |
| 601 | 32 | 7,501 |
| 902 | 1 | 4,920 |

Including them whole would wrongly add ~13,300 outside voters; excluding them under-covers by 647 active voters (0.4%).

## Verification

Dev (dbt_sroberts):
- No fan-out: FL row count identical dev vs prod (15,098,388) after the new left join.
- Assignment exact: 179,967 voters carry the district in the dev view, exactly equal to the row count of the 57 precincts in `int__l2_nationwide_uniform_raw_districts` (FL / VOLUSIA). The source column was 0% populated in FL, so every value came from the seed and L2 stays authoritative elsewhere.
- All six `br_database_id`s exist in prod `int__enhanced_position`.
- `assert_manual_district_assignment_precincts_exist` and `assert_manual_district_assignments_do_not_overlap` pass in dev.

CI preview schema (first run):
- All six positions resolve to `Hospital_District / WEST VOLUSIA HOSPITAL AUTHORITY (EST.)` with registered_voters = 179,967.
- Exactly 6 positions changed `district_id` vs prod.
- Zip coverage: 18 zips for position 172907 (the seat with a 2026 race); the other five correctly have none (next elections outside the 2-year window).
- The first CI run's only error was `write__election_api_db` losing its Postgres connection ("terminating connection due to administrator command") — environmental, during today's election-DB serving cutover; unrelated to this diff. Re-running on the second push.

The two people models are `tag:monthly` (CI-excluded); both compile-checked via dbt Cloud, and the added lines mirror the four existing hospital variants' pattern with the source column verified present in `int__l2_nationwide_uniform_w_haystaq`.

## Deployment

Per the seed's documented process, after merge the `loaded_at`-gated consumers need a one-off `--full-refresh`: `int__l2_district_aggregations`, `int__zip_code_to_l2_district`, `m_election_api__district`, and `int__zip_code_to_br_office` (the district repoint otherwise leaves stale county-wide zip rows for these positions). `assert_manual_district_assignments_resolve` fails loudly until they run. For the people chain, rebuild `m_people_api__voter` (a full-rebuild table) and `--full-refresh m_people_api__districtvoter`, then run the people-db loader; riding the regular monthly cadence instead works but waits on the state's next L2 load to bump loaded_at (documented in the seed schema). A third commit also nullif-guards both assignment coalesce blocks so a state file that loads blanks as ''' cannot suppress an assignment (a no-op today: zero blank-named districts exist in prod).

## Follow-ups (filed)

- DATA-2271: BR vendor escalation — no candidacies on either 2026 WVHA race, duplicate PositionElection (2227165 / 2945433), seat labels disagree with the authority's roster.
- DATA-2272: class audit — up to 229 positions nationwide (30 in FL) with "hospital" in the name LLM-matched to a whole county at >= 90 confidence; also noted there: `int__icp_offices` and `m_election_api__district_top_issues` join the raw LLM snapshot and ignore `l2_br_match_overrides`, and the serving Voter table lacks a plain `Hospital_District` column (only matters for column-filter paths).


sanjayr1:
delegate review

sanjayr1:
delegate review

sanjayr1:
delegate review

sanjayr1:
Independent adversarial review (Codex, read-only, full diff) returned five findings. Dispositions:

1. DistrictVoter will not backfill on a normal incremental run (accepted, docs + runbook). True: its incremental gate keys on updated_at, which is L2 loaded_at, so a seed edit alone propagates only when the state's next L2 load lands. Third commit documents the deterministic path in the seed schema (rebuild m_people_api__voter, then --full-refresh m_people_api__districtvoter) and the PR deployment section now includes it. Declined the code-level backfill machinery: the whole seed pattern's propagation is deliberately runbook-driven (per the pattern's own docs), and a tripwire test on DistrictVoter membership would sit red for weeks between merge and the monthly build.

2. Serving Voter table silently lacks the Hospital_District column (declined, by design). The voter file scopes through DistrictVoter (resolveDistrict -> districtId -> DistrictVoter join in gp-api's voterDownload service), which is column-agnostic. The loader's unload projects from the committed serving DDL precisely so mart-side additions do not alter the serving contract. The column-filter-path gap is recorded on the DATA-2272 audit.

3. Blank-loaded ('') district values would suppress assignments (accepted, fixed). Third commit nullif-guards both assignment coalesce blocks. Verified a no-op today: zero blank-named districts exist in prod aggregations across all types, and dev counts are unchanged (FL 15,098,388 rows; 179,967 assigned).

4. Zip coverage could be vacuously absent if the turnout snapshot lacks the district type (declined, empirically refuted here). The turnout inference derives its type list from int__l2_district_aggregations, which carries Hospital_District (422 districts nationwide), and the first CI preview build produced 18 zip rows for position 172907. The vacuous-test improvement suggestion is noted on DATA-2272.

5. Nothing validates the 57 precincts are the exact boundary (declined, already reconciled). The seed is transcribed from the SOE's own district-precinct cross-reference; the 61 enumerated precincts sum exactly to the report's district total (164,860 active voters), the 4 boundary-split exclusions are quantified in the PR body, and the assigned-voter count equals the 57 precincts' population in the source relation exactly (179,967). Future precinct drift is caught by assert_manual_district_assignment_precincts_exist; boundary redefinitions inherently require re-transcription, which the seed's source column cites.

Codex's proposed verification (each override id resolves to exactly one FL position) is already evidenced: all six ids exist in prod int__enhanced_position and the CI preview position mart carries exactly one row per id, all six on the new district.

sanjayr1:
delegate review

PR #797, opened 2026-08-11
[DATA-2205] BR seats fallback for the viability scorer

sanjayr1:
## Goal

Many BallotReady-linked candidacies have no `seats_available` value on their civics election link, so the viability scorer cannot use the multi-seat race feature for them and falls back to an unscored or lower-confidence path. This PR adds a trust-gated intermediate model that derives a seat count from BallotReady race and position data for exactly those candidacies, and wires it into the scorer as a fallback for the `multi_seat` feature only.

## Coverage

On the November 2025 with-result campaign population (the ticket's acceptance test), viability coverage moves from 62.9% to 95.6% (dry run predicted 95.5%). The fallback is fail-closed: any candidacy that fails a trust check gets no fallback row rather than a guessed seat count, so the coverage gain only comes from candidacies where the fallback can be trusted.

## Design highlights

- **Fail-closed trust gates.** A candidacy is excluded from the fallback (no row, not a guess) if it is on the curated 2025-position nullout seed, if its name-derived office type contradicts the BallotReady position crosswalk, or if same-date BallotReady races disagree on seat count for its position.
- **Two source tiers, kept for provenance.** `race_exact` (a clean, non-primary, non-runoff BallotReady race on the candidacy's own general election date) wins when available; `position` (the position's standing seat count) is the fallback of the fallback. The matched race id is kept on tier-1 rows as provenance evidence for the seat count, not as a roster-verified race link.
- **`multi_seat` only, on both tiers.** The scorer consumes fallback seats for the `multi_seat` feature only. Opponent count is computed at a different (per-election) grain that can span positions, so fallback seats intentionally do not feed `log_n_losers`. Opponent-count wiring is a deferred follow-up.

## Validation gates

All five validation gates plus the model-version precheck passed, comparing a dev-schema baseline snapshot (taken immediately before the scorer-wiring commit) against the new build:

| Gate | Result | Verdict |
|---|---|---|
| Precheck | `model_version` identical per `scoring_model` on both sides; WoE Delta versions unchanged | PASS |
| G1 (grain / key-set parity) | fallback rows = fallback keys = 145,164; scored rows unchanged at 324,133; keyset diff = 0 | PASS |
| G2 (non-viability parity) | 0 rows differ on any non-viability column, both directions | PASS |
| G3 (additive-only scores) | lost = 0, changed = 0 (binding); 14,305 newly-scored candidacies added, globally | PASS |
| G4 (published-value changes) | see below | PASS |
| G5 (acceptance coverage) | see below | PASS |

**G4, on the published mart (`candidacy_scored`):** zero previously-published scores were lost. 5,027 published scores were overridden (changed value), about 9% under the roughly 5,505-row ceiling measured against the legacy-displayed population. Of those, 379 overrides are in November 2025, about 19% under its roughly 469-row ceiling. 9,253 candidacies got a published score for the first time (not gated; this is the fallback's coverage win). Full band-transition matrix for the 5,027 overridden rows:

| band before | band after | rows |
|---|---|---|
| Unlikely to Win | No Chance | 1179 |
| Has a Chance | No Chance | 958 |
| No Chance | No Chance | 486 |
| Frontrunner | Has a Chance | 473 |
| Frontrunner | No Chance | 465 |
| Frontrunner | Unlikely to Win | 396 |
| Likely to Win | Unlikely to Win | 270 |
| Likely to Win | Has a Chance | 212 |
| Likely to Win | No Chance | 200 |
| Has a Chance | Unlikely to Win | 95 |
| Has a Chance | Has a Chance | 65 |
| Frontrunner | Likely to Win | 61 |
| Unlikely to Win | Has a Chance | 41 |
| Unlikely to Win | Unlikely to Win | 39 |
| Likely to Win | Likely to Win | 30 |
| Frontrunner | Frontrunner | 18 |
| No Chance | Unlikely to Win | 14 |
| Likely to Win | Frontrunner | 12 |
| Has a Chance | Likely to Win | 6 |
| No Chance | Has a Chance | 4 |
| Unlikely to Win | Likely to Win | 2 |
| Has a Chance | Frontrunner | 1 |

Rows where the band is unchanged (for example "No Chance to No Chance") are cases where the underlying numeric score moved but the displayed band held.

A post-review fix wave hardened the nullout gate so the curated seed is rejected directly on its own tested key (`gp_election_id`), not only through the position-plus-date proxy; that removed 3 seeded rows that had been escaping the gate, each enumerated in the validation file. The wave also added two tests: a nullout-rejection assertion on the seed key, and a coverage floor asserting that fallback rows whose election link supplies no usable seats still come out of the scorer with a score (measured 100% at introduction), which guards against the fallback silently becoming a no-op.

**G5, the ticket's acceptance population:** on the 11,687 November 2025 with-result campaigns, coverage moves from 62.9% (7,352 scored) to 95.6% (11,172 scored). The 3,820-campaign gain reconciles exactly against the fill mix on this same population: 3,646 filled by `race_exact`, 174 by `position`, 515 left with no fallback fill (fail-closed).

## Docs and metric ownership

The docs commit repoints the Viability Score 2.0 row in `canonical_metrics.md` from the retired TechSpeed model to `mart_civics.candidacy_scored.viability_score` (producer `int__civics_viability_scoring`). This is a source-pointer correction, not a definition change, but `canonical_metrics.md` is human-ratified, so the `Ratified` cell is left `pending` until the metric owner acks the new pointer.

## Follow-ups (not in this PR)

- **Archive election fill.** This PR only fills seats for the general-election path. Filling the archive path is a separate, consumer-impact ticket: `seats_available` is hashed into `gp_election_id` and ships downstream in reverse-ETL payloads, so changing it for archived elections needs its own review of who reads that field.
- **Opponent-count wiring.** Wiring fallback data into the opponent-count feature (`log_n_losers`) is deferred until after the win-rate reread. The tier and matched race id carried here are provenance evidence, not a starting point for it: `matched_br_race_id` is a position-plus-date pick, not verified against the race's candidacy roster, so it can name a race the candidacy does not appear on. That follow-up has to select races by roster membership and cannot reuse this column as a race link.

## Rollback

`git revert` the commits on this branch, then rebuild `int__civics_viability_scoring` and `candidacy_scored`. Also `drop table` the now-orphaned `int__civics_viability_seats_fallback` relation: reverting the model file removes it from future runs but does not drop the table already sitting in the warehouse. Because the scorer's MLflow model versions and WoE lookup tables can move between builds, a reverted build's scores will match the pre-change state only modulo those floating inputs, not byte-for-byte. The dated baseline snapshots taken before this change (`viability_scoring_baseline_0811a`, `candidacy_scored_baseline_0811a`) are the restore-evidence anchor for confirming a clean revert.

**Deployment gate.** Before the win-rate reread consumes the new scores, snapshot `int__civics_viability_scoring` and `candidacy_scored` in production immediately before the first post-merge rebuild, then rerun the G3-style parity aggregate against those snapshots: a full outer join on `gp_candidacy_id` counting lost (scored before, unscored after), changed (scored both sides, different value) and added. `lost = 0` and `changed = 0` are the gate; a nonzero either way holds the reread and triggers the rollback path above. The G4 published-value spot check runs on the same pair.


sanjayr1:
delegate review

PR #799, opened 2026-08-11
[DATA-2273] Haystaq loader: isolate per-state failures and self-heal unlogged S3 uploads

sanjayr1:
## Root cause this fixes

On May 27 the 00:02 scheduled run of `load__l2_haystaq_sftp_to_s3` started while the vendor was mid-upload of the May 2026 drop. It uploaded fresh files for 12 states to S3, then died on the 13th state's partial zip (`Extracted tab file not found`). Because the sync-log rows only append when the whole model succeeds, those 24 uploads were never logged — and the skip check (`file already in S3`) then treated them as complete on every later run. `load__l2_haystaq_s3_to_databricks` is log-driven, so the 12 states' Databricks tables kept serving the December 2025 drop while the other 39 states moved to May 2026 (ticket: DATA-2273).

## Changes

1. **Per-state error isolation with durable partial progress** (`_collect_load_details` / `_finalize_load_details`): one state's failure no longer aborts the loop, and the successful states' log rows persist that same run (failures are logged with a summary and retry on the next run via re-pull or self-heal). The run fails loudly only when *every* extraction failed — a systemic signal (e.g. SFTP outage) where there is nothing to persist anyway.
2. **Partial zips are retryable** (`_locate_extracted_tab`): a zip whose single `.tab` never materialized (vendor mid-upload — the exact May failure) is now a logged skip like the existing locked-zip and failed-extract paths. A zip with a wrong member count still raises (contract violation, becomes a collected failure).
3. **Skip requires a log row, not just S3 presence** (`_resolve_pickup_action`): a file in S3 with no sync-log row means a prior run died between upload and append; it is re-logged without re-download ("self_heal"). The logged name is the S3 key's exact suffix after the state prefix (S3 GETs are case-sensitive; downstream reads prefix + logged name verbatim).
4. **Canonical upload naming**: the S3 object is always uploaded under the zip-derived filename. Previously a zip whose extracted member name differed was uploaded under the extracted spelling, which no future run's skip check could ever find — an endless re-download/re-log loop (pre-existing bug).

## Adversarial review (independent model)

A sandboxed read-only review by a separate model returned five findings. Findings 1–3 (raise-at-end starves healthy states under a persistently failing state; self-heal could log a spelling that 404s downstream; the member-name mismatch loop) are fixed in the last commit and are what shaped changes 1, 3, and 4 above. Finding 4 (SFTP-wide outage costs ~17 min of per-state connection retries) is partially mitigated by the total-failure raise; a circuit breaker was declined as statefulness a twice-daily batch loader doesn't warrant. Finding 5 (failed old-version deletion hidden by self-heal) was declined: the consequence is unread storage cruft that the next vendor drop's download path deletes anyway.

## Testing

- 17 unit tests in `dbt/tests/test_load__l2_haystaq_sftp_to_s3.py` (TDD; house pattern of testing module helpers), including regressions for the exact incident (mid-loop failure not stopping later states; in-S3-but-unlogged resolving to self-heal) and for the review findings (partial failure persists successes; total failure raises; exact-spelling and nested-key propagation).
- Full dbt suite: 99 passed. `uv run mypy tests` clean. pre-commit (ruff, ruff-format) clean.
- CI cannot execute these models (`tag:load` is excluded from the CI job); prod-shape verification was done operationally during the incident remediation: the 12 states' log rows were backfilled and loaded via a targeted job run, and both sync logs were restored to full 51-state coverage.

## Deployment notes

- Both load models pin `full_refresh: false`, so the on-merge `state:modified+ --full-refresh` build runs them incrementally; the only dbt-graph descendant is `load__l2_haystaq_s3_to_databricks`.
- The sync-log restoration this change depends on (the 39 fresh states' rows, wiped by pre-#717 full refreshes) is already completed, so the first post-merge run's self-heal check is a no-op for healthy states.
- Known limitation: self-heal derives the expected filename from the latest zip on the SFTP, so an orphan whose zip has been rotated off the SFTP is not recovered (warning only). Acceptable for now; the vendor keeps drops up for months.

## Same-class risk in the sibling loader (deliberately out of scope)

`load__l2_sftp_to_s3` (the weekly VM2 uniform loader) has the same two structural patterns this PR fixes in the Haystaq loader: an S3-existence-only skip check and a per-state exception that aborts the whole run. It is exposed to the same upload-interrupted orphan class. Kept out of this PR per smallest-fix scoping (the quantified incident is Haystaq-only); the helpers here are written to be portable to it. Follow-up ticket recommended.

### DIFF
--- a/dbt/project/models/load/load__l2_haystaq_sftp_to_s3.py
+++ b/dbt/project/models/load/load__l2_haystaq_sftp_to_s3.py
@@ -77,22 +77,33 @@ def _select_latest_zip_file(file_list: list[str], file_pattern: re.Pattern) -> s
 
 def _resolve_pickup_action(
     tab_file_name: str,
+    s3_state_prefix: str,
     s3_keys: list[str],
     logged_file_names: set[str],
-) -> Literal["download", "self_heal", "skip"]:
+) -> tuple[Literal["download", "self_heal", "skip"], str | None]:
     """
     Decide how to handle the latest SFTP file. "In S3" alone is not proof of
     completion: a run that dies after uploading but before its sync-log append
     leaves the file in S3 with no log row, and the downstream Databricks load
     only consumes logged files. Such orphans must be re-logged ("self_heal"),
     not skipped.
+
+    Returns (action, matched_file_name). Matching is case-insensitive, but the
+    returned name is the S3 key's exact suffix after `s3_state_prefix`: S3 GETs
+    are case-sensitive and downstream reads prefix + logged name verbatim, so
+    only the spelling that actually exists in S3 may be logged.
     """
     tab_lower = tab_file_name.lower()
-    if not any(key.lower().endswith(f"/{tab_lower}") for key in s3_keys):
-        return "download"
-    if tab_lower in {name.lower() for name in logged_file_names}:
-        return "skip"
-    return "self_heal"
+    matched_key = next(
+        (key for key in s3_keys if key.lower().endswith(f"/{tab_lower}")),
+        None,
+    )
+    if matched_key is None:
+        return "download", None
+    matched_file_name = matched_key[len(s3_state_prefix) :]
+    if matched_file_name.lower() in {name.lower() for name in logged_file_names}:
+        return "skip", matched_file_name
+    return "self_heal", matched_file_name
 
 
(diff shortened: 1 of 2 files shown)


sanjayr1:
delegate review

PR #813, opened 2026-08-12
[DATA-2282] Map Clay County FL school board seats to the county electorate

danpelota:
## Root cause

A sales call reported that our votes-to-win number for Clay County FL School Board District 4 was 8,600, against an actual countywide requirement north of 20,000.

BallotReady has the position, the 2026 race exists, and it resolves both a district and a place, so the picker and race page work.

`m_election_api__position.district_id` for these seats points at the L2 `School_Board_District` (33,819 registered voters for District 4) rather than the county (173,368). election-api sizes the win number as `floor(projectedTurnout / 2) + 1` off the position's district, so District 4's 2026 General projection of 17,931 produced 8,966, which is the ~8,600 the reporter saw.

But Clay County school board members are elected countywide. Candidates run for a numbered residency district but every county voter votes in every district's race ([Ballotpedia](https://ballotpedia.org/Clay_County_District_Schools,_Florida)). BallotReady already models it that way, giving all five seats one `geo_id` (1200300, mtfcc G5420, the whole school district). The L2 Gold Match keyed on the district number in the position name instead.

## The fix

Five rows in `l2_br_match_overrides` pointing the Clay County school board positions at `County` / `CLAY`. This is the established repair for a seat whose officeholder must reside in a subdivision but whose electorate is the parent geography, alongside the Kansas City at-large, Georgia PSC, Yuma County commission, and Bridgewater-Raritan rows already in the seed.

Keyed at position grain because that is where the defect lives: the same L2 `School_Board_District` values stay correct for any other consumer, and the county district already exists with turnout projections, so nothing new is created.

## Deploy step

`int__zip_code_to_br_office` needs a one-time `--full-refresh`. The LLM path stops emitting these positions once an override is active and the override path re-emits them under the county, but an incremental merge cannot delete the old sub-district rows. Left in place they would put two `district_type`s on one position, failing `assert_zip_to_br_office_one_district_type_per_br` and letting `pct_districtzip_to_zip` exceed 1.0. This caveat is already documented on the seed in `seeds_schema.yaml`.

## Verification

`m_election_api__position` built incrementally in dev, then diffed against prod on `district_id`. Baseline before the change was 0 differing rows, so the diff is entirely this change:

| Position | Before | After |
|---|---|---|
| Clay County School Board - District 1 | CLAY CNTY SCHL BD DIST 1 (33,694) | CLAY (173,368) |
| Clay County School Board - District 2 | CLAY CNTY SCHL BD DIST 2 (32,830) | CLAY (173,368) |
| Clay County School Board - District 3 | CLAY CNTY SCHL BD DIST 3 (31,187) | CLAY (173,368) |
| Clay County School Board - District 4 | CLAY CNTY SCHL BD DIST 4 (33,819) | CLAY (173,368) |
| Clay County School Board - District 5 | CLAY CNTY SCHL BD DIST 5 (41,499) | CLAY (173,368) |

Five differing rows, no rows added or dropped (286,419 in both).

Resulting win numbers for 2026, using election-api's `floor(turnout / 2) + 1`:

| Position | General before | General after | Primary before | Primary after |
|---|---|---|---|---|
| District 1 | 9,713 | 44,487 | 3,132 | 13,948 |
| District 2 | 6,332 | 44,487 | 1,680 | 13,948 |
| District 3 | 8,510 | 44,487 | 2,833 | 13,948 |
| District 4 | 8,966 | 44,487 | 2,850 | 13,948 |
| District 5 | 10,969 | 44,487 | 3,456 | 13,948 |

`int__zip_code_to_br_office` full-refreshed in dev and `m_election_api__zip_to_position` rebuilt on top. Zip coverage for the three Clay seats with a future race went from 3, 8, and 3 sub-district zips to 12 countywide zips each. Districts 2 and 3 have no rows in either environment, since the mart only spans elections in the next two years.

All tests pass, including `assert_zip_to_br_office_one_district_type_per_br`, `assert_zip_to_br_office_voters_in_zip_invariant`, `assert_override_positions_have_zip_coverage`, and `assert_override_seed_resolves_to_position_district`.

The wider dev-vs-prod diff on the zip mart shows 81,189 changed `pct_districtzip_to_zip` values and about 12,900 added or removed rows. That is not this change: a stale `dbt_dball.int__zip_code_to_l2_district` from 2026-08-07 meant the dev build read an L2 snapshot five days behind prod's 2026-08-12. Grouping the zip table by `br_database_id` and comparing the set of districts each position resolves from returns exactly 5 changed positions, all five Clay seats. Everything else that differs is voter-count drift from the older snapshot.


danpelota:
delegate review

danpelota:
delegate review

PR #814, opened 2026-08-13
[DATA-2107] Land day-type and turnout projection columns on the race mart

sanjayr1:
**The five NEW columns are delivery-inert until they join RACE_COLUMNS (a follow-up PR) — the loader selects a fixed column list. The mart's EXISTING delivery surface (row membership, existing columns) continues to swap nightly as with any mart change; the membership-parity validation below plus the mart's grain tests are what bound that surface. Merge order: after the turnout-inference PR's post-merge gate passes — that gate PASSED 2026-08-13 (green nightly run 70471892610579; full-key propagation exact on all 744,954 keys; recorded in the repo validation file).**

## Design

- Adopts the BallotReady primary-day calendar model `int__election_calendar_primary_ballotready` from the closed calendar-lookup branch at its head (research-authored adoption commit + a scoped description cleanup). One row per (state, even year): the derived statewide primary date, with November-general collisions excluded by construction so the General/Primary branches are disjoint. The superseded pieces of that branch (seed, calendar mart, its tests, the sync group) are deliberately not adopted.
- `m_election_api__race` gains a `race_projection_key` CTE that derives the day-type tag and the projection join key together from one resolved traversal, so tag and join can never disagree about whose primary day applies. The state that picks the calendar row is the resolved district's (override-corrected) state where a district exists, the race's own state otherwise. Day rule: even-year Tuesday-after-first-Monday-in-November → `General`; a date matching the state's derived primary day → `Primary`; everything else → `LocalOrMunicipal`. The row lands the API's `ElectionCode` enum spelling; the join speaks the model-output vocabulary (`Local_or_Municipal`). `ConsolidatedGeneral` is deliberately never emitted — it is a legacy serving-path concept.
- Five landed columns: `election_code`, `projected_turnout`, `projected_turnout_lower`, `projected_turnout_upper` (int-cast at the landing site for the Postgres delivery contract), and `inference_at` (freshness stamp, NULL exactly where the projections are NULL). Values join directly from `int__voter_turnout_inference` on (district tuple, year, day type). Years outside the model horizon and districts without model coverage simply do not join: those races carry NULL projections by design — no fallback, nothing invented.
- The CTE repeats the mart's serving-window predicate (cross-referenced comments keep the two in sync) so the second traversal of the race view stays bounded to the serving window.

## Gates

- yaml: `not_null` + `accepted_values` on `election_code`; `accepted_range` (min 3, the model's serving floor) on all three value columns.
[...]

PR #823, opened 2026-08-14
[DATA-2107] Harden the calendar and race-window plumbing

sanjayr1:
Fast-follow hardening for the race-turnout columns change (#814), folding the actionable items from that PR's two independent pre-merge reviews. Every item is additive or refactor-equivalent: **served data is byte-identical** (verified below).

## What and why

**Calendar hardening**
- `int__election_calendar_primary_ballotready` is now `materialized="table"` (justified override): several consumers read it per build (the race mart's key CTE, its own tests, downstream singular tests), and a pinned snapshot keeps build-time and test-time reads coherent — as a view, each consumer re-executed the derivation against live vendor data, so a mid-build BallotReady sync could make the day-rule test disagree with the mart it validates.
- Both candidate CTEs now filter `state is not null`: a NULL-state vendor artifact (national/party-level rows exist in sibling BR feeds) would duplicate through the full outer join's non-null-safe key and fail the calendar's grain tests loudly over ignorable noise.
- The `state` column gains an `accepted_values` postal-vocabulary guard via the repo's canonical `get_us_states_list` macro — deliberately a format guard, NOT a completeness test (vendor forward-coverage gaps are expected reality; a format drift like full state names silently un-matching every join is not).
- New anchor test pins Wisconsin's derived primary day for the three completed even cycles (2020-08-11, 2022-08-09, 2024-08-13 — L2-ground-truth-confirmed, immutable historical facts). WI runs the calendar's only fewest-races-wins rule, making it uniquely junk-sensitive: one stray low-count "Primary"-named row would flip the pick; this fails loudly instead.

**Single-definition plumbing on the race mart**
- The serving-window predicate is now defined once (`{% set serving_window_predicate %}`) and used at both former sites, so a future window edit cannot drift the two copies apart (the drift mode: a one-sided widening silently drops races from the served table via the inner key join).
- The federal November-election-day expression now lives in one macro (`november_general_election_day`) used by the calendar and the mart CASE. The day-rule singular test deliberately KEEPS its own inline copy so its recompute stays independent of the code it checks.

## Verification

- `dbt build` over both models: 31/31 green — including both new tests and both pre-existing singular tests (the day-rule test passing proves the macro'd CASE still agrees with its independent inline recompute).
- Calendar confirmed built as a table; dev race mart row count 1,013,467 (grain untouched).
- Task-scoped adversarial review: approved, zero blocking findings; it independently verified the macro call sites compile to token-identical SQL and the shared-predicate substitution is whitespace-inert.

Not included (deliberately, per the merged design's adjudication): any calendar completeness gate or mart-level coverage floor — those remain the standing-dissent items from the original plan gate, and a possibly-viable window-filtered coverage-spine formulation is left for its own ticket.

---

> [!NOTE]
> **Medium Risk**
> Changes how primary dates and election_code (General/Primary) are derived in the race mart, but the PR positions this as hardening with byte-identical served output; drift in the shared window predicate would silently change which races are served if edited incorrectly.
> 
> **Overview**
> **Hardens BallotReady primary-calendar and race-mart plumbing** so build-time reads stay consistent and duplicated logic cannot drift, without changing served race data (refactor-equivalent per PR verification).
> 
> **Calendar (`int__election_calendar_primary_ballotready`):** Materializes as a **table** so the race mart, model tests, and singular tests all read the same snapshot during a build. Both primary-candidate CTEs now drop **`state is null`** vendor rows. **`state`** gets an **`accepted_values`** guard via `get_us_states_list`. November general-day exclusion uses the new **`november_general_election_day`** macro instead of inline SQL.
> 
> **Race mart (`m_election_api__race`):** The **6-year past / 2-year future** filter is defined once as **`serving_window_predicate`** and reused in `race_projection_key` and the final `where`. **General** election detection uses the same **`november_general_election_day`** macro.
> 
> **New:** `november_general_election_day` macro (2 U.S.C. §7 Tuesday-after-first-Monday-in-November). Singular test **`assert_election_calendar_wisconsin_known_dates`** pins WI partisan primary dates for 2020, 2022, and 2024.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff94f8d99d87c523fc617ae7b23ee356326d46d4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sanjayr1:
## Preview verification (head `ff94f8d9`)

- CI run `70471892790657`: Success at the exact head. Delegate reviewer: APPROVED.
- **Served-data neutrality proven:** full-payload comparison (every column, `to_json(struct(*))`) of prod vs this preview across all 1,013,467 races — **zero differences on all five turnout columns** (`election_code`, `projected_turnout`, `_lower`, `_upper`, `inference_at`) and zero membership changes. The 60 rows that differ do so ONLY on civics-fed attributes (`number_of_seats` 40, `is_partisan` 23, office fields ~17, overlapping) — the same overnight civics generation-skew visible on the #814 refresh (prod's mart predates the latest nightly's civics rebuild); unrelated to this change.
- The calendar built as a TABLE in the preview; the WI anchor test, the postal-vocabulary guard, and both pre-existing singular tests are green — the day-rule test passing confirms the macro'd November expression compiles identically to its former inline copy.

Status: CI green, delegate-approved, neutrality proven. Unmerged — owner decides.

dbt/project/tests/assert_election_calendar_wisconsin_known_dates.sql
danpelota, line 1: Hmm...why do we need this? Is it possible for this to actually happen?

sanjayr1: Fair challenge — it's possible but remote: a row would have to carry a 'Primary' name that dodges the Runoff/Consolidated/Special/Presidential filters with a race count below the August partisan primary's, and nothing like it appears anywhere in the 2018-2030 vendor data. The pin's stronger role was anchoring the fewest-races exception against future code edits, and the dates it protects are mostly aging out of the serving window anyway. Agreed it's below the clutter bar — dropped in d621bf22. The calendar's structural guards (grain uniqueness, even-years-only, and the new postal vocabulary check) are unaffected.

sanjayr1:
delegate review

PR #825, opened 2026-08-14
Fix candidacy_hubspot nested-loop anti-join (~25 min -> ~30 s)

hhkarimi:
## What

`candidacy_hubspot` was the single most expensive node in the project: **45.6 execution-hours over the last 30 days**, across 154 full rebuilds averaging **17.6 minutes** each (max 34).

The cost was not data volume. Each rebuild reads ~460 MB and writes ~0.65 MB. It was one predicate — the not-already-in-HubSpot de-dupe, which ORs three different join keys together:

```sql
and not exists (
    select 1 from int__hubspot_contacts hs
    where (... lower(trim(hs.email)) = lower(trim(c.email)))
       or (... regexp_replace(hs.phone_number,'[^0-9]','') = regexp_replace(c.phone_number,'[^0-9]',''))
       or (... hs.br_candidacy_id = try_cast(br_int.br_candidacy_id as int))
)
```

Spark cannot hash-join an OR across disjoint keys. `EXPLAIN` confirms it planned as `PhotonBroadcastNestedLoopJoin LeftAnti` — every other join in the plan is a hash join.

| | |
|---|---|
| Candidacies in the 30-day window | 240,293 |
| ...surviving the other `where` predicates, i.e. entering the anti-join | 28,106 |
| HubSpot contacts | 309,101 |
| **Comparisons per rebuild** | **~8.7 billion** |
| Rows emitted | 8,668 |

with `regexp_replace` re-evaluated on both sides per pair.

## How

Normalize each key once into its own `distinct` CTE, then apply three hash anti-joins.

Row semantics are unchanged:
- each key set is `distinct`, so the left joins cannot fan out;
- candidate-side keys are `null` exactly where the original OR leg did not apply, and `null` never equals a key, so those rows fall through identically;
- the phone `>= 10` digit guard moves to the HubSpot side too, which is equivalent — the original required it on the candidate side, and the two values must be equal to match.

## Results

| | before | after |
|---|---|---|
| model build | ~25 min | **~30 s** |
| `dbt build --select candidacy_hubspot` (model + 9 tests) | — | **38 s** |

At current cadence (10–17 rebuilds/day, up from 1–4 in late July) this is roughly **$275/month** of serverless SQL, and it stops being the long pole that delays the rest of the DAG.

## Verification

- `dbt build --select "candidacy_hubspot"` — `PASS=8 WARN=2 ERROR=0`. Both warnings are pre-existing (`assert_candidacy_hubspot_candidate_code_collision` at 94 rows, matching production exactly; the `trim` expression test at 1 row).
- Ran the old and new filter forms side by side against production data over the **full eligible population (28,106 rows)**: **8,671 rows each, zero rows in either symmetric difference**. A 3,000-row sample run was likewise exact (901 each).
- Dev build emits 8,673 rows vs production's 8,668 — consistent with the 30-day `current_date()` window and upstream drift between the two build times, not a semantic change.
- `pre-commit run --all-files` passes.

### Note on `incremental`

Deliberately not used. The output is a rolling *exclusion* feed — rows must leave as they age out or get claimed into HubSpot — so an append would retain stale leads and a merge-with-deletes rescans anyway. The cost was never in the write (0.65 MB/rebuild); it was in the join plan.

## Also: cost-audit skill enhancements

The audit skill ranked nodes in a way that hid this. Changes so the next occurrence is obvious:

- **`hr/TB` ratio + `scan-bound` / `compute-bound` label** on the top-nodes table. This node reads 635 hr/TB; the L2 snapshot beside it reads 2. That single column separates "reads too much" (wants pruning) from "bad join plan" (pruning won't touch it).
- **`>60s` statement count.** dbt stamps trivial `SHOW`/`OPTIMIZE` statements with the same `node_id` as the build, so this node showed ~800 statements for ~150 real rebuilds. Raw counts overstate build frequency ~5x and misled the first pass of this investigation.
- **Fixed test node names rendering as bare hashes** (`0607546a35`) — `node_id` carries a trailing hash for generic tests, so taking the last segment lost the name.
- **Documented** the nested-loop antipattern and its `EXPLAIN` check, why `incremental` is the wrong tool for compute-bound or rolling-window nodes, `execution_duration_ms` vs `total_duration_ms` (the latter includes queue and provisioning time), and that `target_name` separates prod from PR-CI runs.

For context, the skill's existing view-test guidance already worked: the July change materializing these marts as tables dropped dbt test hours from 60.9/week to 10.5/week, and tests are now only ~5% of Databricks spend.

### DIFF
--- a/.claude/skills/dbt-databricks-cost-audit/SKILL.md
+++ b/.claude/skills/dbt-databricks-cost-audit/SKILL.md
@@ -47,7 +47,7 @@ commit generated `cost_dashboard.*`; commit only the tool.
 | cost by product / month + week | `system.billing.usage` × `list_prices` | overall trend, SQL vs ALL_PURPOSE vs APPS… |
 | dbt exec-hr by node type | `system.query.history` (`statement_text` node_id) | **testing vs model vs snapshot cost** |
 | warehouse exec-hr by consumer | `system.query.history` `client_application` | dbt vs airbyte vs BI vs dev |
-| top dbt nodes | `system.query.history` | the specific expensive models/tests |
+| top dbt nodes | `system.query.history` | the specific expensive models/tests, and whether each is scan- or compute-bound |
 | high-freq-job estimate | exec-hr in the 4-hourly-only UTC windows ×6/4 | is killing it sufficient? |
 
 $ for the dbt/consumer views is the **SQL-warehouse bill prorated by query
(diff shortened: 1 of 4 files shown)


hhkarimi:
### Full-population equivalence check: exact match

Ran the current production filter form and the rewritten form side by side against the same production snapshot, over the full eligible population rather than a sample:

| | rows |
|---|---|
| eligible rows entering the anti-join | 28,106 |
| old form (`not exists` with OR'd keys) | 8,671 |
| new form (three hash anti-joins) | 8,671 |
| in old but not new | **0** |
| in new but not old | **0** |

Compared on the full projected tuple (`gp_candidacy_id`, `email`, `phone_number`, `br_candidacy_id`) with `except all`, so duplicates are counted too, not just set membership.

### Correction to the PR description

The original description said ~74.3 billion comparisons per rebuild. That used the 30-day window count (240,293) as the anti-join input, but the remaining `where` predicates — email-or-phone present, non-major-party, `hubspot_contact_id is null` — cut it to **28,106** before the join. The correct figure is **~8.7 billion**, and the description has been updated.

The diagnosis and the fix are unaffected: it was still a `BroadcastNestedLoopJoin LeftAnti` evaluating `regexp_replace` per pair, and the measured build time (~25 min to ~30 s) is empirical.

PR #826, opened 2026-08-15
[DATA-2325] Materialize all election-api mart models as tables

danpelota:
Closes DATA-2325.

Every model in `models/marts/election_api/` is now a plain table with no model-level `{{ config() }}` block. 13 models, 8 config blocks removed.

## Why

The incremental machinery in this directory is legacy. The dbt writer was retired and all 13 election-api tables load through the Airflow set-wise atomic swap DAG, which reads the full mart on every run. The watermark filters and `{{ this }}` self-joins in `issue`, `district`, `place`, and `position` no longer serve a purpose, and an incremental mart feeding a full-read loader is a source of drift rather than a saving.

The remaining four config blocks (`candidacy`, `race`, `projected_turnout`, `stance`) were already `table`, and every key in them was redundant:

- `materialized` — `marts: +materialized: table` in `dbt_project.yml` covers the subtree
- `auto_liquid_cluster` — already set catalog-wide on `goodparty_data_catalog`
- `unique_key`, `incremental_strategy`, `on_schema_change` — meaningless once the model is not incremental

No `dbt_project.yml` change was needed. The `marts` default already applies to this directory, so adding an `election_api:` key would only restate it.

## What changed per model

`district` — dropped three upstream watermark filters, the self-join to `{{ this }}`, and the `coalesce(existing.created_at, now())` fallback.

`place` — dropped the wall-clock `updated_at` bump and the incremental self-join filter. Also removed `source_updated_at`: it existed only to keep the watermark in the source-time domain, its own description said it was not published to election-api, and the sync DAG loads the live Postgres table's own columns, so nothing consumed it.

`position`, `issue` — dropped their watermark filters.

`candidacy`, `race`, `projected_turnout`, `stance` — config block removed, no SQL change.

No tags existed anywhere in the directory.

## Verification

Full mart build in a dev schema. All 13 models built as tables.

Wall clock for 13 models and 153 tests: **2m51s** on 4 threads (433s of model time, 171s of test time). `stance` was skipped in that run behind a failed upstream test and built separately in 10.2s, 9 tests passing.

| model | seconds |
|---|---|
| district_top_issues | 155.0 |
| projected_turnout | 39.1 |
| race | 37.6 |
| district | 37.1 |
| place | 32.9 |
| zip_to_position | 27.4 |
| candidacy | 23.0 |
| position | 22.3 |
| elected_official_support | 18.9 |
| person | 15.5 |
| office_holder | 13.0 |
| issue | 11.6 |
| stance | 10.2 |

`district_top_issues` dominates the critical path at 155s; everything else is under 40s.

152 of 153 tests passed. Two notes, neither caused by this change:

**`not_null_m_election_api__candidacy_gp_candidate_id`** failed in dev with 6,890 nulls against its `error_if: ">1000"` threshold. This is stale dev state, not the materialization change. Dev and prod both produce exactly 248,480 candidacy rows, but prod has 616 nulls. Of the 6,890 dev nulls, 6,274 resolve in prod and the remaining 616 are prod's own known nulls. Cause: the dev schema held an old copy of `int__civics_person_canonical_ids` (951,339 rows / 460,669 BallotReady keys) so the run used it instead of deferring to prod (1,085,714 / 466,807). `gp_candidate_id` reads only that crosswalk, which this PR does not touch.

**`eos_support_never_exceeds_total_constituents`** warned on 2 rows. It is configured `severity: warn` and matches prod.

## Operational note for the cutover

The first production run rebuilds `place`, `district`, `position`, and `issue` from scratch rather than merging. Two consequences worth watching on that run:

- `place` and `district` will drop any rows that accumulated incrementally but no longer derive from current upstream. That is the intended correction, but it means the swap DAG's `min_id_overlap` quality gate sees its largest movement on this run.
- `district` recomputes `created_at` as `now()` for every row, since the `coalesce(existing.created_at, ...)` path is gone. The column becomes build time rather than first-seen time. Nothing downstream reads it, and the swap replaces the table wholesale either way.

---

> [!NOTE]
> **Medium Risk**
> Cutover rebuilds formerly incremental marts and changes `place`/`district` timestamp semantics; first production swap may show large overlap movement and shed rows that only existed under incremental merges.
> 
> **Overview**
> **Aligns election-api marts with full-table loads** by stripping model-level `{{ config() }}` blocks and incremental machinery across `models/marts/election_api/`. The Airflow swap DAG already reads entire marts each run, so watermark filters and `{{ this }}` self-joins were drift risk, not savings.
> 
> **`district`, `issue`, `place`, and `position`** no longer slice upstream on `updated_at` / `loaded_at` / `inference_at`, and no longer merge against prior mart rows. **`place`** also drops the wall-clock `updated_at` bump for slug changes and the unpublished `source_updated_at` column (and its schema test). **`district`** always sets `created_at` to `now()` instead of preserving first-seen timestamps via a self-join.
> 
> **`candidacy`, `race`, `projected_turnout`, and `stance`** only lose redundant config; SQL is unchanged. First prod run after cutover fully rebuilds the four formerly incremental tables—expect possible row churn vs stale incremental state and a larger `min_id_overlap` movement on the swap DAG.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 171dcbfd52cfc43bc1ba5605812c53b2e59fd763. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/project/models/marts/election_api/m_election_api.yaml
+++ b/dbt/project/models/marts/election_api/m_election_api.yaml
@@ -192,17 +192,6 @@ models:
         tests:
           - not_null
           - unique
-      - name: source_updated_at
-        description: >
-          Un-bumped source updated_at, carried separately from updated_at so
-          the model's incremental watermark stays in the source-time domain.
-          updated_at is bumped to wall-clock when a row is new to the mart or
-          its slug changed (the election-api write model's publish gate);
-          gating ingestion on those bumped values would skip later source
-          updates whose BallotReady stamps lag wall-clock. Not published to
-          election-api (the writer's Place upsert selects explicit columns).
-        tests:
-          - not_null
 
   - name: m_election_api__position
     description: "Mart model that serves the Election API"
--- a/dbt/project/models/marts/election_api/m_election_api__candidacy.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__candidacy.sql
@@ -1,12 +1,3 @@
-{{
-    config(
-        materialized="table",
-        unique_key="id",
-        auto_liquid_cluster=true,
-    )
-}}
-
-
 with
     active_candidacy as (
         select
--- a/dbt/project/models/marts/election_api/m_election_api__district.sql
+++ b/dbt/project/models/marts/election_api/m_election_api__district.sql
@@ -1,29 +1,14 @@
-{{
-    config(
-        materialized="incremental",
-        unique_key="id",
-        on_schema_change="fail",
-        auto_liquid_cluster=True,
-    )
-}}
-
 with
     turnout_districts as (
         select distinct
             state, district_type as l2_district_type, district_name as l2_district_name
         from {{ ref("int__model_prediction_voter_turnout") }}
-        {% if is_incremental() %}
-            where inference_at >= (select max(updated_at) from {{ this }})
-        {% endif %}
     ),
     l2_data as (
         select
             state_postal_code,
             {{ get_l2_district_columns(use_backticks=true, cast_to_string=true) }}
         from {{ ref("int__l2_nationwide_uniform") }}
-        {% if is_incremental() %}
-            where loaded_at >= (select max(updated_at) from {{ this }})
-        {% endif %}
     ),
     l2_data_districts as (
         select distinct
@@ -44,9 +29,6 @@ with
             'State' as l2_district_type,
             state_postal_code as l2_district_name
         from {{ ref("int__l2_nationwide_uniform") }}
-        {% if is_incremental() %}
-            where loaded_at >= (select max(updated_at) from {{ this }})
-        {% endif %}
     ),
     unioned_w_id_districts as (
         select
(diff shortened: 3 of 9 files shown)


danpelota:
CI note: the dbt Cloud check is red on `assert_candidacy_stage_one_person_per_br_candidate` (1 result). This is not caused by this PR.

**Same failure on an unrelated PR.** Run `70471892918563` for #822 (BR enrichment incrementals, sha `cb53b2e`) failed on the identical test about 15 hours before this branch existed.

**This PR cannot change that test's inputs.** CI step 4 runs `dbt build --full-refresh --select="state:modified+"`. Under `--full-refresh`, `is_incremental()` is already false, so on `main` the four converted models render exactly the `{% else %}` branches this PR hardcodes. Per model, `main` under that command and this branch emit the same SQL:

- `issue`, `position` — watermark filters render to nothing either way
- `district` — no filters, `now() as created_at`, no `{{ this }}` join either way
- `place` — `tbl_place.updated_at`, no join or where either way

The only content delta anywhere in the diff is `place.source_updated_at` being dropped, and nothing references it (`grep` across `dbt/`, `airflow/`, and `people-api-loader/` returns only the model and its own description).

`candidacy_stage` reaches this PR's diff through `m_election_api__position` -> `campaigns`, which is why the test was selected at all, but `position` is byte-identical under `--full-refresh`.

The underlying condition looks like a recent regression in the civics person-resolution chain rather than anything in this directory. It needs its own ticket; flagging rather than folding a civics fix into a materialization change.

danpelota:
delegate review

PR #827, opened 2026-08-15
[DATA-1964] Move the L2 voter-file and Haystaq loads from dbt into Airflow

danpelota:
Moves every L2 SFTP load out of dbt python models and into one Airflow DAG,
`load_l2_voter_files`. The whole `dbt/project/models/load/` directory goes, along with the
separate `l2_expired_voters` DAG. Net: 2,887 lines deleted, 1,487 added.

Companion terraform is thegoodparty/gp-terraform-dataplatform#52.

## Why

The dbt loader raised on the first state whose archive was briefly absent from SFTP, which ended
the run and left every state after it in the loop unsynced. That matches the load history: about
one state a day against a cadence implying seven. Per-file task mapping isolates that, and a state
L2 has pulled while rebuilding is now simply absent from the plan.

It also ran python-on-Spark on an all-purpose cluster with static S3 access keys, and kept four
append-only load logs to decide what to do next.

## What it does

Four tasks. Nothing is recorded between runs — both plans are derived from live state, so a run
that dies partway is repaired by the next one and a night with no new files does nothing.

```
plan_sources ──> sync (mapped, 1 at a time) ──┐
                                              ├─> plan_table_loads ──> load (mapped, 8 at a time)
                                              ┘   (trigger_rule=all_done)
```

1. `plan_sources` — lists L2's SFTP dirs, diffs against what S3 already holds. A source is pending
   if any member is missing or older than the SFTP mtime.
2. `sync` — one mapped task per archive. Downloads the zip, streams the members we keep straight
   into S3. Data dictionaries get their preamble and trailing legend trimmed.
3. `plan_table_loads` — newest staged file per table, minus tables already built from it (S3
   `LastModified` against `information_schema.tables.last_altered`).
4. `load` — one mapped task per table: `CREATE OR REPLACE TABLE ... AS SELECT * EXCEPT
   (_rescued_data), current_timestamp() AS loaded_at FROM read_files(...)` on a SQL warehouse.

`SOURCE_GROUPS` is the only place that knows what L2 publishes. Each entry states its family's
grammar once — the directory, a stem regex capturing the state as `folder` and the publication as
`version`, the archive suffix, and the members to keep mapped to the file type each becomes (which
is also the destination table's suffix). The archive patterns, the staged-file classifier and the
table names all derive from it. That is what makes the Haystaq feeds two entries rather than a
second loader, so roughly 680 lines of duplicated dbt python are deleted rather than ported.

| Group | Directory | Archive | Tables |
|---|---|---|---|
| `VM2` | `/VMFiles` | `VM2--CA--2026-08-05.zip` | `l2_s3_ca_demographic`, `_vote_history`, + 2 dictionaries |
| `VM2Uniform` | `/VM2Uniform` | `VM2Uniform--CA--2026-08-05.zip` | `l2_s3_ca_uniform`, `_uniform_data_dictionary` |
| `HaystaqFlags` | `/L2-Haystaq Issue Model Scores` | `ca_haystaqdnaflags_20260520.tab.zip` | `l2_s3_ca_haystaq_dna_flags` |
| `HaystaqScores` | `/L2-Haystaq Issue Model Scores` | `ca_haystaqdnascores_20260520.tab.zip` | `l2_s3_ca_haystaq_dna_scores` |

Expired voter IDs stop being a special case. They used to land in their own `airflow_source` schema
via a bulk `MERGE` in python; the file now rides the same path as everything else into
`l2_s3_expired_voters`, and the dedupe and blank-row filter move into the staging model, which is
where this repo puts that work.

Reading the server settled two things the old code got wrong. Its two Haystaq directories are the
same 104 files, so the flags/scores directory split bought nothing. And L2 leaves superseded
publications in place, so transferring anything but the newest per state was wasted bandwidth on
files `plan_table_loads` would discard anyway.

## Things that look odd but are deliberate

- **`sync` is `max_active_tis_per_dag=1` on its own queue; `load` is 8.** Only `sync` touches local
  disk. `load` just issues SQL to the warehouse — no bytes pass through the worker.
- **Only the compressed archive hits disk.** Members stream out of the zip to S3. CA's Haystaq
  scores zip is 7.12 GB but its member is 19.8 GiB. Peak disk is the zip, never the contents.
- **`inferColumnTypes => false`.** L2 carries zero-padded codes (ZIP, ZipPlus4, DPBC, FIPS).
  Inference turns `01234` into `1234` before any model sees it. Haystaq needs the same:
  `LALVOTERID` is the join key into `int__l2_nationwide_uniform_w_haystaq`, and a dropped leading
  zero on one side silently misses the join.
- **`_SequentialReader` exposes only `read()`.** `ZipExtFile` fakes `seek` by re-decompressing from
  the start of the member, which boto3's seekable upload path would trigger once per part.
- **Missing SFTP mtime means always re-transfer.** Defaulting to the epoch would mark a source
  permanently up to date.
- **`plan_table_loads` is `trigger_rule="all_done"`.** The Databricks side reads S3, not the sync
  results, so one state failing to copy must not block the states that succeeded.
- **`dbt_source_haystaq` is pinned to `dbt_source`** rather than a var. That var only worked
  because the deleted loader read the same one to pick its write target, so one knob moved writer
  and reader together. With the writer in Airflow it could only move the reader, which is a silent
  mismatch.

## Verification

Run end to end on `astro-dev` into a scratch S3 prefix and scratch schema, 2026-08-14/15.

| Run | Scope | Result |
|---|---|---|
| dry_run | all, before Haystaq folded in | 95 sources found, 0 tables to build |
| real | WY + EXPIRED | 3 syncs, 7 tables |
| real | + RI | 2 syncs, 13 tables |
| real | CA | 2 syncs (6.00 GB, 6.84 GB), 6 tables |
| real | HaystaqFlags + WY | 1 sync, 1 table |
| real | HaystaqScores + WY | 1 sync, 1 table |

- **Row counts match prod exactly** where L2 has not republished the source: CA demographic and CA
  uniform both 22,847,425, identical to `dbt_source`.
- **Reads are byte-exact.** WY demographic: 338,073 physical lines = 338,072 rows + header, 0
  rescued. CA vote history: 22,811,625 physical lines = 22,811,624 rows + header.
- **Both Haystaq tables are identical to prod's, not merely similar.**

  | Table | Rows (new = prod) | Columns (new = prod) | Rows differing either direction |
  |---|---|---|---|
  | `l2_s3_wy_haystaq_dna_flags` | 339,827 | 163 | 0 |
  | `l2_s3_wy_haystaq_dna_scores` | 339,827 | 328 | 0 |

  The last column is a two-way `EXCEPT` over every column but `loaded_at`.
- **Zero-padded codes survive as strings** — `ZipPlus4='0060'`, `Voters_FIPS='015'`.
- **Data dictionaries trim correctly** — first row is `1, LALVOTERID, 50`, no preamble or legend.
- **Largest archive works.** CA's 6.84 GB zip passed the disk precheck and completed.

`plan_sources` against the live server returns 195 sources — 46 VM2, 46 VM2Uniform, 51
HaystaqFlags, 51 HaystaqScores, 1 expired, 179.9 GB.

Tests: 25 for the L2 loaders, 89 passed / 2 skipped for the airflow suite. `pre-commit run
--all-files` clean.

### Row counts below prod on some states are L2, not us

WY, RI and CA vote history came up short against `dbt_source`. L2 republishes archives under the
same dated filename — `VM2--CA--2026-08-05.zip` has an mtime of 2026-08-10, days after prod loaded
it on 2026-08-07. This DAG compares SFTP mtime against S3 and picks those up; the old name-keyed
loader did not. Byte-exactness was verified above before concluding this.

### Task kills in dev are environmental

`astro-dev` has `is_development_mode: true` (prod does not) and kills task pods mid-task, which
Airflow reports as "Process terminated by signal. Likely out of memory error (OOM)." It is not
memory: the rate held at 2-3 per 8 concurrent tasks across `default_task_pod_memory` of 0.5Gi, 1Gi
and 2Gi, across an X-Small and a Medium warehouse, hitting tiny data-dictionary tables as readily
as multi-GB ones. The unrelated `sync_election_api` DAG shows the same in dev. `retries=2` absorbs
it.

## Before this can run in prod

The DAG ships paused (`is_paused_upon_creation=True`) on a `0 8 * * *` schedule. Merging is safe;
these are the cutover steps.

1. **Prod's `l2_s3_bucket` must be `goodparty-warehouse-databricks`, not the old bucket.** Prod's
   L2 data lives at `s3://goodparty-databricks-workspace/l2_data/...`, which has no Unity Catalog
   external location, so `read_files` on a SQL warehouse cannot read it (`[INSUFFICIENT_PERMISSIONS]
   ... does not have permission SELECT on any file`). The old loader got away with it by reading
   from an all-purpose cluster with static keys. `goodparty-warehouse-databricks` already has a
   bucket-root external location, so no new one is needed. Existing prod data under the old bucket
   is not migrated — the first prod run re-syncs from SFTP, which is intended. Haystaq's old prefix
   is likewise not migrated; it now shares the voter-file prefix and layout.
2. **Terraform (gp-terraform-dataplatform#52) lands first** — the `l2-voter-files` worker queue,
   `READ_FILES` on the storage external location for both Airflow SPs, and a
   `gp-l2-voter-files-<env>` role to replace the hand-made static-key IAM user.
3. **Prod needs `l2_sftp`, `aws_default` and `databricks` connections** plus the four variables
   (`l2_s3_bucket`, `l2_voter_files_s3_prefix`, `l2_voter_files_databricks_schema`,
   `databricks_conn_id`).
4. **dbt Cloud job `70471823543327` ("Haystaq load models smoke test") breaks on merge** — it
   selects two models this PR deletes. Delete the job. The project env vars `DBT_L2_SFTP_HOST` /
   `_PORT` / `_USER` and `DBT_S3_ACCESS_KEY` become unused, as does the `l2-sftp-password` secret
   in the `dbt-secrets-prod` Databricks scope. Keep `l2_state_allowlist` and
   `l2_haystaq_databricks_schema` — the intermediate models and source definitions still use them.
5. **Plan for a long first run.** The initial Haystaq backfill is 78.4 GB on a concurrency-1 queue,
   on top of L2's ~117 GB, and `max_active_runs=1` skips the next night while it goes. It happens
   once. Consider seeding a few states at a time with the `folders` param, and a bigger warehouse
   for the first load — CA's Haystaq scores file is 19.8 GiB of TSV and the Serverless Starter is
   X-Small.

---

> [!NOTE]
> **High Risk**
> This is a cutover of core voter-file infrastructure (~300+ `l2_s3_*` tables), with prod depending on new bucket/IAM/UC permissions, a long initial backfill, and removal of dbt load jobs—misconfiguration would leave dbt reading empty or stale sources.
> 
> **Overview**
> **Moves all L2 ingestion off dbt Spark and the dedicated expired-voters DAG into a single paused daily DAG, `load_l2_voter_files`.** It plans SFTP vs S3, syncs one archive at a time on a concurrency-1 queue, then plans and runs per-table `CREATE OR REPLACE … read_files` loads (up to 8 parallel) using live S3 and `information_schema` timestamps instead of append-only load logs.
> 
> The deleted `dbt/project/models/load/` PySpark models (`load__l2_sftp_to_s3`, `load__l2_s3_to_databricks`, and both Haystaq loaders) and `l2_expired_voters` are replaced by **`l2_voter_loaders`** (`SOURCE_GROUPS` for VM2, VM2Uniform, Haystaq flags/scores, and `Manual_ID_Omits.tab` → `l2_s3_expired_voters`) plus generic **`sftp_utils`**. Expired IDs follow the same S3 path as state files rather than worker-side MERGE into `airflow_source`.
> 
> **`databricks_utils`** drops expired-voter staging helpers and adds shared `connect_from_conn_id` / parameterized **`execute_with_retry`**. Dependencies add **`apache-airflow-providers-amazon`** for S3Hook; tests cover loader planning, sync safety, and DAG ordering (`plan_table_loads` uses **`all_done`** after sync). Minor dbt YAML cleanup removes Haystaq schema override meta from intermediate models.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d19bc6a878b8dd40e0e7e75532f0491ef0b568f2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danpelota:
delegate review

danpelota:
delegate review

danpelota:
delegate review

danpelota:
delegate review

danpelota:
delegate review

danpelota:
On the three findings from the last review:

**Taken — missing-member branch (`dd5741ee`).** Added `test_archive_missing_a_member_raises_and_leaves_no_file`. It pins both the `ValueError` and the `finally: os.remove`, which matters because a leaked 7 GB archive on a fixed 10 GiB worker would break the next sync.

**Disagreeing — Haystaq sync coverage (`e4e27bba`, `8290852b`).** `sync_source` has no Haystaq-specific path. It branches once on `file_name.lower().endswith(".zip")`, and `.tab.zip` takes the same branch as a VM2 `.zip`. Everything after that iterates `source["members"]`, which the caller supplies. A Haystaq sync test would run the identical code as `test_uploads_expected_members_and_trims_the_dictionaries` with a one-element list.

The member-name derivation is already covered elsewhere: `test_every_group_classifies_its_members_to_their_declared_type` asserts, for all four groups including both Haystaq ones, that `list_remote_sources` derives member names which `source_file_type` classifies to the declared type. That is the `removesuffix(".tab.zip") + ".tab"` logic the finding points at.

On the flat-name assumption specifically: a fixture we author cannot verify how L2 names members inside its real archives, only that `zipfile` round-trips a file we just wrote. What validated the real layout is the dev run in the PR description — `l2_s3_wy_haystaq_dna_flags` and `_dna_scores` came out byte-identical to prod, 0 rows differing on a two-way `EXCEPT` over every column but `loaded_at`.

This PR is a simplification pass over an already-reviewed change, so I am holding the test count down to cases where a plausible edit would break production behavior.

danpelota:
delegate review

danpelota:
delegate review

PR #838, opened 2026-08-17
[DATA-2326] Window the sales feeds on vendor event time, not extract time

sanjayr1:
## Problem

Both sales export feeds windowed and emitted "recent activity" as
`greatest(created_at, updated_at)`. On the BallotReady and TechSpeed intermediates those
two columns are *both* set to the Airbyte extraction timestamp, so they tracked when the
pipeline last saw a record rather than when anything happened to the candidacy.

BallotReady switched to re-sending its entire candidate file weekly. In the 2026+ slice of
BR staging, **101,373 of 103,823 rows carry a single extract day**, against a prior
weekly-delta norm of 8-319 rows:

| Extract day | Rows |
|---|---|
| 2026-08-17 | 101,373 |
| 2026-08-10 | 82 |
| 2026-08-03 | 58 |
| 2026-07-27 | 8 |
| 2026-07-20 | 23 |

Every weekly sync therefore re-marked the whole vendor universe as active today and
flooded both exports with long-dormant candidacies, which produced duplicate contacts in
HubSpot. The vendor's own `candidacy_updated_at` is untouched by re-delivery — on the same
rows it spreads across ten months — so the real signal was there all along and being
discarded.

## Change

Destination-facing only. **Canonical `created_at` / `updated_at` are not redefined** —
they feed every civics consumer through the provider precedence chain, so the blast radius
of changing them is the whole mart.

**1. New additive `vendor_activity_at` on both vendor intermediates,** carrying the
vendor's own event time: `candidacy_updated_at` for BallotReady (falling back to that
vendor's `candidacy_created_at` where the update time sits implausibly later than the row's
extract stamp, with a day of slack for ordinary clock skew), and the form's
`date_processed` for TechSpeed. `max()`-rolled to the candidacy grain, then re-maxed at the
published `gp_candidacy_id` grain so the dedup picker cannot discard a colliding rollup
group's value.

**Nullable by design.** Null means the provider supplied no usable event time. Extract time
is deliberately *not* substituted: as a non-null value it would outrank a correctly-dated
sibling provider wherever consumers take the greatest across providers, so an unparsed row
would beat an accurate one, and the next full-file re-delivery would sweep the whole
unparsed cohort back into the window at once.

**2. Both feeds window on, and emit, `feed_activity_at`:**

```sql
coalesce(
    greatest(br_vendor_activity_at, ts_vendor_activity_at, gp_api_case),
    greatest(created_at, updated_at)
)
```

`greatest()` skips nulls on Spark, so the coalesce fires only where no provider supplies an
event time — DDHQ-only and null-vendor rows, which keep documented status-quo behavior. The
window is bounded above as well as below: the gp_api leg and the canonical fallback are not
clamped, and one future-dated value would otherwise satisfy the lower bound forever.

**3. `candidacy_techspeed` has no TechSpeed leg, deliberately.** The candidacy mart sets
`'techspeed'` in `source_systems` exactly when a row exists in that intermediate, and this
model's ALREADY-SENT filter keeps only candidacies without it, so such a join can never
match (verified 0 of 2,030 rows). It becomes live when that filter becomes the sent_log
anti-join; a comment in the model records that.

**4. The emitted column keeps the name `last_activity_at`** in both feeds. Ops already
sorts and filters on it; the semantics improve, the contract does not break.

### Why the recency filter sits in its own CTE

It is defined once and both filtered on and emitted, so the two cannot drift apart — drift
between exactly those is the defect class being fixed. This costs real work: the old basis
referenced only the candidacy relation so it pushed into that scan, while the new one spans
the vendor intermediates and cannot push below them wherever it is written. The remaining
candidacy-only predicates still push down. See runtime below.

## Verification

`dbt build --favor-state` over the four changed models: **PASS=57, WARN=1, ERROR=0,
SKIP=0.** The one warning is the pre-existing
`assert_candidacy_hubspot_candidate_code_collision` at 2 rows, against 105 colliding codes
in prod under the old basis.

### Before / after

Each feed's **compiled** SQL run twice against identical upstream relations and the same
`current_date()`, swapping only the window basis so both arms keep an identical window
structure. The delta is attributable to the basis and nothing else.

| Feed | Before (extract-stamp basis) | After (vendor-event basis) | Delta |
|---|---|---|---|
| `candidacy_hubspot` | 7,102 | 4,190 | −2,912 (−41.0%) |
| `candidacy_techspeed` | 3,130 | 2,025 | −1,105 (−35.3%) |

### The recency is now entirely vendor-derived

| Measure | Result |
|---|---|
| `candidacy_hubspot` rows | 4,190 |
| …resolving to a real vendor event time | **4,190 (100%)** |
| …resolving to the extract-stamp fallback | **0** |
| Future-dated rows, either feed | **0** |
| Rows outside their own window, either feed | **0** |

Per-source composition of the surviving cohorts:

| Feed | Source | Rows | Activity range |
|---|---|---|---|
| `candidacy_hubspot` | ballotready | 1,754 | 2026-07-20 → 2026-08-16 |
| `candidacy_hubspot` | techspeed | 2,436 | 2026-07-20 → 2026-08-07 |
| `candidacy_techspeed` | (all) | 2,025 | 2026-08-02 → 2026-08-15 |

Every surviving row has genuine vendor activity inside its window, and the TechSpeed-return
cohort is retained.

## Operational notes

**Runtime.** Multiple samples, because a single cold build is misleading here:

| Model | Prod baseline | This branch (dev CTAS samples) |
|---|---|---|
| `candidacy_hubspot` | 22s (post anti-join refactor) | 23s, 24s, 39s, 34s |
| `candidacy_techspeed` | 48s min / 79s avg / 138s max over 25 builds | 113s, 65s, 61s, 60s |
| `int__civics_candidacy_ballotready` | not previously reported | ~27s |
| `int__civics_candidacy_techspeed` | not previously reported | ~28s |

`candidacy_techspeed`'s steady state (~60s) sits **inside** its existing prod band. An
earlier revision of this description reported 148s as a roughly 2x regression; that was a
cold first build, not a trend. `candidacy_hubspot` rises modestly from a 22s base. The
window predicate no longer prunes at the scan, so the join tree does process more rows, but
the measured cost is smaller than that suggests.

**Both feeds emit substantially fewer rows**, which is the point, but the first post-merge
run will look like a large volume drop to anyone watching export counts rather than the
incident.

**Uneven resolution.** TechSpeed captures a date, not a time, so 58.2% of
`candidacy_hubspot` rows land at midnight and sorting on `last_activity_at` ties within a
day. Stated in the column descriptions.

**Match-back completeness falls.** Rows with neither `candidate_code` nor `candidacy_id` go
from 9.8% to 15.3%, because event-time recency selects for TechSpeed-only rows over
BallotReady ones. A composition effect distinct from the volume drop: ops can reconcile
outcomes back to civics for a smaller share of the export, falling back to the internal
`gp_candidacy_id`.

**Reproducing locally** needs `--favor-state`. Without it, dbt Cloud's deferral prefers a
relation that already exists in your dev schema over production's, and a stale copy of
`stg_airbyte_source__techspeed_gdrive_candidates` predating `date_processed_date` fails the
build with `UNRESOLVED_COLUMN`. CI is unaffected — a fresh preview schema has nothing to
shadow.

## Tests

- **One singular regression test per feed, stated as a basis identity**: wherever a vendor
  supplies an event time, the feed's emitted recency must equal it, never the pipeline's
  extract stamp. Phrasing it as an identity rather than a staleness cutoff means it covers
  every row with a vendor leg (100% of the feed, 0 violations), fires if the window basis is
  reverted or either vendor join is dropped, and carries no date arithmetic that could
  false-fail when tests run after the build day. gp_api rows are excluded because their leg
  *is* the canonical product timestamps.
- `not_null` on the emitted `last_activity_at` in both feeds.
- No test mirrors a model's `WHERE` clause.

## Review history

Three rounds beyond the initial delegate pass: the delegate reviewer, Cursor Bugbot, an
independent model via the Codex CLI, and two high-effort in-repo review passes. They found
four real defects that CI and the first approval both missed, and corrected several claims
in earlier versions of this description. Every accepted finding, every decline with its
reason, and the owner decisions are recorded in the ticket's decision ledger so later rounds
cannot silently re-litigate them.

What the rounds changed, briefly:

1. The future-date clamp anchored to `_airbyte_extracted_at`, which is *last*-seen on an
   id-deduped table and so advances weekly — it did not deliver its own stated intent.
   Re-anchored to vendor-owned timestamps.
2. The dead TechSpeed join (above), and a regression test that asserted 630 of 4,245 rows
   while being structurally blind to the TechSpeed leg. Both tests rewritten as identities.
3. The extract-time substitution described above, which reintroduced the incident through
   the column built to remove it. Removing it also deleted three tests round 2 had added,
   including one that was itself unsound: it compared the fallback's `max()` of the extract
   stamp against `updated_at`'s `any_value()` of the same column, so it passed green for the
   22% of rollup groups spanning more than one extract.

## Known residuals and accepted tradeoffs

- **Eligibility narrows.** Because `sent_log` does not exist yet, the recency window is
  simultaneously the dedup mechanism and the eligibility gate, so tightening recency also
  narrows eligibility for leads not already in HubSpot. The "correctly suppressed re-send"
  and "never-sent new lead" halves cannot be separated without first-seen data, which this
  pipeline does not retain. **Accepted as a known tradeoff by the owner**, on the grounds
  that sent_log lands within a few sprints and separates the two concerns properly.
- **DDHQ still windows on extract stamps.** That intermediate exposes no event-time column,
  so DDHQ-only rows fall through to the canonical fallback — 538 reach this feed's base, 530
  sharing a single day, which is the bulk-restamp signature. The spec accepts status-quo for
  providers with no event time; closing it needs a DDHQ event-time column and is a separate
  ticket.
- **A future-dated row is excluded rather than capped**, so it is silently absent instead of
  visibly wrong. Capping was rejected because the only available anchor advances daily,
  which would re-cap the row fresh on every build and send it forever. 0 rows affected.
- **Routed as follow-ups:** the staging `date_processed` parse rule (accepts slash- but not
  hyphen-separated dates; recovers 3,757 of 6,335 unparsed rows), casting
  `candidacy_updated_at` in staging rather than the intermediate, moving the recency
  expression into `candidacy.sql`, and converting `candidacy_techspeed`'s correlated OR
  anti-join to the three hash anti-joins `candidacy_hubspot` already uses.

## Out of scope

`sent_log`, the person-graph email/phone tiers, any change to canonical `created_at` /
`updated_at`, the wider `any_value` determinism cleanup, and `added_to_mart_at` (already
documented as misleading and retired with the reverse-ETL work).

### DIFF
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -525,17 +525,21 @@ models:
 
       - name: vendor_activity_at
         description: >
-          Latest TechSpeed event time for the candidacy: the max of the form row's
-          date_processed across its stages, clamped by that row's extract stamp.
-          Unlike the BallotReady leg this source carries no stable vendor creation
-          timestamp, so an implausible future form date can only be clamped to the
-          extract stamp, which does move with each delivery; that still beats
-          passing a future value through, since a future timestamp sits inside
-          every rolling window permanently. Additive and destination-facing --
-          created_at/updated_at stay extract stamps for the provider precedence
-          chain, so consumers that need real candidate activity (the sales export
-          windows) read this instead. Non-null: rows with no parseable form date
-          fall back to extract time, which every row has.
+          Latest TechSpeed event time for the candidacy: the max across its stages of
+          the form's date_processed, clamped by that stage row's extract stamp. Stages
+          whose form date did not parse contribute nothing, so one unparseable stage
+          cannot lift a whole candidacy to ingestion time; extract time substitutes
+          only where NO stage parsed, which is why this stays non-null. Unlike the
+          BallotReady leg this source carries no stable vendor creation timestamp, so
+          an implausible future form date can only be clamped to the extract stamp,
+          which does move with each delivery; that still beats passing a future value
+          through, since a future timestamp sits inside every rolling window
+          permanently. A meaningful share of form dates do not parse today because the
+          staging rule accepts slash-separated US dates but not hyphen-separated ones
+          (a known follow-up); those candidacies fall back to extract time. Additive
+          and destination-facing -- created_at/updated_at stay extract stamps for the
+          provider precedence chain, so consumers that need real candidate activity
+          (the sales export windows) read this instead.
         data_tests:
           - not_null
 
@@ -1381,17 +1385,37 @@ models:
         description: >
           Latest BallotReady event time for the candidacy: the max across its stages
           of the vendor's own candidacy_updated_at, or -- where that is implausibly
-          later than the extract that carried it -- the vendor's candidacy_created_at.
-          Both are stable across re-deliveries, which matters: anchoring an
-          implausible value to the extract stamp instead would drag it forward on
-          every weekly full-file re-delivery and keep the row perpetually fresh.
+          later than the row's extract stamp -- the vendor's candidacy_created_at.
+          Both values are vendor-owned, so neither is re-written by a re-delivery of
+          an unchanged record; anchoring an implausible value to the extract stamp
+          instead would drag it forward on every weekly full-file re-delivery and
+          keep the row perpetually fresh. Caveat: the COMPARISON bound is the row's
+          latest extract stamp rather than the delivery that first carried it
+          (Airbyte id-dedup keeps one row per id), so which of the two branches a
+          row takes can change across re-deliveries even though the values cannot.
           Additive and destination-facing -- created_at/updated_at stay extract
           stamps for the provider precedence chain, so consumers that need real
-          candidate activity (the sales export windows) read this instead. A
-          re-delivery of an unchanged record does not move it. Non-null: rows with no
-          usable vendor time at all fall back to extract time, which every row has.
+          candidate activity (the sales export windows) read this instead. Non-null:
+          extract time substitutes only for candidacies where NO stage carried a
+          usable vendor timestamp.
         data_tests:
           - not_null
+
+    data_tests:
+      # The one check that can see the clamp itself regressing. Every other guard
+      # compares vendor_activity_at against something derived from it, so if the
+      # expression started returning extract time they would all still pass. Here,
+      # equality with updated_at (which IS the extract stamp) means the vendor time
+      # was lost. Every row supplies a usable vendor timestamp today, so this is
+      # currently zero. Warn, not error: a non-zero result means the vendor stopped
+      # populating its own timestamps, which is worth surfacing but is not our bug
+      # and should not block a build.
+      - dbt_utils.expression_is_true:
+          name: br_candidacy_vendor_activity_at_is_not_the_extract_stamp
+          arguments:
+            expression: "vendor_activity_at != updated_at"
+          config:
+            severity: warn
 
   - name: int__civics_elected_official_ballotready
     description: >
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -523,6 +523,19 @@ models:
         data_tests:
           - not_null
 
+      - name: vendor_activity_at
+        description: >
+          Latest TechSpeed event time for the candidacy: the max of the form row's
+          date_processed across its stages, clamped to that row's extract time so a
+          future-dated form value settles at when we first saw it rather than
+          reading as perpetually fresh. Additive and destination-facing --
+          created_at/updated_at stay extract stamps for the provider precedence
+          chain, so consumers that need real candidate activity (the sales export
+          windows) read this instead. Non-null: the clamp falls back to extract
+          time, which every row has.
+        data_tests:
+          - not_null
+
     tests:
       - dbt_utils.expression_is_true:
           name: techspeed_candidacy_has_general_or_primary_date
@@ -1358,6 +1371,20 @@ models:
           - not_null
 
       - name: updated_at
+        data_tests:
+          - not_null
+
+      - name: vendor_activity_at
+        description: >
+          Latest BallotReady event time for the candidacy: the max of the vendor's
+          own candidacy_updated_at across its stages, clamped to that stage row's
+          extract time so the vendor's known future-dated values settle at when we
+          first saw them rather than reading as perpetually fresh. Additive and
+          destination-facing -- created_at/updated_at stay extract stamps for the
+          provider precedence chain, so consumers that need real candidate activity
+          (the sales export windows) read this instead. A re-delivery of an
+          unchanged record does not move it. Non-null: the clamp falls back to
+          extract time, which every row has.
         data_tests:
           - not_null
 
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -526,13 +526,16 @@ models:
       - name: vendor_activity_at
         description: >
           Latest TechSpeed event time for the candidacy: the max of the form row's
-          date_processed across its stages, clamped to that row's extract time so a
-          future-dated form value settles at when we first saw it rather than
-          reading as perpetually fresh. Additive and destination-facing --
+          date_processed across its stages, clamped by that row's extract stamp.
+          Unlike the BallotReady leg this source carries no stable vendor creation
+          timestamp, so an implausible future form date can only be clamped to the
+          extract stamp, which does move with each delivery; that still beats
+          passing a future value through, since a future timestamp sits inside
+          every rolling window permanently. Additive and destination-facing --
           created_at/updated_at stay extract stamps for the provider precedence
           chain, so consumers that need real candidate activity (the sales export
-          windows) read this instead. Non-null: the clamp falls back to extract
-          time, which every row has.
+          windows) read this instead. Non-null: rows with no parseable form date
+          fall back to extract time, which every row has.
         data_tests:
           - not_null
 
@@ -1376,15 +1379,17 @@ models:
 
       - name: vendor_activity_at
         description: >
-          Latest BallotReady event time for the candidacy: the max of the vendor's
-          own candidacy_updated_at across its stages, clamped to that stage row's
-          extract time so the vendor's known future-dated values settle at when we
-          first saw them rather than reading as perpetually fresh. Additive and
-          destination-facing -- created_at/updated_at stay extract stamps for the
-          provider precedence chain, so consumers that need real candidate activity
-          (the sales export windows) read this instead. A re-delivery of an
-          unchanged record does not move it. Non-null: the clamp falls back to
-          extract time, which every row has.
+          Latest BallotReady event time for the candidacy: the max across its stages
+          of the vendor's own candidacy_updated_at, or -- where that is implausibly
+          later than the extract that carried it -- the vendor's candidacy_created_at.
+          Both are stable across re-deliveries, which matters: anchoring an
+          implausible value to the extract stamp instead would drag it forward on
+          every weekly full-file re-delivery and keep the row perpetually fresh.
+          Additive and destination-facing -- created_at/updated_at stay extract
+          stamps for the provider precedence chain, so consumers that need real
+          candidate activity (the sales export windows) read this instead. A
+          re-delivery of an unchanged record does not move it. Non-null: rows with no
+          usable vendor time at all fall back to extract time, which every row has.
         data_tests:
           - not_null
 
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -526,15 +526,18 @@ models:
       - name: vendor_activity_at
         description: >
           Latest TechSpeed event time for the candidacy: the max across its stages of
-          the form's date_processed, clamped by that stage row's extract stamp. DAY
+          the form's date_processed, taken only where it is not implausibly later than
+          the extract that carried it (a day of slack for clock skew, matching the
+          BallotReady leg). DAY
           RESOLUTION ONLY -- the form captures a date, not a time, so values land at
           midnight and consumers sorting on this will see ties within a day.
-          NULLABLE BY DESIGN: null means the form date did not parse, and consumers
-          coalesce that to the canonical timestamps. Extract time is deliberately NOT
-          substituted -- as a non-null value it would outrank a correctly-dated
-          sibling provider wherever consumers take the greatest across providers, so
-          the next full-file re-delivery would sweep every unparsed candidacy back
-          into a recency window at once. Parse failures are not rare: the staging rule
+          NULLABLE BY DESIGN: null means the form date is missing, unparseable, or
+          implausibly future, and consumers coalesce that to the canonical timestamps.
+          Extract time is deliberately NOT substituted -- as a non-null value it would
+          outrank a correctly-dated sibling provider wherever consumers take the
+          greatest across providers, and because it advances with every delivery the
+          row would tick forward forever, sweeping the whole affected cohort back into
+          a recency window on each re-delivery. Parse failures are not rare: the staging rule
           accepts slash-separated US dates but not hyphen-separated ones, which is a
           known follow-up. Additive and destination-facing -- created_at/updated_at
           stay extract stamps for the provider precedence chain, so consumers that
--- a/dbt/project/models/intermediate/civics/int__civics.yaml
+++ b/dbt/project/models/intermediate/civics/int__civics.yaml
@@ -526,22 +526,19 @@ models:
       - name: vendor_activity_at
         description: >
           Latest TechSpeed event time for the candidacy: the max across its stages of
-          the form's date_processed, clamped by that stage row's extract stamp. Stages
-          whose form date did not parse contribute nothing, so one unparseable stage
-          cannot lift a whole candidacy to ingestion time; extract time substitutes
-          only where NO stage parsed, which is why this stays non-null. Unlike the
-          BallotReady leg this source carries no stable vendor creation timestamp, so
-          an implausible future form date can only be clamped to the extract stamp,
-          which does move with each delivery; that still beats passing a future value
-          through, since a future timestamp sits inside every rolling window
-          permanently. A meaningful share of form dates do not parse today because the
-          staging rule accepts slash-separated US dates but not hyphen-separated ones
-          (a known follow-up); those candidacies fall back to extract time. Additive
-          and destination-facing -- created_at/updated_at stay extract stamps for the
-          provider precedence chain, so consumers that need real candidate activity
-          (the sales export windows) read this instead.
-        data_tests:
-          - not_null
+          the form's date_processed, clamped by that stage row's extract stamp. DAY
+          RESOLUTION ONLY -- the form captures a date, not a time, so values land at
+          midnight and consumers sorting on this will see ties within a day.
+          NULLABLE BY DESIGN: null means the form date did not parse, and consumers
+          coalesce that to the canonical timestamps. Extract time is deliberately NOT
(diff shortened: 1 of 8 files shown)


sanjayr1:
delegate review

sanjayr1:
Both findings addressed. Details and evidence below.

---

### 1. Missing parallel regression test on `candidacy_techspeed` — **accepted, fixed**

`tests/assert_candidacy_techspeed_vendor_recency_basis.sql` (new). You are right that
the sibling test does not cover this model: it has its own window predicate and its
own vendor joins, including the LEFT join added in this PR, so a revert or a dropped
join there would have gone unnoticed.

Built as suggested, with two deviations worth naming:

- 17-day interval (16-day window plus one day of midnight-rollover slack), matching
  the 31-for-30 pattern in the sibling.
- The gp_api exclusion comes from a join to `candidacy`, not from a column on the
  feed. That feed's output is a fixed 42-column legacy contract, so it does not emit
  `candidate_id_source` and cannot without breaking the contract.

Dry-run against the CI preview schema before committing: **0 rows**, so it passes
rather than landing red.

---

### 2. `least()` clamp is not sync-stable — **accepted, fixed on the BallotReady leg; declined on the TechSpeed leg, with reasons**

The mechanism claim is correct and it was a real gap between the intent and the
implementation. `_airbyte_extracted_at` on an id-deduped Airbyte table is *last*-seen,
not first-seen, so it advances with every weekly full-file re-delivery. The intent was
an anchor that does not move; `least(vendor_ts, extract)` was not that.

**It is not theoretical.** A current-snapshot query shows zero rows with a vendor
timestamp after their own extract stamp, which would suggest the clamp never bites —
but that is a snapshot artifact. In data delivered on 2026-08-13 the vendor's maximum
`candidacy_updated_at` was `2026-08-14T01:24Z`, i.e. future-dated relative to the
delivery that carried it. The 08-17 full re-delivery overwrote those rows, so the
receipts are gone from the current table. Future-dated vendor timestamps recur.

**Fix (BallotReady leg).** Explicit preference order, both leading branches anchored to
values the vendor owns and that do not move across re-deliveries:

1. `candidacy_updated_at`, when it is at or before the extract that carried it.
2. `candidacy_created_at`, when *that* is at or before the extract — used when the
   update time is implausibly in the future.
3. Extract time, last resort only, for rows carrying no usable vendor time at all.

**Why not an actual first-seen timestamp:** it is not recoverable from the warehouse.
This connector id-dedups, so each id exists once, carrying its most recent extract
stamp — verified independently on the raw table: 341,817 rows against 341,817 distinct
ids, `_airbyte_generation_id` = 0 across all 18 months (no reset ever), and 0 of 78
delivered files extracted on more than one day (no history re-reads). There is no
per-id history to take a `min()` over. A different stable anchor was the only option.

**The trade-off, stated plainly.** Falling back to creation time can make a
genuinely-just-updated record read as stale. That is the safer of the two errors, and
it is self-limiting: as soon as the next delivery's extract stamp exceeds the
future-dated value, branch 1 takes over and the row gets its real vendor timestamp
back. Anchoring to the moving extract stamp has no such recovery — the row sits inside
every rolling window permanently. Given this PR exists because re-stamped rows flooded
a lead export and produced duplicate contacts, the conservative reading is the correct
default.

**Declined on the TechSpeed leg**, and documented in the model and the yaml rather than
papered over. That source carries no stable vendor creation timestamp — only
`date_processed` — so an implausible future form date can only be clamped to our own
extract stamp, which does move. It is still strictly better than passing the future
value through: a future timestamp sits inside every rolling window permanently, versus
being re-anchored weekly. Measured on that source: 0 of 90,248 rows have
`date_processed` after their extract stamp, and the 6,335 rows with no parseable date
fall back to extract time, which is the pre-existing documented behavior for rows with
no vendor signal.

---

### Verification of the fix

The change is behavior-preserving on current data: every row takes branch 1 today, so
`vendor_activity_at` is byte-identical before and after. Verified by snapshotting the
column, rebuilding, and diffing — numbers in the updated PR body. Only the
future-dated path changes behavior, which is the point.

sanjayr1:
delegate review

sanjayr1:
Round 2. Two further independent reviews were run against this PR — one from a
different model via the Codex CLI, one a high-effort in-repo review pass. They
converged on two defects that the first round and CI both missed. Nine findings
accepted, seven declined; every disposition is recorded in the ticket's decision
ledger so later rounds do not re-litigate settled ones.

### The two that mattered

**The LEFT join to `int__civics_candidacy_techspeed` was dead, and I had claimed
otherwise.** The candidacy mart sets `'techspeed'` in `source_systems` exactly when a
row exists in that intermediate (`candidacy.sql:163`), and this model's ALREADY-SENT
filter keeps only candidacies *without* it — so the join could never match. Verified 0
of 2,030 rows on same-vintage data. My earlier "58 rows prove it is live" compared a
stale feed against a newer intermediate, which is the same vintage-mismatch error I had
already caught myself making once in this PR. Removed, along with its leg of the recency
expression; a comment records that it comes back when that filter becomes the sent_log
anti-join.

**The regression test asserted 630 of 4,245 rows and was blind to the TechSpeed leg.**
`inner join` on the BR intermediate dropped every row without a BR leg, and
`ts.vendor_activity_at is null` dropped every row that had one — so a regression
confined to the TechSpeed leg produced zero rows and a green build. Both tests are now
basis identities instead of staleness cutoffs: wherever a vendor supplies an event time,
the emitted recency must equal it. Coverage went from 14.8% to **100% of the feed (4,190
of 4,190 rows, 0 violations)**, it now fires if the window basis is reverted or either
vendor join is dropped, and it carries no date arithmetic — which also removes a
false-fail whenever tests run a day or more after the table build, and retires four
uncoupled hardcoded window literals.

### Also fixed

- **Both windows are now bounded above.** The vendor legs are clamped; the gp_api
  product-timestamp leg and the canonical fallback are not, so one future-dated value
  would have satisfied the lower bound forever and parked a row on every export. Bounded
  the window rather than clamping that leg to the current time, since clamping to "now"
  is precisely the perpetual-freshness behavior this PR exists to remove.
- **The TechSpeed fail-open no longer propagates across stages.** Vendor event time is
  NULL per stage row where the form date did not parse, and each rollup substitutes
  extract time only when *no* stage of the candidacy had one. Previously a single
  unparseable stage contributed its delivery stamp to the candidacy-level max, lifting
  the whole candidacy to ingestion time — the incident shape, inside the column built to
  remove it. ~260 candidacies were exposed. Same shape applied to the BallotReady leg.
- **One warn-severity check** asserting the BallotReady leg is not simply the extract
  stamp. Every other guard compares that column against something derived from it, so a
  clamp regression was invisible; equality with the extract stamp is its only observable
  signature. Currently 0. Warn, not error: a hit means the vendor stopped populating its
  own timestamps.
- Removed a dead lexicographic string `max()` over the vendor's `updated_at` (dropped by
  the next CTE's projection, no reader anywhere), and corrected descriptions that
  overstated re-delivery stability, described the recency expression as either/or when it
  is a simultaneous `greatest()`, and claimed the tests could detect a clamp regression.

### Declined, so they are not re-raised

- **Fix the staging `date_processed` parse rule** to accept hyphen-separated dates —
  real, and it would recover 3,757 of 6,335 unparseable rows, but it changes a staging
  model shared with the officeholders sibling. Routed as a follow-up; the propagation fix
  above removes the dangerous half.
- **Move the recency expression into `candidacy.sql`** so it is not duplicated across
  both marts — the four-way FOJ is the right long-term home, but that is a mart-contract
  change on an urgent destination-facing fix.
- **Gate gp_api on `array_contains(source_systems, 'gp_api')`** — that records a gp_api
  row *exists*, whereas `candidate_id_source = 'gp_api'` records that gp_api *won
  precedence*. Swapping would widen which rows receive the product-timestamp leg. That is
  a behavior change, not a hardening.
- **`any_value` → `max` on the extract stamp** to make the canonical timestamps
  deterministic — canonical `created_at`/`updated_at` are a binding constraint on this
  work. 0 colliding rollup groups today (84,768 groups, 84,768 distinct ids).
- **Pin the Spark session timezone** — a project-wide config change. Measured 0 of 85,331
  rows tripping the clamp under the current default.
- **Replace the `not_null` tests on `last_activity_at`** — they cannot fail today because
  the window predicate already drops nulls. Kept because they still guard a future change
  that emits a different nullable column; the misleading rationale in the yaml was
  corrected instead.
- **The eligibility narrowing** — because `sent_log` does not exist yet, the recency
  window is simultaneously the dedup mechanism and the eligibility gate, so tightening
  recency also narrows eligibility for leads not already in HubSpot. Raised, quantified
  (2,492 rows on fixed upstream; the "correctly suppressed re-send" and "never-sent"
  halves cannot be separated without first-seen data, which this pipeline does not
  retain), and **accepted as a known tradeoff by the owner** on the grounds that the
  sent_log table lands within a few sprints and will separate the two concerns properly.

### Verification

Rebuild: `PASS=60 WARN=1 ERROR=0 SKIP=0`. The one warning is the pre-existing
`assert_candidacy_hubspot_candidate_code_collision`, which sits at 2 here against 105
colliding codes in prod under the old basis. Both feeds: 0 future-dated rows, 0 outside
their window. BallotReady leg: 0 rows falling back to extract time, 0 exceeding it.

sanjayr1:
delegate review

sanjayr1:
Round 3. A third review pass (high effort, ten parallel finder angles, most findings
carrying measured numbers from prod and this PR's own preview schema) found that round 2's
fix was incomplete in a way that mattered. Six findings accepted, nine declined or
disclosed; all recorded in the ticket's decision ledger.

### The one that mattered

**The extract-time substitution was worse than no value at all.** Both intermediates fell
back to the Airbyte extract stamp when a vendor supplied no usable event time. Because the
feeds take `greatest()` across providers, that non-null stamp *outranked* a correctly-dated
sibling: a TechSpeed row whose form date failed to parse beat an accurate BallotReady event
time. 6,039 of 85,331 TechSpeed candidacies were on extract time, and most feed rows carry
a TechSpeed leg — so the next full-file re-delivery would have swept that whole cohort into
the window at once. That is this incident, reintroduced through the column built to remove
it.

Both columns are now **vendor-time-or-NULL**. Null means the provider supplied no event
time, and the feeds already coalesce that to the canonical timestamps, which is the
documented status-quo path. Measured after: **all 4,190 feed rows resolve to a real vendor
event time, none to an extract stamp.**

The fix deletes machinery rather than adding it, and it retired three tests round 2 had
justified — the two `not_null` tests on the now-nullable columns, and the warn check
asserting the BallotReady value was not the extract stamp. That check was unsound anyway:
it compared the fallback's `max(_airbyte_extracted_at)` against `updated_at`'s
`any_value()` of the same column, so it passed green for the 22% of rollup groups spanning
more than one extract. Round 3 is net **−8 lines**.

### Also fixed

- **BallotReady branch selection now allows a day of clock skew.** Comparing the vendor's
  clock against Airbyte's meant a record updated minutes after the file was cut fell to its
  creation time months earlier, then jumped back the following week. A row leaving and
  re-entering the window is a duplicate send, because the window IS the interim dedup.
- **The BallotReady vendor time is re-maxed at the published grain** before the dedup
  picker. The rollup groups on candidate/position/year but the published key is the
  cluster-derived id, and two groups can mint the same one — which is why that picker
  exists. Without this it discarded the losing group's vendor event time. The TechSpeed
  sibling already did this.
- Removed a dead `is not null` guard in the HubSpot test (the inequality against a null
  `greatest()` already filters the row), and corrected the pushdown comment, which claimed
  the candidacy-only predicates still push down without admitting the recency predicate
  used to be one of them.

### A runtime claim in this PR was wrong, and is corrected

Earlier revisions of the description reported `candidacy_techspeed` at 148s against a
48-138s prod band, i.e. a roughly 2x regression. Multiple samples show the steady state is
~60s (113s, 65s, 61s, 60s) — **inside** the existing band. The 148s was a cold first build.
The reviewer that refuted this was right. The description now carries all the samples plus
the two intermediates' runtimes, which had not been disclosed at all.

### Disclosed rather than fixed

- **TechSpeed captures a date, not a time**, so 58.2% of `candidacy_hubspot` rows land at
  midnight and sorting on `last_activity_at` ties within a day. No code fix exists; now
  stated in the column descriptions.
- **Match-back completeness falls**: rows with neither `candidate_code` nor `candidacy_id`
  go from 9.8% to 15.3%, because event-time recency selects for TechSpeed-only rows over
  BallotReady ones. A composition effect distinct from the volume drop.
- **DDHQ still windows on extract stamps** — that intermediate exposes no event-time column,
  so DDHQ-only rows use the canonical fallback: 538 reach this feed's base, 530 sharing a
  single day, which is the bulk-restamp signature. The spec accepts status-quo for providers
  with no event time; closing it needs a DDHQ event-time column and is a separate ticket.

### Declined, so they are not re-raised

- **Cap a future-dated leg instead of excluding the row** — the proposed anchor
  (`current_date() + 1 day`) advances *daily*, so a future-dated row would be re-capped
  fresh on every build and sent forever. That is the perpetual-freshness failure this PR
  exists to remove. Excluding an impossible date is the lesser harm; residual accepted, 0
  rows affected.
- **A test on the window bound**, so that reverting `from in_window` is caught — it mirrors
  the model's `WHERE`, the pattern the team retired in July, and reintroduces the date
  arithmetic this round removed. A window reversion moves the feed from ~4k to ~84k rows,
  which is not a subtle signal.
- **Both singular tests compare a materialized mart value against a live intermediate read**
  and could false-fail on a partial rebuild — real, and logged. The column is now purely
  vendor-derived so it moves only when the vendor changes the record; `dbt build` runs
  models and tests in one invocation; and the repo's existing
  `assert_candidacy_hubspot_company_ids_populated` carries the same cross-relation
  exposure. Will downgrade to warn if it ever flakes.
- **Cast `candidacy_updated_at` in staging** and **move the recency expression into
  `candidacy.sql`** — both rebuttals were well evidenced and I am not disputing them on the
  merits; both widen CI scope on an urgent fix, so both are routed as follow-ups with the
  evidence attached.
- **The feeds' `not_null` tests are vacuous** — consistent with round 2's decline; they
  still guard a future change that emits a different nullable column.

### Verification

`PASS=57 WARN=1 ERROR=0 SKIP=0`. Both feeds: 0 future-dated, 0 outside their window, 100%
of recency vendor-derived.

sanjayr1:
delegate review

sanjayr1:
Round 4. The blocker is correct and it caught an inconsistency I introduced. Accepted and
fixed in `0b5717ea`.

The previous commit was titled "stop substituting extract time for a missing vendor event
time" and left exactly one path that still did it. `least(date_processed, extract)` returns
the extract stamp whenever the form date parses as later than the extraction that carried
it, and because that stamp advances on every full-file re-delivery, such a row would tick
forward with each one and sit inside every recency window permanently. That is the flood
behaviour this column exists to prevent.

The BallotReady leg had already been given the null-on-implausible shape with a day of
slack for clock skew; I did not carry it across. Both legs now share it, so the rule is one
rule rather than two: take the vendor's own event time where it is plausible, otherwise
supply nothing and let consumers fall back to the canonical timestamps.

Two small departures from the suggested patch, both in the same direction:

- `least()` is removed rather than kept alongside a guard, so the two legs are now
  structurally identical.
- The explicit `date_processed_date is not null` check is dropped, because a null
  comparison is not true and the row falls out of the `CASE` on its own. The BallotReady
  leg never had one, and a redundant null guard was removed from one of these tests last
  round for the same reason.

Measured: **0 of 90,248** staged TechSpeed rows carry a form date later than their own
extract stamp, with or without the day of slack — so this is latent rather than live. It is
still worth fixing: it was a principle violation against this PR's own stated rule, and the
fix deletes code.

The column description is updated too; it still described a clamp that no longer exists.

Rebuild: `PASS=57 WARN=1 ERROR=0 SKIP=0`, both recency-basis tests green. The one warning is
the pre-existing `assert_candidacy_hubspot_candidate_code_collision` at 2 rows, against 105
colliding codes in prod under the old basis.

sanjayr1:
delegate review

PR #839, opened 2026-08-17
calib(data): dashboard-view era 3, view-count caveat, impersonation marker

datawithtristan:
Calibration batch from the **DATA-2206** dashboard viewer-leg analysis (calibration log dated 2026-08-17), combined with an approved consolidation pass. Combined rather than split into two PRs because both batches rewrite the same `engagement.md` section.

## Calibration — Track 1 (data)

`.claude/skills/win-analytics-knowledge/references/engagement.md`

- **Dashboard view counts measure instrumentation as much as behaviour.** The 30-second dedup removes **67–68%** of raw dashboard events (2025: 115,494 → 38,114; 2026: 75,853 → 24,105). In 2025 only one dashboard event existed, so this is genuine re-firing rather than the Apr–Jun 2026 co-fire; the rate falls to ~18% after the July rebuild. Recommends distinct view-days over any view count, since the 30s boundary has no natural valley (2.8% of events at 31–60s, 5.5% at 61–300s).
- **The sales-impersonation marker.** `event_properties.impersonation` is live from 2026-03-24 across all event types, but the capability shipped 2024-08 as a client-side context swap and fired unmarked as the impersonated user — so the pre-marker era cannot be assumed clean. Marked-era rates: 9.0% of dashboard views vs 0.2% of sends. Also records that `total_campaigns_sent` is Amplitude-sourced, so both legs are subtractable on the marker.

This serves the standing impersonation caveat, which previously said "look for a marker" without saying where one is.

## Consolidation — approved item by item

`engagement.md`, `gotchas.md`

Three warnings about the 2026-06-13 dashboard-event death were stale: DATA-2173 shipped 2026-07-24 and the flags recovered. Verified against live data before touching anything — `is_active_candidate_30d` reads **382**, not 0.

None are plain deletions. Each is rewritten, because a **third** break landed **2026-07-31** with a different mechanism and different dates, and leaving the old warning risked reading as though the new one were covered.

| item | file | was | now |
|---|---|---|---|
| C1 | engagement.md | "reads FALSE for all ~62k users", fix ticketed | verified era-3 timeline, successor event, current flag readings |
| C2 | engagement.md:34 | stale `fix ticketed DATA-2173` clause | pointer to the era-3 section |
| C3 | gotchas.md:44 | "Canonical dashboard-view event is dead", `fix ticketed — re-check` | "instrument keeps moving", live symptom, DATA-2337 open |

[...]

PR #842, opened 2026-08-18
calib(data): family-touch comparability and nav-click classification

datawithtristan:
Track 1 calibration batch from the surface-map vs nav-click reconciliation
(follows the ledger-only #840). Two commits.

## Commit 1 — engagement.md, `+6 / -2`

Both edits sharpen existing text rather than adding new rules.

**1. Generalize the `win_pro_upgrade` caveat into a rule about family touch.**
Event families are instrumented to very different depths, so "share of users who
fired any event in family X" partly measures how thoroughly X was instrumented.
Ranking a thin family against a thick one is not a like-for-like comparison.

The existing impression-dominated paragraph is preserved **verbatim** as the first
of three instances. Two join it:

- `win_resources` holds exactly one event type, `Resources - Resource Clicked`,
  live 2025-10-15 to 2026-02-27 and silent since.
- `win_voter_data` swapped event sets mid-2026: legacy `Voter Data -%` dark
  2026-07-06..07-18, successors from 2026-07-21. Called out explicitly: **a
  family-level availability check does not catch this**, because the successors
  keep the family's own first/last-seen span continuous.

**2. Record that navigation events are classified by name, not destination.**
A nav click never reaches the `win_*` family of the surface it opens, so a
win-family touch rate measures **use**, not visitation. That distinction is the
whole of the reconciliation this batch came from: an Amplitude nav-click chart
and the surface map were read as contradicting each other when they were
answering different questions.

## Commit 2 — record the 2026-08-17 consolidation pass

The consolidation pass that landed in #839 edited `engagement.md` and `gotchas.md`
and appended its ledger rows, but never updated the `Last consolidation pass`
header, which still read 2026-07-21. Anything reading that header to decide
whether a pass is due sees a four-week gap that does not exist — it fired that way
the next day. Header corrected, prior pass moved to the previous slot, and one
process row appended noting that a pass which updates docs but not its own
timestamp is invisible to the check meant to schedule the next one.

## Verification

Instrumentation breadth, queried against `int__amplitude_event_catalog`
2026-08-17 (`where is_win`, count of event types per family):

| family | event types | first seen | last seen |
|---|---|---|---|
| win_onboarding | 72 | 2025-05-28 | 2026-08-17 |
| win_pro_upgrade | 52 | 2025-05-28 | 2026-08-17 |
| win_candidate_profile | 32 | 2025-05-28 | 2026-08-17 |
| win_voter_data | 31 | 2025-05-28 | 2026-08-17 |
| ... | | | |
| win_p2p_upgrade | 3 | 2025-09-10 | 2026-08-16 |
| **win_resources** | **1** | **2025-10-15** | **2026-02-27** |

72:1 spread, which is the number quoted in the doc.

The `win_voter_data` row is the evidence for the second bullet: it reports
`last_seen` 2026-08-17 at family level even though the event set analysts actually
query died in July. The successors keep the span continuous, so the obvious health
check passes green on a broken number.

This is not hypothetical. The analysis that produced this batch published a product
recommendation off a `win_resources` rate before the single-dead-event problem was
found; correcting for the missing instrument moved that rate from 0.30 to 0.59 and
removed it as an outlier.

The four candidate rows from the reconciliation run itself landed in #840.
`pre-commit` passes on both changed files.


datawithtristan:
delegate review

